PR #2, opened 2021-02-10
Remove Google Maven repo

fmeum:
Google's Maven repository does not offer most of our dependencies, which
makes Bazel print quite a few warnings while downloading.


simonresch (approved):
Looks good to me.

PR #3, opened 2021-02-10
Optionally read fuzz target class from JAR manifest

fmeum:
By reading the fuzz target class from a JAR manifest entry rather than
a command-line argument, fuzz targets can be fully self-contained.

This commit adds a Java function that looks for a unique
`Fuzz-Target-Class` attribute in all manifests on the classpath and
returns it to the driver if found. If no such entry is found, it
falls back to the `--target_class` commandline parameter.

### DIFF
--- a/driver/fuzz_target_runner.cpp
+++ b/driver/fuzz_target_runner.cpp
@@ -52,6 +52,9 @@ DEFINE_string(reproducer_path, ".",
               "Path at which fuzzing reproducers are stored. Defaults to the "
               "current directory.");
 
+constexpr auto kManifestUtilsClass =
+    "com/code_intelligence/jazzer/runtime/ManifestUtils";
+
 namespace jazzer {
 // split a string on unescaped spaces
 std::vector<std::string> splitOnSpace(const std::string &s) {
@@ -78,9 +81,13 @@ FuzzTargetRunner::FuzzTargetRunner(
     JVM &jvm, const std::vector<std::string> &additional_target_args)
     : ExceptionPrinter(jvm), jvm_(jvm), ignore_tokens_() {
   auto &env = jvm.GetEnv();
-  // the first positional argument should be the fuzz target class
   if (FLAGS_target_class.empty()) {
-    std::cerr << "Missing argument --target_class <fuzz_target_class>"
+    FLAGS_target_class = DetectFuzzTargetClass();
+  }
+  // If automatically detecting the fuzz target class failed, we expect it as
+  // the value of the --target_class argument.
+  if (FLAGS_target_class.empty()) {
+    std::cerr << "Missing argument --target_class=<fuzz_target_class>"
               << std::endl;
     exit(1);
   }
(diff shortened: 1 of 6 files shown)


agent/src/main/java/com/code_intelligence/jazzer/runtime/FuzzTargetFinder.kt
```diff
@@ -0,0 +1,37 @@
...
+    @JvmStatic
+    fun detectFuzzTargetClass(): String? {
+        val manifests = FuzzTargetFinder::class.java.classLoader.getResources("META-INF/MANIFEST.MF")
+        val fuzzTargets = manifests.asSequence().mapNotNull { url ->
+            url.openStream().use { inputStream ->
+                val manifest = Manifest(inputStream)
+                manifest.mainAttributes.getValue("Fuzz-Target-Class")
+            }
+        }.toList()
+        return if (fuzzTargets.size == 1) fuzzTargets.first() else null
```
simonresch, line 35: If two jar files with this manifest entry end up in the class path the user gets no indication that this manifest entry is being ignored. Do you think it makes sense to log a simple warning message in that case?

driver/fuzz_target_runner.cpp
```diff
@@ -78,7 +81,11 @@ FuzzTargetRunner::FuzzTargetRunner(
     JVM &jvm, const std::vector<std::string> &additional_target_args)
     : ExceptionPrinter(jvm), jvm_(jvm), ignore_tokens_() {
   auto &env = jvm.GetEnv();
-  // the first positional argument should be the fuzz target class
+  if (FLAGS_target_class.empty()) {
+    FLAGS_target_class = DetectFuzzTargetClass();
+  }
+  // If automatically detecting the fuzz target class failed, we expect it as
```
simonresch, line 87: The command was inaccurate before as well, maybe we can fix it now: the fuzz target class is not passed as a positional argument.

driver/fuzz_target_runner.cpp
```diff
@@ -78,7 +81,11 @@ FuzzTargetRunner::FuzzTargetRunner(
...
+  // the first positional argument.
   if (FLAGS_target_class.empty()) {
     std::cerr << "Missing argument --target_class <fuzz_target_class>"
```
simonresch, line 90:
```suggestion
    std::cerr << "Missing argument --target_class=<fuzz_target_class>"
```

agent/src/main/java/com/code_intelligence/jazzer/runtime/FuzzTargetFinder.kt
fmeum: I added a warning.

fmeum:
I have addressed the comments and generalized the approach to the list of custom hooks.

fmeum:
I improved the error handling for invalid hook class names as these were the reason for a hard-to-understand CI failure.

simonresch (approved):
Looks good to me. Thanks for the changes. :)

agent/src/main/java/com/code_intelligence/jazzer/runtime/FuzzTargetFinder.kt
simonresch: Yes, I think the warning would be enough. Thanks!

PR #13, opened 2021-02-19
Add pickValue(Collection) and consumeChar(char min, char max)

fmeum:


simonresch (approved):
Good addition. LGTM. :)

PR #14, opened 2021-02-22
Use NeverZero instead of Saturated Counters for coverage

fmeum:
According to
https://www.usenix.org/system/files/woot20-paper-fioraldi.pdf,
letting coverage 8-bit counters wrap from 255 to 1 on increment performs
better than having them stay at 255. In fact, the latter has been
observed to hurt overall coverage.


sebastianpoeplau (approved):
Looks good to me!

PR #18, opened 2021-02-23
Make fuzzerTestOneInput void

fmeum:
Java assertion errors are impossible to deduplicate and easily replaced
by an assert or a check and a custom exception.

This commit makes both variants of fuzzerTestOneInput void methods and
adds a note about this change to the respective error message.


simonresch (approved):
Looks good to me.

We should only keep in mind to also update the fuzz targets in our core repository when we update the jazzer version.

PR #22, opened 2021-02-26
Provide basic macOS compatibility

fmeum:
With the exception of the native and the fork test, everything seems to work fine on macOS 10.15 with this PR.


fmeum:
GitHub Actions is currently broken, but the tests passed previously. I tried to make the native fuzz test work on macOS, but since the issue is purely a Bazel bug, I will not put anymore effort into working around it.

fmeum:
I rebased onto main.

sebastianpoeplau (approved):

PR #27, opened 2021-03-01
Enable assertions and severity markup for fuzz targets

fmeum:
Allows using `assert` in fuzz targets combined with a custom exception type for security bugs (with an optionally specified severity).

### DIFF
--- a/agent/src/main/java/com/code_intelligence/jazzer/api/FuzzerSecurityIssueCritical.java
+++ b/agent/src/main/java/com/code_intelligence/jazzer/api/FuzzerSecurityIssueCritical.java
@@ -1,24 +1,39 @@
+// Copyright 2021 Code Intelligence GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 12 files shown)


agent/src/main/java/com/code_intelligence/jazzer/api/Severity.java
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * The severity level of a detected security issue.
+ */
+public enum Severity {
+  /**
+   * Examples of security issues of critical severity include: remote code execution
+   */
+  CRITICAL,
+  /**
+   * Examples of security issues of high severity include: SQL injections
```
fmeum, line 26: These could use some more examples (as well as consensus on whether these rankings are indeed sensible). @sebastianpoeplau Do you have some ideas on what kind of issues to add here that apply to memory-safe languages?

simonresch: We could take some items from the OWASP Top 10 list, no? E.g. "Broken Authentication", "Sensitive Data Exposure" could be checked with specialized fuzz targets.

simonresch (approved):
Cool addition. Changes look good to me.

agent/src/main/java/com/code_intelligence/jazzer/api/Severity.java
fmeum: I think that the severity of such rather generic issues would depend a lot more on context than for something like RCE. I have weakened the wording a bit.

PR #45, opened 2021-03-19
Do not intercept JVM-internal C stdlib calls 

fmeum:
The JVM frequently calls strcmp/memcmp/..., which fills up the table of
recent compares with entries that are either duplicates of values
already reported by the bytecode instrumentation or JDK-internal strings
that are not relevant for fuzzing.

This commit adds an ignorelist to the C stdlib interceptors that filters
out calls from known JVM libraries. If the fuzz target has not yet
loaded a native library, all such callbacks are ignored, which greatly
improves fuzzer performance for string-heavy targets. E.g.,
JsonSanitizerDenylistFuzzer takes < 1 million runs now when it used to
take over 3 million.

@sebastianpoeplau This includes most of the non-macOS-specific commits from #22 which you have already reviewed, only the last one is new.


sebastianpoeplau:
Looks good to me. Do we need to add special handling for MacOS though? As far as I know, `/proc/self/maps` isn't available there.

fmeum:
Thanks!

I'll make the required changes as part of the macOS PR. If I can't find a direct replacement, I might just make the ignorelist Linux-only as native library fuzzing is untested on macOS anyway.

sebastianpoeplau:
Sounds good!

sebastianpoeplau (approved):

PR #54, opened 2021-03-25
Allow reliable finding reports from hooks

fmeum:
See the individual commit messages for details.


simonresch (approved):
Nice addition! Looks good to me.

sebastianpoeplau:
I agree :) And I like the self-throwing exception - it seems simpler than the alternatives that we discussed but still sufficient in most cases.

Issue #65, opened 2021-04-07
3 tests fail on master on OSX


eustas:
Bazel report:
```
//examples:ExampleFuzzerWithNative                                       FAILED in 3.0s
//examples:ExampleStackOverflowFuzzer                                    FAILED in 2.3s
//examples:JpegImageParserFuzzer                                         FAILED in 23.0s
```

`ExampleFuzzerWithNative` and `ExampleStackOverflowFuzzer` fail with `Unexpected exit code: 139` and no other useful output.

`JpegImageParserFuzzer` really starts fuzzing, discovers `java.lang.NegativeArraySizeException` and then fails with `Unexpected exit code: 0`

fmeum:
I can reproduce all these test failures in the CI, which is why these tests are currently not run automatically for macOS.

Let me elaborate a bit on why they fail so that we can turn this issue into a tracking bug:

- `ExampleFuzzerWithNative` seems to fail purely because of Bazel's inability to create correctly named shared libraries on macOS (`.dylib` instead of `.so`). Fixing this is not a high priority as native library fuzzing should still work for user-supplied projects with correctly named native libraries.
- `JpegImageParserFuzzer` runs in libFuzzer's experimental fork mode, which uses subprocesses. Fuzzing works fine, but the main fuzzer process does not correctly pick up the exit codes of the subprocesses. Given that libFuzzer uses its [own pretty complex polyfill for the `system()` libc function](https://github.com/llvm/llvm-project/blob/62ec4ac90738a5f2d209ed28c822223e58aaaeb7/compiler-rt/lib/fuzzer/FuzzerUtilDarwin.cpp#L41) that is missing from macOS, I'm not even sure that the issue lies on our side.
- `ExampleStackOverflowFuzzer` is still a mystery to me, but JVMs on macOS have a history of surprising handling of `StackOverflowError`s. I would really like to fix this issue as it could hint at some JNI UB in Jazzer, but at least the verbose JNI mode does not produce any findings.

eustas:
Simply marking tests with tag does not prevent them from execution (I just run `bazel test ...:all`).
Perhaps it is better to have something like `BROKEN_ON_DARWIN_TAG` variable that is used as tags on broken tests. This variable itself should be assigned using selector and contains empty list by default, but is equal to `['manual']` if platform is darwin.

If you like, I can prepare PR for that.

fmeum:
> If you like, I can prepare PR for that.

[...]

PR #66, opened 2021-04-08
Disable broken tests on OSX.

eustas:
Unfortunately, this is fix not as elegant as it could be (adding "manual" tag, depending on target OS).
See #65


fmeum (approved):
Thanks for the contribution, I like this solution much more.

PR #68, opened 2021-04-11
Revert "Disable broken tests on OSX."

fmeum:
Reverts CodeIntelligenceTesting/jazzer#66

The PR breaks the IntelliJ Bazel plugin sync.

### DIFF
--- a/examples/BUILD.bazel
+++ b/examples/BUILD.bazel
@@ -27,9 +27,9 @@ java_fuzz_target_test(
         "src/main/java/com/example/ExampleFuzzerWithNative.java",
     ],
     native_libs = ["//examples/src/main/native"],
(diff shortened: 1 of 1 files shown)


fmeum:
@eustas Sorry, but I had to revert the PR as it breaks the IntelliJ Bazel plugin sync. It seems that the plugin uses a custom platform and thus fails to sync the tests. Do you happen to have an idea how this could be fixed?

eustas:
Not yet. Which version on IntelliJ you are using? Will check and see what could be done...

fmeum:
I'm on IDEA Ultimate 2020.2.2, which I believe is the latest version supported by the Bazel plugin. Thanks for looking into it!

PR #73, opened 2021-04-15
Expose trace callbacks to API

fmeum:
Exposes the traceStrstr and traceStrcmp callbacks through the Jazzer
API. Together with Jazzer#reportFindingFromHook, this allows for
stand-alone sanitizer libraries.

### DIFF
--- a/agent/src/main/java/com/code_intelligence/jazzer/api/Jazzer.java
+++ b/agent/src/main/java/com/code_intelligence/jazzer/api/Jazzer.java
@@ -14,20 +14,85 @@
 
 package com.code_intelligence.jazzer.api;
 
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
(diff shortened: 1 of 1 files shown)


sebastianpoeplau (approved):
Just out of curiosity: Is it a problem if someone calls these methods with strings that contain null characters? I suppose libfuzzer expects C strings, so the comparison will probably terminate at any null character.

fmeum:
> Just out of curiosity: Is it a problem if someone calls these methods with strings that contain null characters? I suppose libfuzzer expects C strings, so the comparison will probably terminate at any null character.

Good question, but the answer should be "no": These functions eventually call the `GetStringUTFChars` JNI method [here](https://github.com/CodeIntelligenceTesting/jazzer/blob/ba4a770952bfa696d842ff11e4a3e66751030d03/driver/libfuzzer_callbacks.cpp#L66). This returns the strings in modified UTF-8, which encodes null characters as two non-null bytes.

sebastianpoeplau:
I see, makes sense!

PR #91, opened 2021-05-04
Implement basic remote code execution sanitizers

fmeum:
The sanitizers combine mutation guides with a dynamic proof of reachability in the form of a honeypot class that raises findings whenever it is interacted with.

The suggested severities are not fixed yet as there doesn't appear to be a broad consensus on this matter. I'm open for suggestions here. Later, we would want to extend the findings with more detailed information on the type of vulnerability.

### DIFF
--- a/agent/BUILD.bazel
+++ b/agent/BUILD.bazel
@@ -1,13 +1,18 @@
 load("@rules_java//java:defs.bzl", "java_binary")
+load("//sanitizers:sanitizers.bzl", "SANITIZER_CLASSES")
 
 java_binary(
     name = "jazzer_agent",
     create_executable = False,
     deploy_manifest_lines = [
         "Premain-Class: com.code_intelligence.jazzer.agent.Agent",
+        "Jazzer-Hook-Classes: {}".format(":".join(SANITIZER_CLASSES)),
     ],
     visibility = ["//visibility:public"],
-    runtime_deps = ["//agent/src/main/java/com/code_intelligence/jazzer/agent:agent_lib"],
+    runtime_deps = [
+        "//agent/src/main/java/com/code_intelligence/jazzer/agent:agent_lib",
+        "//sanitizers",
+    ],
 )
 
 java_binary(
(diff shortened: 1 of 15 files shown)


fmeum:
I based this on #90 as that should be reviewed first.

sebastianpoeplau (approved):
Looks good! What would happen in the case of a reflective call to a method that `jaz.Zer` doesn't have? If I understand correctly, we would rely on the exception from the finalizer in that case. Couldn't we also hook into the method resolution machinery to detect when the `jaz.Zer` class is queried for any method? That would be a sign that we've been able to inject it, so we could signal a finding while still processing the right input. Maybe I'm misunderstanding something...

fmeum:
> Looks good! What would happen in the case of a reflective call to a method that `jaz.Zer` doesn't have? If I understand correctly, we would rely on the exception from the finalizer in that case. Couldn't we also hook into the method resolution machinery to detect when the `jaz.Zer` class is queried for any method? That would be a sign that we've been able to inject it, so we could signal a finding while still processing the right input. Maybe I'm misunderstanding something...

Do you mean methods like `Class#getDeclaredMethods` with "method resolution machinery"? That would be possible, but there are many such methods and we would also need to track that they are actually used. Merely looking up a method would not constitute a bug/vulnerability. 

It's also important to think about what we really want to detect: method calls on arbitrary objects that are potentially exploitable. If the code under test attempts to execute a hardcoded `Zer#verySpecificName(VerySpecificParameterType, OtherType)`, it would likely not have real-world impact as there won't be any useful gadgets an attacker could inject here.

sebastianpoeplau:
> Do you mean methods like `Class#getDeclaredMethods` with "method resolution machinery"? That would be possible, but there are many such methods and we would also need to track that they are actually used. Merely looking up a method would not constitute a bug/vulnerability.

Ah yes, requiring them to be called is an important point.

> It's also important to think about what we really want to detect: method calls on arbitrary objects that are potentially exploitable. If the code under test attempts to execute a hardcoded `Zer#verySpecificName(VerySpecificParameterType, OtherType)`, it would likely not have real-world impact as there won't be any useful gadgets an attacker could inject here.

I was thinking of an attacker who learns about that very specific method name and arranges for a malicious replacement class to be present in the system. But that would additionally require the ability to inject classes, at which point the attacker could probably just use static initializers to do their mischief. So never mind :)

PR #93, opened 2021-05-05
Ensure default -Xmx value is below -rss_limit_mb

fmeum:
libFuzzer defaults to an rss_limit_mb of 2048, but we start the JVM with
-Xmx4096m. This can lead to libFuzzer OOM reports when a single
allocation in Java exceeds 2 GB but still fits into the JVM heap.

This is solved by letting the JVM heap size default to slightly less
than 2 GB.

This change is not fully backwards compatible, but will only cause
targets to crash more often than they used to.

### DIFF
--- a/bazel/fuzz_target.bzl
+++ b/bazel/fuzz_target.bzl
@@ -63,8 +63,6 @@ def java_fuzz_target_test(
             "$(rootpath %s)" % driver,
             "--cp=$(rootpath :%s_deploy.jar)" % target_name,
             "--agent_path=$(rootpath //agent:jazzer_agent_deploy.jar)",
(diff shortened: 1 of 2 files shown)


simonresch (approved):
Do we want to adjust the JVM heap size with the `rss_limit_mb` value set by the user?

fmeum:
> Do we want to adjust the JVM heap size with the `rss_limit_mb` value set by the user?

I thought about this as well, but are users more likely to change `rss_limit_mb` or `-Xmx` via `jvm_args`? I would have guessed the latter, but am not sure. We would probably have to parse both and warn if they conflict, but that requires quite a bit of logic. What do you think?

PR #98, opened 2021-05-11
Allow instrumenting internal classes

fmeum:
Previously these would silently remain uninstrumented as they come with
a null ClassLoader that does not pass the null check.


fmeum:
@simonresch Instrumenting internal classes outside the default module (e.g., java.desktop) is tricky as of Java 9: As the agent runs in a context that is not module-aware, it can't export `CoverageMap` to other modules. This leads to `NoClassDefFound` exceptions when classes from other modules are instrumented.

In order to work around this, we may have to split the `runtime` package off and turn it into a proper Java 9 module.

simonresch (approved):
That looks like a great solution to the modules problem. Very excited about this! :)

PR #104, opened 2021-05-16
Remove dependency on java.util.Regex

fmeum:
The glob matching on class names to instrument introduced a dependency
on java.util.Regex, which results in two problems:

* Since the logic required to determine that java.util.Regex should not
  be instrumented requires java.util.Regex, we obtain
  ClassCircularityErrors. While harmless, they pollute the logs and
  we don't have a good way to filter them out.
* Jazzer can't be used to fuzz java.util.Regex.

Since we don't need the full power of globs, this commit adds a very
simple glob matcher that only supports the cases foo.bar, foo.* and
foo.**.


simonresch (approved):
Looks good to me.
The two glob patterns should be sufficient.

PR #120, opened 2021-05-27
Pass quoted arguments to child processes

fmeum:
libFuzzer does not quote the arguments it passes to child processes during merge
and fork, which leads to arguments being lost if passing multiple jvm_args with
delimiter ';'.

This commit adds a libFuzzer patch that properly quotes all arguments as well as
a test that fails if quoting is not appropriate.

### DIFF
--- a/examples/src/main/java/com/example/JpegImageParserFuzzer.java
+++ b/examples/src/main/java/com/example/JpegImageParserFuzzer.java
@@ -22,6 +22,16 @@ import org.apache.commons.imaging.formats.jpeg.JpegImageParser;
 
 // Found https://issues.apache.org/jira/browse/IMAGING-275.
 public class JpegImageParserFuzzer {
+  public static void fuzzerInitialize() {
+    // Only used to verify that arguments are correctly passed down to child processes.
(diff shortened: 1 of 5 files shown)


examples/src/main/java/com/example/JpegImageParserFuzzer.java
```diff
@@ -22,6 +22,16 @@
 
 // Found https://issues.apache.org/jira/browse/IMAGING-275.
 public class JpegImageParserFuzzer {
+  public static void fuzzerInitialize() {
+    // Only used to verify that arguments are correctly passed down to child processes.
+    if (System.getProperty("foo") == null || System.getProperty("bar") == null) {
+      // The child process did not have a sufficiently high memory limit,
```
simonresch, line 28: Is this comment relevant here?

simonresch (commented):
LGTM

examples/src/main/java/com/example/JpegImageParserFuzzer.java
fmeum: Good catch, it's a left over from a previous version of this check. I will remove it.

PR #124, opened 2021-05-30
Fix Jazzer agent lookup

fmeum:
If Jazzer is imported as an external workspace from another Bazel
workspace, the runfiles path of the agent is
../jazzer/agent/jazzer_agent_deploy.jar rather than
agent/jazzer_agent_deploy.jar. Since the first path applies more
generally, we switch to it, but perform a check for a Bazel env variable
to prevent loading agents from potentially untrusted sibling dirs.


simonresch (approved):
Good fix. Thanks!

PR #125, opened 2021-05-30
Close sync file instead of just releasing the lock

fmeum:
Speculative fix for synchronization failures observed on ClusterFuzz.


fmeum:
@sebastianpoeplau If you have any ideas on how to improve the logging in the failures case, please let me know. We don't need to fix this urgently, but I would really like to understand the root cause so that we can provide a full fix.

sebastianpoeplau (approved):

PR #136, opened 2021-06-11
Allow dumping instrumented classes

fmeum:
With `--dump_classes_dir=<dir>`, all classes that are instrumented by the
agent at runtime will be dumped into a subdirectory of `<dir>` according
to their internal class name.


stheid:
image
Seems not to work. The option is also not listed in the --help

fmeum:
@stheid You are building `bazel-bin/jazzer_release.tar.gz`, but then you `cd` into `bazel-bin/jazzer_release`, which may be outdated. Try extracting the archive somewhere and run the resulting `jazzer` script.

stheid:
I trusted in `git reset` but you are absolutely right. I worked on old files...

PR #143, opened 2021-06-28
Register native callbacks on ClassInstrumentor initialization

fmeum:
The native callbacks in the driver should be registered as early as
possible since they may already be needed during agent startup (e.g.,
if a JDK-internal class is to be instrumented).


fmeum:
@simonresch I fixed the linker issues on macOS, this is ready for review now.

JaggaJax (approved):
Looks good to me!

PR #146, opened 2021-07-08
Use a hermetic LLVM toolchain

fmeum:


fmeum:
@sebastianpoeplau Do you have an idea why asan is still linked dynamically on macOS even with `-static-libsan`? This is what causes the pipeline to fail as the toolchain does not export the dynamic versions of the sanitizer runtime libraries.

sebastianpoeplau:
I was initially thinking [this](https://github.com/llvm/llvm-project/blob/4504e1134c9118f3c322685f8a90129e09bab92c/clang/lib/Driver/ToolChains/CommonArgs.cpp#L829) could be the problem. But it doesn't explain why the dynamic object is still requested. Do you have the full compiler invocation? Maybe the flag doesn't end up there the way we expect.

fmeum:
The full failing command is:

```
exec env - \
    CC=clang \
    PATH=/bin:/usr/bin:/usr/local/bin \
    *** \
  external/llvm_toolchain/bin/cc_wrapper.sh -shared -o bazel-out/darwin-opt/bin/examples/src/main/native/libnative_asan.so bazel-out/darwin-opt/bin/examples/src/main/native/_objs/native_asan/com_example_ExampleFuzzerWithNative.pic.o '-fsanitize=fuzzer-no-link,address' -static-libsan -lm -no-canonical-prefixes -lc++ -lc++abi -headerpad_max_install_names -undefined dynamic_lookup '--sysroot=/Applications/Xcode_12.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk'
```

(see https://github.com/CodeIntelligenceTesting/jazzer/runs/3113361833#step:6:58)

I started another run with explicit `-fno-sanitize-link-runtime` linkopts on the DSOs.

fmeum:
Still failing, even if not linking the runtime is requested explicitly.

sebastianpoeplau:
Strange, and the `-static-libsan` flag did make it to the command line...

fmeum:
@sebastianpoeplau I missed that we need `-static-libsan` on the *driver*, not the native library dependencies. I fixed this and will see how the next run goes.

fmeum:
Well, that's fun: asan is linked dynamically, ubsan is linked statically in https://github.com/CodeIntelligenceTesting/jazzer/runs/3240624828?check_suite_focus=true.

fmeum:
@sebastianpoeplau This PR finally works and is ready for review. I learned more about ASan runtimes than I had planned though.

Note that the first commit is #160, which I intend to merge first.

sebastianpoeplau (approved):

PR #147, opened 2021-07-09
Use libc++ for building Jazzer

fmeum:
This works around incomplete C++17 support in GCC.


fmeum:
Blocked on https://github.com/bazelbuild/bazel/pull/13666.

PR #160, opened 2021-08-06
Find libjvm with a repository rule

fmeum:
libjvm lives in different subpaths of JAVA_HOME, depending both on the
OS and the Java version. Since it is currently not possible to select a
dependency based on the Java version, supporting Java 8 required a
custom build setting. This also broke bazel query (but not cquery).

By loading libjvm from a simple repository rule, we can cover all
OSes and Java versions with a single dependency, even if libjvm.so is
installed in a non-standard location.


bazel/local_jdk_libjvm.bzl
```diff
@@ -0,0 +1,60 @@
...
+    ])
+    if result.return_code != 0:
+        return None
+    file_path = result.stdout.strip()
+    if not file_path:
+        return None
+    return repository_ctx.path(file_path)
+
+LIBJVM_NAMES = [
+    "libjvm.dll",
```
simonresch, line 32: Could this rule work on Windows? The find command above and the unix path separators below would probably need some adjustments for Windows, no? Maybe we can drop `libjvm.dll` since we don't support Windows?

simonresch (approved):
Very cool! Looks good to me.

bazel/local_jdk_libjvm.bzl
fmeum: You are right, having it here doesn't make much sense without a `cmd` command to find it.

PR #162, opened 2021-08-06
Fix reproducer template

fmeum:
The Java reproducer used `getMethod` incorrectly and thus did not call
fuzzerInitialize(String[]).


simonresch (approved):
Looks good!

PR #175, opened 2021-08-20
Add Jsoup Findings to README

0roman:


fmeum:
@0roman Could you move the new findings to the top of the table? I think that it makes sense to have the most recent CVEs come first in the table.

0roman:
@fmeum done.

PR #183, opened 2021-09-13
Report coverage for all classes on the classpath 

fmeum:
Using the (very fast) classpath traverser ClassPath, we can generate
coverage data for *all* classes on the classpath rather than just those
that were loaded during the fuzzing run.


simonresch (approved):
Very cool addition!
Looks good to me.

PR #197, opened 2021-10-13
Pre-Windows improvements

fmeum:
A collection of fixes and improvements required for Windows support that also make sense without it.


driver/libfuzzer_driver.cpp
```diff
@@ -44,27 +42,20 @@ DECLARE_string(id_sync_file);
...
-// Suppress libFuzzer warnings about missing sanitizer methods in non-ASan
-// builds.
-extern "C" __attribute__((weak)) int __sanitizer_acquire_crash_state() {
-  return true;
-}
-
-extern "C" __attribute__((weak)) void __sanitizer_print_stack_trace() {
-  jazzer::DumpJvmStackTraces();
-}
+#ifndef _WIN32
```
kyakdan, line 57: Why do we need this empty preprocessor directive?

PR #200, opened 2021-10-15
Windows support

fmeum:
All targets build and only `//examples:ExampleFuzzerWithUBSan` fails at runtime (and is thus labelled as incompatible).

### DIFF
--- a/examples/BUILD.bazel
+++ b/examples/BUILD.bazel
@@ -183,6 +183,7 @@ sh_test(
         ":JsonSanitizerDenylistCrash",
         ":json_sanitizer_denylist_crash",
     ],
+    target_compatible_with = SKIP_ON_WINDOWS,
 )
 
 java_fuzz_target_test(


driver/sanitizer_hooks_with_pc.cpp
```diff
@@ -36,6 +36,20 @@
 
 #define REPEAT_4096(a) REPEAT_16(REPEAT_16(REPEAT_16(a)))
 
+// The first four registers to pass arguments in according to the
```
kyakdan, line 39: Should we mention in the docs that we only support X64 platforms?

driver/BUILD.bazel
```diff
@@ -129,11 +129,16 @@ cc_binary(
...
-        "@platforms//os:windows": [],
+        "@platforms//os:windows": [
+            # Sanitizer runtimes have to be linked manually on Windows:
+            # https://devblogs.microsoft.com/cppblog/addresssanitizer-asan-for-windows-with-msvc/
+            "/wholearchive:clang_rt.asan-x86_64.lib",
+            "/wholearchive:clang_rt.asan_cxx-x86_64.lib",
+        ],
         "//conditions:default": [
+            "-fsanitize=address",
+            "-static-libsan",
```
kyakdan, line 141: isn't this the default behavior for clang on Linux?

driver/sanitizer_hooks_with_pc.cpp
fmeum: We currently say `Jazzer supports x64 Linux and (experimentally) macOS 10.15.`. But good point, we should update this for Windows.

driver/BUILD.bazel
fmeum: It is on Linux, but not on macOS. Since it may change on Linux in the future (albeit unlikely), I found having this here explicitly to be safer.

.bazelrc
```diff
@@ -23,6 +23,11 @@ build --tool_java_language_version=9
 build:windows --extra_toolchains=@local_config_cc//:cc-toolchain-x64_windows-clang-cl
 build:windows --extra_execution_platforms=//:x64_windows-clang-cl
 build:windows --features=static_link_msvcrt
+# Required so that the JDK can find itself when loading the instrument agent lib.
+# FIXME: Debug why this is the case even though sun.boot.library.path is correctly set to
+#        external\local_jdk\bin under the Bazel cache.
+test:windows --noincompatible_strict_action_env
```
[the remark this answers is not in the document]
fmeum: That is true, but if that happens we will simply have to maintain our own value for `PATH`. Note that the fix here is to *disable* this option which we enabled above, so it will continue to work after the flag has been removed.

driver/jvm_tooling.cpp
```diff
@@ -32,9 +32,8 @@
 #include "utils.h"
 
 DEFINE_string(cp, ".",
-              "the jvm class path which should include the fuzz target class, "
-              "instrumentor "
-              "runtime and further dependencies separated by a colon \":\"");
+              "the classpath to use for fuzzing. Behaves analogously to java's "
```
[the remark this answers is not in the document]
fmeum: Will do.

driver/BUILD.bazel
kyakdan: agreed

kyakdan (approved):

PR #203, opened 2021-10-18
Add sanitizer for expression language injection

simonresch:


sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ExpressionLanguageInjection.kt
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * Detects injectable inputs to an expression language interpreter which may lead to remote code execution.
+ */
+@Suppress("unused_parameter")
+object ExpressionLanguageInjection {
+
+    /**
+     * jaz.Zer is a honeypot class: All of its methods report a finding when called.
+     */
+    private const val HONEYPOT_CLASS_NAME = "jaz.Zer"
```
kyakdan, line 33: This constant is duplicated in all of the three sanitizers. Does it make sense to extract this constant so that any future changes become easier and less error prone?

sanitizers/src/main/java/jaz/Zer.java
```diff
@@ -29,6 +29,9 @@
  */
 @SuppressWarnings("unused")
 public class Zer implements java.io.Serializable {
+
+  public static final String HONEYPOT_CLASS_NAME = "jaz.Zer";
+
```
fmeum, line 34: This is dangerous: We should not reference `jaz.Zer` from anywhere explicitly as it might crash or confuse the sanitizers. Could we move this constant into a Utils file in the sanitizers package?

fmeum (changes requested):

fmeum (approved):

PR #205, opened 2021-10-19
Autofuzz

fmeum:
@kyakdan I used `git revise` and thus shouldn't have lost any changes during the rebase. I tried to merge commits with their follow-up fix up commits to get a cleaner history, but stopped whenever I would have had to solve any merge conflicts.

I also removed all mentions of the `autofuzz` branch - after the merge, everything should again live in `main`.

### DIFF
--- a/agent/src/main/java/com/code_intelligence/jazzer/autofuzz/Meta.java
+++ b/agent/src/main/java/com/code_intelligence/jazzer/autofuzz/Meta.java
@@ -18,7 +18,6 @@ import com.code_intelligence.jazzer.api.FuzzedDataProvider;
 import io.github.classgraph.ClassGraph;
 import io.github.classgraph.ClassInfoList;
 import io.github.classgraph.ScanResult;
-import java.beans.PropertyDescriptor;
 import java.io.ByteArrayInputStream;
 import java.lang.reflect.Array;
 import java.lang.reflect.Constructor;
@@ -32,7 +31,10 @@ import net.jodah.typetools.TypeResolver;
 import net.jodah.typetools.TypeResolver.Unknown;
 
 public class Meta {
-  static WeakHashMap<Class<?>, List<Class<?>>> cache = new WeakHashMap<>();
+  static WeakHashMap<Class<?>, List<Class<?>>> implementingClassesCache = new WeakHashMap<>();
+  static WeakHashMap<Class<?>, List<Class<?>>> nestedBuilderClassesCache = new WeakHashMap<>();
+  static WeakHashMap<Class<?>, List<Method>> originalObjectCreationMethodsCache =
+      new WeakHashMap<>();
 
   public static Object autofuzz(FuzzedDataProvider data, Method method) {
     if (Modifier.isStatic(method.getModifiers())) {
@@ -135,7 +137,7 @@ public class Meta {
     } else if (type.isEnum()) {
       return data.pickValue(type.getEnumConstants());
     } else if (type.isInterface() || Modifier.isAbstract(type.getModifiers())) {
-      List<Class<?>> implementingClasses = cache.get(type);
+      List<Class<?>> implementingClasses = implementingClassesCache.get(type);
       if (implementingClasses == null) {
         try (ScanResult result =
                  new ClassGraph().enableClassInfo().enableInterClassDependencies().scan()) {
@@ -143,7 +145,7 @@ public class Meta {
               type.isInterface() ? result.getClassesImplementing(type) : result.getSubclasses(type);
           implementingClasses =
               children.getStandardClasses().filter(cls -> !cls.isAbstract()).loadClasses();
-          cache.put(type, implementingClasses);
+          implementingClassesCache.put(type, implementingClasses);
         }
       }
       if (implementingClasses.isEmpty()) {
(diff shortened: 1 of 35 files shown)


agent/src/main/java/com/code_intelligence/jazzer/autofuzz/Meta.java
```diff
@@ -0,0 +1,337 @@
...
+        if (isDebug()) {
+          throw new AutofuzzConstructionException(String.format(
+              "Could not find classes implementing %s on the classpath", type.getName()));
+        } else {
+          throw new AutofuzzConstructionException();
+        }
+      }
+      return consume(data, data.pickValue(implementingClasses));
+    } else if (type.getConstructors().length > 0) {
+      Constructor<?> constructor = data.pickValue(type.getConstructors());
```
kyakdan, line 177: Does the constructor list returned by `type.getConstructors()` have the same order in different calls? I had this issue with `type.getMethods()` and had to sort the returned list so that we don't produce bugs that cannot be reproduced with the reported crashing input.

agent/src/main/java/com/code_intelligence/jazzer/autofuzz/Meta.java
```diff
@@ -0,0 +1,337 @@
...
+                                                 .collect(Collectors.toList());
+
+      List<Method> originalObjectCreationMethods = getOriginalObjectCreationMethods(pickedBuilder);
+
+      int pickedMethodsNumber = data.consumeInt(0, cascadingBuilderMethods.size());
+      List<Method> pickedMethods = data.pickValues(cascadingBuilderMethods, pickedMethodsNumber);
+
+      Method builderMethod = data.pickValue(originalObjectCreationMethods);
+
+      Object builderObj = autofuzz(data, data.pickValue(pickedBuilder.getConstructors()));
```
kyakdan, line 211: Should we check here for returned 'null' objects?

agent/src/main/java/com/code_intelligence/jazzer/autofuzz/Meta.java
```diff
@@ -0,0 +1,337 @@
...
+  }
+
+  static boolean isDebug() {
+    String value = System.getenv("JAZZER_AUTOFUZZ_DEBUG");
+    return value != null && !value.isEmpty();
+  }
+
+  private static String getDebugSummary(
+      Executable executable, Object thisObject, Object[] arguments) {
+    return String.format("%nMethod: %s::%s%s%nthis: %s%nArguments: %s",
```
kyakdan, line 251: Should we add a prefix to the debug summary to make clear that it comes from Jazzer's autofuzz mode so that they become easier to spot?

agent/src/main/java/com/code_intelligence/jazzer/autofuzz/Meta.java
```diff
@@ -0,0 +1,337 @@
...
+      return data.pickValue(type.getEnumConstants());
+    } else if (type == Class.class) {
+      return YourAverageJavaClass.class;
+    } else if (type == Method.class) {
+      return data.pickValue(YourAverageJavaClass.class.getMethods());
+    } else if (type == Constructor.class) {
+      return data.pickValue(YourAverageJavaClass.class.getConstructors());
+    } else if (type.isInterface() || Modifier.isAbstract(type.getModifiers())) {
+      List<Class<?>> implementingClasses = implementingClassesCache.get(type);
+      if (implementingClasses == null) {
```
kyakdan, line 152: I'm wondering if this would work when classes are loaded dynamically during execution

driver/fuzz_target_runner.cpp
```diff
@@ -42,9 +43,9 @@ DEFINE_string(target_args, "",
               "Arguments passed to fuzzerInitialize as a String array. "
               "Separated by space.");
 
-DEFINE_uint32(keep_going, 1,
+DEFINE_uint32(keep_going, 0,
               "Continue fuzzing until N distinct exception stack traces have"
-              "been encountered");
+              "been encountered. Defaults to exit after the first finding.");
```
kyakdan, line 48: Not correct for the autofuzz mode, right?

agent/src/main/java/com/code_intelligence/jazzer/autofuzz/YourAverageJavaClass.java
```diff
@@ -0,0 +1,227 @@
+// Copyright 2021 Code Intelligence GmbH
```
kyakdan, line 1: Can you mention the rational behind adding this class?

agent/src/main/java/com/code_intelligence/jazzer/autofuzz/Meta.java
fmeum: Good point, I will sort the output of every method of this kind. I would be surprised if the return value wouldn't be the same across fuzzing runs, but of course it might change when code is recompiled or just across different versions of the JVM.

agent/src/main/java/com/code_intelligence/jazzer/autofuzz/Meta.java
fmeum: See above, since this calls through to `Constructor.newInstance`, I believe it can never return `null`.

agent/src/main/java/com/code_intelligence/jazzer/autofuzz/YourAverageJavaClass.java
fmeum: I tested gson and found that we weren't dealing very well with signatures like `deserialize(byte[] data, Class<?> klazz)`. Special casing `Class`, `Method` and `Constructor` to an "interesting" class dramatically increases the coverage here.

PR #215, opened 2021-10-19
Autofuzz: allows globbing for exceptions to ignore

kyakdan:
Users can choose to ignore all exceptions in a specific package, i.g., `java.lang.*` or `java.lang.Nu*`


agent/src/main/java/com/code_intelligence/jazzer/autofuzz/FuzzTarget.java
```diff
@@ -122,19 +123,20 @@ public static void fuzzerInitialize(String[] args) {
...
-                System.exit(1);
-              }
-              throw new Error("Not reached");
-            })
-            .collect(Collectors.toList());
+
+    Set<Class<?>> alwaysIgnore = new HashSet<>();
+    ClassGraph classGraph = new ClassGraph().enableClassInfo().enableSystemJarsAndModules();
+    try (ScanResult result = classGraph.scan()) {
+      List<Class<?>> allExceptions = result.getSubclasses(Throwable.class).loadClasses();
```
fmeum, line 130: I think we could get away with loading only the classes that match the pattern.

agent/src/main/java/com/code_intelligence/jazzer/autofuzz/FuzzTarget.java
```diff
@@ -122,19 +123,20 @@ public static void fuzzerInitialize(String[] args) {
...
-              } catch (ClassNotFoundException e) {
-                System.err.printf("Failed to find class '%s' specified in --autofuzz_ignore", name);
-                System.exit(1);
-              }
-              throw new Error("Not reached");
-            })
-            .collect(Collectors.toList());
+
+    Set<Class<?>> alwaysIgnore = new HashSet<>();
+    ClassGraph classGraph = new ClassGraph().enableClassInfo().enableSystemJarsAndModules();
```
fmeum, line 128: Could we avoid this potentially very slow search if the arguments are not regexes (e.g. contain only `[a-zA-Z0-9.]` + Unicode)?

agent/src/main/java/com/code_intelligence/jazzer/autofuzz/FuzzTarget.java
```diff
@@ -122,19 +123,20 @@ public static void fuzzerInitialize(String[] args) {
...
+    try (ScanResult result = classGraph.scan()) {
+      List<Class<?>> allExceptions = result.getSubclasses(Throwable.class).loadClasses();
+      List<Pattern> patterns =
+          Arrays.stream(args).skip(1).map(Pattern::compile).collect(Collectors.toList());
```
fmeum, line 132: Note that most Java class names contain '.', which has a special meaning here. Not sure how to handle this in the best way, but maybe we could just extract and reuse `com.code_intelligence.jazzer.instrumentor.SimpleGlobMatcher` here instead of using full-blown regexes?

kyakdan: Changed the logic to use `SimpleGlobMatcher`

agent/src/main/java/com/code_intelligence/jazzer/autofuzz/FuzzTarget.java
kyakdan: Changed based on our discussion to a simpler globbing logic

PR #222, opened 2021-10-21
Make Autofuzz accessible from the Jazzer API

fmeum:


agent/src/test/java/com/code_intelligence/jazzer/api/AutofuzzTest.java
```diff
@@ -42,4 +59,49 @@ public void testConsumeFailsWithoutException() {
...
+            (byte) 5, (byte) 1 /* do not return null */, 0 /* first class on the classpath */,
+            (byte) 1 /* do not return null */, 0 /* first constructor */));
+    Jazzer.autofuzz(data, AutofuzzTest::checkAllTheArguments);
+  }
+
+  @Test
+  public void testAutofuzzConsumerThrowsException() {
+    FuzzedDataProvider data =
+        CannedFuzzedDataProvider.create(Arrays.asList((byte) 1 /* do not return null */,
+            6 /* string length */, "foobar", 42, (byte) 5, (byte) 0 /* *do* return null */));
```
kyakdan, line 98: the last 0 is to return `null`, right?

fmeum: Yes, that is also why I left the comment with an emphasis on `do`. Is that clear enough?

PR #230, opened 2021-10-29
Implement code generation for consume and autofuzz

fmeum:
Nested builders, setters and Method/Constructor are not yet implemented.

### DIFF
--- a/agent/src/main/java/com/code_intelligence/jazzer/autofuzz/AutofuzzCodegenVisitor.java
+++ b/agent/src/main/java/com/code_intelligence/jazzer/autofuzz/AutofuzzCodegenVisitor.java
@@ -1,0 +1,117 @@
+// Copyright 2021 Code Intelligence GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.code_intelligence.jazzer.autofuzz;
+
+import java.util.Stack;
+import java.util.stream.Collectors;
+
+public class AutofuzzCodegenVisitor {
+  private final Stack<Group> groups = new Stack<>();
+  private int variableCounter = 0;
+
+  AutofuzzCodegenVisitor() {
+    init();
+  }
+
+  private void init() {
+    pushGroup("", "", "");
+  }
+
+  public void pushGroup(String prefix, String delimiter, String suffix) {
+    groups.push(new Group(prefix, delimiter, suffix));
+  }
+
+  public void pushElement(String element) {
+    groups.peek().push(element);
+  }
+
+  public void popElement() {
+    groups.peek().pop();
+  }
+
+  public void popGroup() {
+    if (groups.size() == 1) {
+      throw new AutofuzzError(
+          "popGroup must be called exactly once for every pushGroup: " + toDebugString());
+    }
+    pushElement(groups.pop().toString());
+  }
+
+  public String generate() {
+    if (groups.size() != 1) {
+      throw new AutofuzzError(
+          "popGroup must be called exactly once for every pushGroup: " + toDebugString());
+    }
+    return groups.pop().toString();
+  }
+
+  public void addCharLiteral(char c) {
+    pushElement("'" + escapeForLiteral(Character.toString(c)) + "'");
+  }
+
+  public void addStringLiteral(String string) {
+    pushElement('"' + escapeForLiteral(string) + '"');
+  }
+
+  public String uniqueVariableName() {
+    return String.format("autofuzzVariable%s", variableCounter++);
+  }
+
+  private String escapeForLiteral(String string) {
+    // The list of escape sequences is taken from:
+    // https://docs.oracle.com/javase/tutorial/java/data/characters.html
+    return string.replace("\t", "\\t")
+        .replace("\b", "\\b")
+        .replace("\n", "\\n")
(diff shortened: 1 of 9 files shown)


agent/src/main/java/com/code_intelligence/jazzer/autofuzz/AutofuzzCodegenVisitor.java
```diff
@@ -0,0 +1,112 @@
...
+  public void popGroup() {
+    if (groups.size() == 1) {
+      throw new AutofuzzError("popGroup must be called exactly once for every pushGroup");
+    }
+    pushElement(groups.pop().toString());
+  }
+
+  public String generate() {
+    if (groups.size() != 1) {
+      throw new AutofuzzError("popGroup must be called exactly once for every pushGroup");
```
kyakdan, line 53: @fmeum I've just started with the review and give the new functionality a try with commons imaging (` --autofuzz="org.apache.commons.imaging.Imaging::getBufferedImage"`) and this triggers this exception in some runs.

fmeum: I pushed two new commits that should fix this. I let it run for a while it did not encounter any more `AutofuzzError`s.

agent/src/main/java/com/code_intelligence/jazzer/autofuzz/Meta.java
```diff
@@ -161,46 +207,112 @@ public static Object autofuzz(FuzzedDataProvider data, Method method, Object thi
...
 
   public static Object consume(FuzzedDataProvider data, Class<?> type) {
+    return consume(data, type, null);
+  }
+
+  static Object consume(FuzzedDataProvider data, Class<?> type, AutofuzzCodegenVisitor visitor) {
     if (type == byte.class || type == Byte.class) {
-      return data.consumeByte();
+      byte result = data.consumeByte();
+      if (visitor != null)
```
kyakdan, line 216: Do you have an idea about the performance overhead caused by the if statements when `visitor` is `null`?

agent/src/main/java/com/code_intelligence/jazzer/autofuzz/FuzzTarget.java
```diff
@@ -187,11 +191,14 @@ public static void fuzzerTestOneInput(FuzzedDataProvider data) throws Throwable
...
       if (targetExecutable instanceof Method) {
-        returnValue = Meta.autofuzz(data, (Method) targetExecutable);
+        returnValue = Meta.autofuzz(data, (Method) targetExecutable, codegenVisitor);
       } else {
-        returnValue = Meta.autofuzz(data, (Constructor<?>) targetExecutable);
+        returnValue = Meta.autofuzz(data, (Constructor<?>) targetExecutable, codegenVisitor);
       }
       executionsSinceLastInvocation = 0;
+      if (codegenVisitor != null) {
+        System.err.println(codegenVisitor.generate());
```
kyakdan, line 200: In order to help users creating a reproducer even when fuzzing without setting `JAZZER_AUTOFUZZ_DEBUG`, we can show a command how to create it using the crashing input, e.g., `JAZZER_AUTOFUZZ_DEBUG=1 ./jazzer ... <crashing input>`. WDYT?

fmeum: I would prefer to keep `JAZZER_AUTOFUZZ_DEBUG` an internal flag that is only used to debug Autofuzz, otherwise it will just become a feature people depend on. I don't think that implementing a proper reproducer generator will be a difficult task once this PR has been merged and I would certainly work on this next.

agent/src/main/java/com/code_intelligence/jazzer/autofuzz/Meta.java
```diff
@@ -52,19 +53,51 @@
...
+        // This group will always have two elements: The thisObject and the method call.
+        visitor.pushGroup("", ".", "");
+      }
+      Object thisObject = consume(data, method.getDeclaringClass(), visitor);
       if (thisObject == null) {
         throw new AutofuzzConstructionException();
       }
-      return autofuzz(data, method, thisObject);
+      result = autofuzz(data, method, thisObject, visitor);
+      if (visitor != null) {
```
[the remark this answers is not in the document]
fmeum: That's true, but would you find this more readable? The `pushGroup` and `popGroup` calls always appear in pairs that bracket one or more autofuzz or consume invocations and moving the `popGroup` call out of the top-level `if` would potentially obscure this structure.

agent/src/main/java/com/code_intelligence/jazzer/autofuzz/Meta.java
fmeum: I expect it to be negligible for two reasons:
1. The entire method consists of a chain of `if` statements and these calls only every add a constant number per `consume` call. Most branches do far more work than this `if`.
2. Since the branch structure is pretty simple and it is a simple `null` check, I would expect this to JIT compile down to a single compare instruction that is never taken, which should be virtually free due to branch prediction.

Once Autofuzz has become a bit more mature, we should certainly profile it thoroughly though and revisit these points.

kyakdan: OK, then let's revisit this point later

agent/src/main/java/com/code_intelligence/jazzer/autofuzz/FuzzTarget.java
kyakdan: My main point here was that even though the capability of generating code is implemented, the user workflow of generating a reproducer is still not completely there yet. Since you plan this as a follow up work, I'm fine with it.

agent/src/main/java/com/code_intelligence/jazzer/autofuzz/Meta.java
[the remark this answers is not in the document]
kyakdan: OK, then let's keep this pattern consistent with the way it is done in other places in the code

kyakdan (approved):

PR #236, opened 2021-11-10
Extract JavaNoThrowMethods into a JaCoCo fork

fmeum:
The fork of JaCoCo is compatible with the exec files generated by
Jazzer's dumpCoverage.

This also reduces the size of the patches maintained in this repo.


simonresch (approved):
The changes look good to me!

Do you think it would be valuable to offer a JaCoCo release for our fork? Currently users would have to compile it themselves to make use of the exec files, correct?

fmeum:
Thanks. That's going to be the next step, I only need to figure out how to create a release of the agent. When it is ready, I will link it here.

PR #243, opened 2021-11-22
Make trampoline smaller

fmeum:
Since libFuzzer's integer compare hooks only feed the lowest 9 bits of
the address of the compare instruction into the value profile map, we
can reduce the size of the trampoline from 2^12 to 2^9.


kyakdan (approved):

kyakdan (approved):
One minor remark. We should change the comment in the header file mentioning that we only use the lower 12 bits of the caller_pc

fmeum:
> One minor remark. We should change the comment in the header file mentioning that we only use the lower 12 bits of the caller_pc

Fixed, thanks.

PR #246, opened 2021-11-29
Make crash-reproducers compile properly

centic9:
There is a nested catch in the crash-template which uses the 
same name "ignored" for the caught Exception as the catch on 
the outer level. 

This does not compile in Java and thus currently requires changes to crash-reproducers.

When using a unique name, this is resolved.


fmeum:
Thanks for the fix, this looks very reasonable.

fmeum (approved):

PR #248, opened 2021-11-30
Disable GEP instrumentation by default

fmeum:
@kyakdan What do you think about giving this a try on OSS-Fuzz this way?


kyakdan:
@fmeum I'll be in favor of disabling GEP instrumentation by default and giving it a try on OSS-Fuzz. Have you verified the effect with existing examples?

agent/src/main/java/com/code_intelligence/jazzer/agent/Agent.kt
```diff
@@ -95,7 +95,11 @@ fun premain(agentArgs: String?, instrumentation: Instrumentation) {
             "gep" -> setOf(InstrumentationType.GEP)
             "indir" -> setOf(InstrumentationType.INDIR)
             "native" -> setOf(InstrumentationType.NATIVE)
-            "all" -> InstrumentationType.values().toSet()
+            // Disable GEP instrumentation by default as it appears to negatively affect fuzzing
+            // performance. Our current GEP instrumentation only reports constant indices, but even
+            // when we instead reported non-constant indices, they tended to completely fill up the
+            // table of recent compares and value profile map.
+            "all" -> InstrumentationType.values().toSet() - InstrumentationType.GEP
```
kyakdan, line 102: Have you noticed a difference in the performance when trying that with value profiling?

kyakdan (approved):

agent/src/main/java/com/code_intelligence/jazzer/agent/Agent.kt
fmeum: In my very limited experiments, the time to first bug was reduced slightly by disabling GEP, but I don't think they were particularly representative.

PR #250, opened 2021-12-02
consumeChar fix

### LINKED ISSUE #249: FuzzedDataProvider.consumeChar(char min, char max) returns values outside of range
Calls to `FuzzedDataProvider.consumeChar(char min, char max) ` can return values outside of the range [min, max]. For example, fuzzing the following method with Jazzer produces an unexpected assertion error:
```
import com.code_intelligence.jazzer.api.FuzzedDataProvider;

public class FuzzTarget {
    public static void fuzzerTestOneInput(FuzzedDataProvider provider) {
    	char min = 'a';
    	char max = 'b';
    	char c = provider.consumeChar(min, max);
    	assert(c >= min && c <= max);
    }
}
```

I believe that the issue is that line 667 of fuzzed_data_provider.cpp uses a function pointer to `ConsumeIntegral ` instead of `ConsumeIntegralInRange`.

katherine-hough:
Changed function for consumeChar(char min, char max) to ConsumeIntegralInRange from ConsumeIntegral. 
Fixes #249


fmeum (approved):

fmeum:
@katherine-hough Thanks! I ran the formatting script to fix the CI failure.

fmeum:
The CI failure on macOS is certainly unexpected and doesn't seem to have anything to do with this particular PR. I will investigate.

PR #254, opened 2021-12-08
Port the hook trampoline to arm64

fmeum:


fmeum:
@sebastianpoeplau The implementation is not working yet, it segfaults in the function we jump to with `BR`. I just wanted to share this early draft with you in case you see any glaring mistakes on my part. If you want, we could debug this together on Friday.

fmeum:
I had to preserve `x30` and use an absolute offset instead of a label in the `adr` instruction.

I confirmed manually that with this PR, both `//driver:sanitizer_hooks_with_pc_test` and `//examples:ExampleValueProfileFuzzer` pass on an M1 Mac. Without the trampoline, the fuzzer is not able to get past the second `if` in the latter test.

@sebastianpoeplau I don't think there is a way to get the lowest two bits of the return address to be meaningful: Not only are all instructions four bytes long, they also have to be aligned to four bytes. Jumping to an offset that is not 0 mod 4 results in a `SIGBUS`. Do you think we would have a chance to get a modified `GET_CALLER_PC` upstreamed into libFuzzer?

sebastianpoeplau:
> @sebastianpoeplau I don't think there is a way to get the lowest two bits of the return address to be meaningful: Not only are all instructions four bytes long, they also have to be aligned to four bytes. Jumping to an offset that is not 0 mod 4 results in a `SIGBUS`. Do you think we would have a chance to get a modified `GET_CALLER_PC` upstreamed into libFuzzer?

I agree that libfuzzer appears to lose precision on ARM64 by including the lowest two bits of the return address. Maybe we could indeed discuss the issue with the libfuzzer devs - one could, for example, reduce the shift in `TraceCmp` from 7 to 5 bits on ARM64, given that the lowest two bits will always be zero :thinking:

sebastianpoeplau (approved):
Sorry for the long delay - looks great!

PR #257, opened 2021-12-10
Add a sanitizer for javax.naming.Context#lookup

fmeum:


sebastianpoeplau:
Following up on our discussion, I've checked the changes since then - looks great :blush:

PR #277, opened 2022-01-08
Honor JAVA_OPTS

fmeum:
The only mechanism to specify JVM arguments currently supported by Jazzer is the `--jvm_args` argument, which has a syntax that differs between OSes and may require shell quoting.

This commit makes Jazzer honor the `JAVA_OPTS` environment variable, which is used by many popular Java tools to specify arguments to the JVMs they invoke.

Related to #279.


bertschneider (approved):
Looks good to me. No further comments.

PR #280, opened 2022-01-10
Replace hardcoded repository name in rlocation paths

fmeum:
This is in preparation for Bazel module support, where repositories are
assigned internal names different from their declared user-facing name.


fmeum:
The tests don't pass from an external repository. `rootpath` still evaluates to a path under `external` rather than `../`, so this needs special handling.

fmeum:
Waiting for https://github.com/bazelbuild/proposals/pull/269 instead.

Issue #269, opened 2022-01-03
Constant string generated for crash file is too long


cdanielgomes:
Hi there.

I am generating 2D matrix with random dimensions like the code below. 

```java
            int vectorLen = data.consumeInt(1, 500);
            int matrixLen = data.consumeInt(1, 500);

            double[][] matrix = new double[matrixLen][vectorLen];

            for (int i = 0; i < matrixLen; ++i) {
                double[] tmp_vector = new double[vectorLen];
                for (int j = 0; j < vectorLen; ++j) {
                    tmp_vector[j] = data.consumeDouble();
                }
                matrix[i] = tmp_vector;
            }
```

Jazzer runs the code, but the file created has a huge string feeding ```CannedFuzzedDataProvider```. 

I do not know if this is a bug or if this issue is intended to be like it is, but Jazzer is generating a file that does not compile.

I let here a controlc.com for the crash file if you want to check it. [Crash file](https://controlc.com/2a6df309)

fmeum:
Java source files have a limit of 64KB on the size of string constants and Jazzer should honor that limit when generating Java reproducers. We might be able to get away with String concatenation, but I will have to investigate solutions.

bertschneider:
@cdanielgomes As @fmeum pointed out the generated data string is tool long. I created PR #283 to fix this issue.

PR #283, opened 2022-01-11
Split crash reproducer data

bertschneider:
If the recorded `FuzzedDataProvider` invocation data gets too long it can not be assigned to one `String` variable in the crash reproducer template anymore. More concretely one constant pool `CONSTANT_Utf8_info` entry can only hold data of
a size up to `uint16`.
This PR splits up the recorded data into multiple chunks to always generate valid crash reproducers.

Furthermore, a `tests` directory is introduced to hold an explicit Jazzer integration test to verify this issue (#269). 
The integration test compiles and executes the generated crash reproducer to verify it's functionality.


fmeum:
@bertschneider `dd` appears to have a different syntax on macOS (see the pipeline). Maybe there is a common subset of the `dd` arguments among all platforms that we could use here.

bazel/FuzzTargetTestWrapper.java
```diff
@@ -15,32 +15,47 @@
...
     } catch (IOException | ArrayIndexOutOfBoundsException e) {
       e.printStackTrace();
       System.exit(1);
       return;
     }
 
-    ProcessBuilder processBuilder = new ProcessBuilder();
-    Map<String, String> environment = processBuilder.environment();
-    // Ensure that Jazzer can find its runfiles.
-    environment.putAll(runfiles.getEnvVars());
```
fmeum, line 42: We should set these env vars in the ProcessBuilder created below.

bertschneider: The env vars didn't look Jazzer specific and all tests passed without. 
I can revert this but wonder which ones Jazzer needs.

bertschneider:
> @bertschneider `dd` appears to have a different syntax on macOS (see the pipeline). Maybe there is a common subset of the `dd` arguments among all platforms that we could use here.

AFAIK `dd` is not available on Windows, so I checked in an appropriate input file. It's not ideal but in this case the file is small enough and should not blow up the repository too much.

bazel/FuzzTargetTestWrapper.java
fmeum: They are not Jazzer-specific, but encode information about how Jazzer can find it's data dependencies provided by Bazel when invoked by the test wrapper. This is always easy to do from the top-level target executed by Bazel, but subprocesses generally have to be passed these variable to find their runfiles in all scenarios (that is, even when executed with bazel run on Windows). If the variables are not passed, runfile discovery will be non-deterministic on Windows.

bertschneider: Done. Thx for the explanation.

bertschneider:
Added a mapping of all parameters not starting with `-` to their actual locations using `rlocation`. With this change all tests are successful.

fmeum:
Thanks, this looks pretty good now! 

One last nit: Could we drop the `Example` prefix from the files and targets under `test`? The names get pretty long and now that this doesn't live under `examples` any more, the prefix has lost its purpose.

bertschneider:
Ah, yes, good point. I removed the `Example` prefix. Thank you very much for the review(s)!

fmeum (approved):
Thanks! Feel free to merge this now (probably squash merge).

PR #285, opened 2022-01-13
Use non-internal class names for target_class in tests

fmeum:
Some of our older tests currently set the --target_class to an internal
class name, which we don't officially support. This is causing trouble
for the Soot integration on the devassist branch, but is worth fixing on
the main branch as well.


bertschneider (approved):
Looks good. It seems that you have changed all places, at least I could not find any other ones.

PR #287, opened 2022-01-21
Remove duplicated shading

fmeum:
Now that we shade the entire agent, we don't have to shade the instrumentor deps anymore.


fmeum:
@bertschneider I added the missing shading rule as well as a test that should catch these issues in the future.

PR #286, opened 2022-01-19
Update Bazel to 5.0.0

fmeum:

PR #290, opened 2022-01-27
Re-enable pinning local JDK versions in CI

fmeum:
This ensures that Bazel indeed runs with the expected version of the JDK. This previously didn't work because I missed a typo in the migration notes (localjdk instead of local_jdk).


bertschneider:
My understanding was that `JAVA_HOME` is already set in the pipeline and no explicit `java_runtime_version` flag is needed. At least the correct Java version was printed in the run for https://github.com/CodeIntelligenceTesting/jazzer/pull/286/commits/981dd573f5a4f2fb0f77bfc5406c465e31b08cdf.
Do I misunderstand something?

fmeum:
The `JAVA_HOME` is already set correctly if the setup-java action behaves correctly, but with this flag, we ensure that Bazel will produce a hard error if the Java version doesn't match (e.g. also on Windows, where we currently don't print the JDK version and so technically don't know whether we are using the correct JDK).

bertschneider (approved):
Ah, I see. Could you please rebase the PR on `main`?

PR #292, opened 2022-01-27
fixed infinite recursion

stheid:


stheid:
@fmeum

agent/src/main/java/com/code_intelligence/jazzer/static_analysis/SootJazzerCFGIdMapper.kt
```diff
@@ -87,7 +87,10 @@ fun <E> DefaultDirectedGraph<Unit, E>.simplify() {
         }
 }
 
-object SootJazzerCFGIdMapper {
+class SootJazzerCFGIdMapper {
+    val outputGraph = DefaultDirectedGraph<Unit, DefaultEdge>(DefaultEdge::class.java)
+    val visited = ArrayList<Unit>()
```
fmeum, line 92: Based on the usage below, this should probably be a set (otherwise, due to lookups taking linear time, the algorithm below could become quadratic in the number of visited units):
```suggestion
    val visited = mutableSetOf<Unit>()
```

agent/src/main/java/com/code_intelligence/jazzer/static_analysis/SootJazzerCFGIdMapper.kt
```diff
@@ -98,17 +101,15 @@ object SootJazzerCFGIdMapper {
      */
     private fun getJazzerIds(
         icfg: JimpleBasedInterproceduralCFG,
-        m: SootMethod,
-        outputGraph: DefaultDirectedGraph<Unit, DefaultEdge>
+        m: SootMethod
     ): MutableList<Unit> {
```
fmeum, line 105: This return value doesn't seem to be used anywhere, maybe it can be dropped?

agent/src/main/java/com/code_intelligence/jazzer/static_analysis/SootJazzerCFGIdMapper.kt
```diff
@@ -87,7 +87,10 @@ fun <E> DefaultDirectedGraph<Unit, E>.simplify() {
         }
 }
 
-object SootJazzerCFGIdMapper {
+class SootJazzerCFGIdMapper {
```
stheid, line 90:
```suggestion
object SootJazzerCFGIdMapper {
```

agent/src/main/java/com/code_intelligence/jazzer/static_analysis/SootJazzerCFGIdMapper.kt
```diff
@@ -88,6 +88,9 @@ fun <E> DefaultDirectedGraph<Unit, E>.simplify() {
 }
 
 object SootJazzerCFGIdMapper {
```
fmeum, line 90: Keeping this around as an `object` means that `outputGraph` will essentially be a global variable that is never reset, which makes it difficult to use the mapper multiple times. 

To fix the build failure, you could instead update the call site at https://github.com/CodeIntelligenceTesting/jazzer/blob/0062eb78986d1a90851935716aaeb598d30f7b17/agent/src/main/java/com/code_intelligence/jazzer/static_analysis/CallGraph.kt#L36.

fmeum (changes requested):

fmeum:
@stheid The current CI failure is unrelated to your PR and will be fixed with https://github.com/CodeIntelligenceTesting/jazzer/pull/293, onto which I will rebase `devassist` once it has been merged.

fmeum:
@stheid I rebased your branch onto `devassist`, let's see what the CI has to say now.

fmeum:
I fixed some more issues with the soot integration and rebased again.

fmeum:
@stheid The pipeline is green :-)

stheid:
LGTM. i didnt even know what the "object" keyword means. it was in this repository already and i just addapted the code

PR #293, opened 2022-01-27
Disable BuildBuddy integration

fmeum:
GitHub Actions does not emit secrets into workflow started from forks,
which results in the BuildBuddy authorization header being empty for PRs
of external contributors. Rather than interpreting such requests as
unauthenticated, BuildBuddy errors out:
```
ERROR: The Build Event Protocol upload failed: Not retrying publishBuildEvents, no more attempts left: status='Status{code=UNAUTHENTICATED, description=failed to parse API key: missing API Key, cause=null}' UNAUTHENTICATED: UNAUTHENTICATED: failed to parse API key: missing API Key UNAUTHENTICATED: UNAUTHENTICATED: failed to parse API key: missing API Key
```

At this point (as we are not using Remote Build Execution, just build
result upload), I don't see a reason to further complicate our CI setup
by setting the header conditionally. Rather, just disable the upload for
now.


bertschneider (approved):
Disabling the BuildBuddy integration as a quick fix seems ok. Still, we should investigate this further.

fmeum:
@bertschneider I agree, it is useful to troubleshoot rebuilds and cache misses. I will create an issue to track this.

fmeum:
Created #294.

PR #302, opened 2022-02-06
Add an UnsafeProvider

fmeum:


fmeum:
@bertschneider I did some research and have not been able to find any JVM that doesn't ship `Unsafe`. However, some JVMs expose the instance on a field with a different name, hence the slightly more convoluted lookup. Since we will usually use `Unsafe` from the agent, we can for the most part rely on `Unsafe.getUnsafe()` returning what we want without the need to use reflection.

Regarding falling back to a "safe" strategy: Since this would introduce additional complexity and is hard to test (how do we test fallback logic if we don't have an example where the fallback is taken), I personally would rather not maintain two coverage map implementations. If Unsafe should ever become unavailable, we could always bring it back with additional JVM options (which we are in full control of) or migrate to a replacement, such as Foreign Memory Access. What do you think?

agent/src/main/java/com/code_intelligence/jazzer/runtime/UnsafeProvider.java
```diff
@@ -0,0 +1,53 @@
...
+package com.code_intelligence.jazzer.runtime;
+
+import java.lang.reflect.Field;
+import sun.misc.Unsafe;
+
+public final class UnsafeProvider {
+  private static Unsafe unsafe = null;
+
+  public static Unsafe getUnsafe() {
+    if (unsafe == null) {
```
bertschneider, line 24: Why not skip the "double checked locking" pattern and directly use `getUnsafeInternal` to set the value of `unsafe` once during startup? Is there a special reason to initialize it lazily.

bertschneider (commented):
I can vaguely remember some security manager problems accessing `unsafe` but it seems that you already investigated that.

Yes, good point regarding the fallback strategy. I agree that we can omit it.

agent/src/main/java/com/code_intelligence/jazzer/runtime/UnsafeProvider.java
fmeum: I initially included this as I wanted control over when exceptions would be thrown, but since we are now catching `Throwable`, that serves no purpose anymore. Thanks!

bertschneider (approved):

PR #303, opened 2022-02-07
Remove checked-in bazelisk and update README

fmeum:
Now that we support three different platforms and two architectures,
maintaining an in-repo Bazelisk binary for just a single pair doesn't
make much sense anymore. Instead, provide clear and up-to-date
instructions on how to obtain and use Bazelisk to build Jazzer.

Also includes a general overhaul of the "Getting Jazzer" section of the
README and updates the Docker image to function without a checked-in
Bazelisk.


bertschneider (approved):
Nice cleanup!

PR #304, opened 2022-02-07
Get ASM from Maven rather than gitlab.ow2.org

fmeum:
The OW2 GitLab hasn't been very reliable in the past and just
encountered another outage. Getting the ASM jars from Maven should be
more reliable.


fmeum:
@bertschneider What is the recommended way to handle `module-info.class` files from dependencies? Should we drop them from the deploy jar, merge them somehow or simply drop them?

bertschneider:
I would just drop it. asm is shaded and not exposed otherwise.

fmeum:
@bertschneider I solved this by finally writing a small tool that hermetically deletes files from a ZIP. I'm pretty sure this will have other applications. We could e.g. use it to strip the kotlin module files from the deploy JAR as they are only needed for libraries, not applications.

bertschneider (approved):
I like the solution and agree that the tool could come in handy in the future.

PR #308, opened 2022-02-11
Use tags for all dependencies

fmeum:
(Marked as draft as it is based on #304)

GitHub support has confirmed that only the /archive/refs/tags endpoint can be assumed to return stable results over time.

For dependencies that do not offer tags we can use, I created a fork with a "commit-<SHA>" tag for the commit we were previously depending on.

Along the way, update some dependencies to their latest tag.

See this comment for more context regarding the GitHub guarantees for archive stability:
https://github.com/bazel-contrib/SIG-rules-authors/issues/11#issuecomment-1029861300


bertschneider (commented):
Minus the changes from #304 this looks good. Thx for the tedious update of the dependencies.

bertschneider (approved):

PR #309, opened 2022-02-11
Disallow REPLACE hooks on <init>

fmeum:
REPLACE hooks on <init> can't work as only <init> can mark the stack variable as initialized.


bertschneider (approved):
LGTM

PR #310, opened 2022-02-14
Synchronize coverage ID generation

bertschneider:
Even though concurrent fuzzing is not fully supported situations arise in which classes are loaded concurrently (e.g. by used frameworks). This could lead to locking issues in the provided file based `SynchronizedCoverageIdStrategy`.

This PR moves the responsibility to guard access to coverage IDs from `RuntimeInstrumentor` to the concrete strategies by changing the `CoverageIdStrategy` interface. Based on their use-cases strategies can now implement different synchronization and error handling.

Now, both provided strategies synchronize coverage ID generation to support concurrent class loading.

This resolves #306


bertschneider:
Thx for the hint. I wasn't aware that this also works for discussions.

fmeum (approved):

PR #312, opened 2022-02-14
Standalone Autofuzz Java reproducers

fmeum:
The Autofuzz Java reproducers are based on the existing codegen
functionality and have the following advantages compared to the usual
Java reproducers generated for Autofuzz findings:
* They are completely independent of Jazzer and can thus be debugged
  easily, even by developers not familiar with Jazzer.
* They are stable with respect to the classpath as they encode all
  choices made during Autofuzz in literal references to classes.
* They are human-readable, especially after a straightforward pass
  through IntelliJ's formatting and code cleanups.


bertschneider (approved):
Really nice feature!

PR #316, opened 2022-02-16
Add a sanitizer for regex injection with CANON_EQ flag

fmeum:


bertschneider (approved):

bertschneider:
This sanitizer looks good. Still, we should also add checks for regex injections of *evil*, a.k.a. fuzzer provided,  patterns.

PR #318, opened 2022-02-16
Only trim reflection and Autofuzz classes from stack trace

fmeum:
Trimming all classes in java.** and jdk.** from the stack trace made it
useless when fuzzing the Java standard library itself. Also fixes the
matching of Autofuzz classes.


bertschneider (approved):
Good point! Thx.

PR #321, opened 2022-02-18
Make libFuzzer's seed available to hooks

fmeum:


driver/libfuzzer_driver.cpp
```diff
@@ -110,7 +112,35 @@ AbstractLibfuzzerDriver::AbstractLibfuzzerDriver(
...
+  } else {
+    // No explicit seed has been set. Since Jazzer hooks might still want to use
+    // a seed and we have to ensure that a fuzzing run can be reproduced by
+    // setting the seed printed by libFuzzer, we generate a seed for it here so
+    // that the two stay in sync.
+    unsigned int random_seed = std::random_device()();
+    std::string seed_arg = "-seed=" + std::to_string(random_seed);
+    // This argument can be accessed by libFuzzer at any (later) time and thus
+    // cannot be safely freed by us.
+    modified_argv.push_back(strdup(seed_arg.c_str()));
```
fmeum, line 139: @kyakdan This has the side effect of forwarding the `seed=` argument to all children processes in fork mode. I am not sure whether this makes forked fuzzing less efficient. What do you think?

agent/src/main/java/com/code_intelligence/jazzer/api/Jazzer.java
```diff
@@ -491,6 +501,18 @@ public static void reportFindingFromHook(Throwable finding) {
...
+  private static int getLibFuzzerSeed() {
+    // The Jazzer driver sets this property based on the value of libFuzzer's -seed command-line
+    // option, which allows for fully reproducible fuzzing runs if set. If not running in the
+    // context of the driver, fall back to a random number instead.
+    String rawSeed = System.getProperty("jazzer.seed");
+    if (rawSeed == null) {
+      return new SecureRandom().nextInt();
+    }
+    // If jazzer.seed is set, we expect it to be a valid integer.
+    return Integer.parseUnsignedInt(rawSeed);
```
fmeum, line 513: @bertschneider and I are debugging a weird issue where `rawSeed` is not null, but the empty string if this method is called early (e.g. during the agent's `premain`).

fmeum: Fixed, it was just a bug in the C++ part combined with very lazy class loading masking the root cause.

driver/libfuzzer_driver.cpp
fmeum: As per our discussion, I am now only adding `-seed` to the command line if not in fork/merge/join mode.

agent/src/main/java/com/code_intelligence/jazzer/api/Jazzer.java
```diff
@@ -491,6 +501,18 @@ public static void reportFindingFromHook(Throwable finding) {
     }
   }
 
+  private static int getLibFuzzerSeed() {
+    // The Jazzer driver sets this property based on the value of libFuzzer's -seed command-line
+    // option, which allows for fully reproducible fuzzing runs if set. If not running in the
+    // context of the driver, fall back to a random number instead.
+    String rawSeed = System.getProperty("jazzer.seed");
+    if (rawSeed == null) {
+      return new SecureRandom().nextInt();
```
kyakdan, line 510: This might lead to crashes being irreducible when fuzzing with the fork mode. Would it make sense to print this seed to the user so that she/he can set the `-seed` flag when running the fuzzer with the crashing input?

fmeum: It shouldn't be if used correctly, but I will add a comment explaining the restrictions: Hooks should only use the seed to determine the hints sent to the fuzzer (e.g. via `Jazzer.guideToEquality`), but never to decide whether or not to report a finding. If used in this way, reproduction always works and furthermore actual fuzzing runs can be made deterministic via the `-seed` argument. Fork runs are almost impossible to reproduce anyway since each child will be given a different seed and a different portion of the input corpus, so even if we were to print the seed here, I fear it might not be useful.

driver/jvm_tooling.cpp
```diff
@@ -155,7 +155,7 @@ void DumpJvmStackTraces() {
   // Do not detach as we may be the main thread (but the JVM exits anyway).
 }
 
-std::string dirFromFullPath(const std::string &path) {
+std::string_view dirFromFullPath(std::string_view path) {
   const auto pos = path.rfind(kPathSeparator);
   if (pos != std::string::npos) {
```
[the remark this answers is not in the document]
fmeum: Thanks, fixed.

agent/src/main/java/com/code_intelligence/jazzer/api/Jazzer.java
fmeum: I expanded on the issue in the javadoc comment, let me know if you find it clear enough.

kyakdan: It is clear now. You are right about fork runs.

kyakdan (approved):

PR #323, opened 2022-02-19
Replace nop with ret sled on ARM64

fmeum:
Instead of running through 64 nops on average, every instruction in the
trampoline now jumps directly to the end of the function.


fmeum:
@sebastianpoeplau I tested this on my M1 Mac. I didn't run any benchmarks though, so let me know if you think this may fare worse.

sebastianpoeplau (approved):

sebastianpoeplau:
> @sebastianpoeplau I tested this on my M1 Mac. I didn't run any benchmarks though, so let me know if you think this may fare worse.

I find it hard to predict branch prediction :grin: But depending on the position where we start executing NOPs, they might be slower than a single jump most of the time...

PR #324, opened 2022-02-20
Make MethodHandles in Jazzer final

fmeum:
This unlocks JIT optimizations.

### DIFF
--- a/agent/src/main/java/com/code_intelligence/jazzer/api/Jazzer.java
+++ b/agent/src/main/java/com/code_intelligence/jazzer/api/Jazzer.java
@@ -37,25 +37,40 @@ final public class Jazzer {
    */
   public static final int SEED = getLibFuzzerSeed();
 
-  private static Class<?> jazzerInternal = null;
+  private static final Class<?> JAZZER_INTERNAL;
 
-  private static MethodHandle traceStrcmp = null;
-  private static MethodHandle traceStrstr = null;
-  private static MethodHandle traceMemcmp = null;
+  private static final MethodHandle TRACE_STRCMP;
+  private static final MethodHandle TRACE_STRSTR;
+  private static final MethodHandle TRACE_MEMCMP;
 
-  private static MethodHandle consume = null;
-  private static MethodHandle autofuzzFunction1 = null;
-  private static MethodHandle autofuzzFunction2 = null;
-  private static MethodHandle autofuzzFunction3 = null;
-  private static MethodHandle autofuzzFunction4 = null;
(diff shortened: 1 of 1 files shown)


agent/src/main/java/com/code_intelligence/jazzer/api/Jazzer.java
```diff
@@ -23,68 +23,83 @@
...
+    MethodHandle autofuzzFunction1Local = null;
+    MethodHandle autofuzzFunction2Local = null;
+    MethodHandle autofuzzFunction3Local = null;
+    MethodHandle autofuzzFunction4Local = null;
+    MethodHandle autofuzzFunction5Local = null;
+    MethodHandle autofuzzConsumer1Local = null;
+    MethodHandle autofuzzConsumer2Local = null;
+    MethodHandle autofuzzConsumer3Local = null;
+    MethodHandle autofuzzConsumer4Local = null;
+    MethodHandle autofuzzConsumer5Local = null;
```
kyakdan, line 59: Could you clarify why we need the temporary variables? IISC, they are assigned once and we could therefore use the final fields directly

fmeum: We unfortunately can't due to the checked exceptions: All finals have to end up initialized in all code paths, including if an exception is thrown in the middle. The only alternative would be to extract the individual initializations into functions.

agent/src/main/java/com/code_intelligence/jazzer/api/Jazzer.java
```diff
@@ -23,68 +23,83 @@
  * Helper class with static methods that interact with Jazzer at runtime.
  */
 final public class Jazzer {
-  private static Class<?> jazzerInternal = null;
+  private static final Class<?> jazzerInternal;
```
kyakdan, line 26: According to Google's style guide, static final fields should use upper shake case https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names.

agent/src/main/java/com/code_intelligence/jazzer/api/Jazzer.java
kyakdan: I see. Thanks!

agent/src/main/java/com/code_intelligence/jazzer/api/Jazzer.java
fmeum: Thanks, this also neatly solves the naming problem.

kyakdan (approved):

PR #328, opened 2022-02-21
Expose value profile bitmap via Jazzer#exploreState

fmeum:


agent/src/main/java/com/code_intelligence/jazzer/api/Jazzer.java
```diff
@@ -509,6 +515,54 @@ public static void guideTowardsContainment(String haystack, String needle, int i
...
+   * @param id a (probabilistically) unique identifier for this particular state hint
+   */
+  public static void exploreState(byte state, int id) {
+    // We only use the lower 7 bits of state, which allows for 128 different state values tracked
+    // per id. The particular amount of 7 bits of state is also used in libFuzzer's
+    // TracePC::HandleCmp:
+    // https://github.com/llvm/llvm-project/blob/c12d49c4e286fa108d4d69f1c6d2b8d691993ffd/compiler-rt/lib/fuzzer/FuzzerTracePC.cpp#L390
+    // This value should be large enough for most use cases (e.g. tracking the length of a prefix in
+    // a comparison) while being small enough that the bitmap isn't filled up too quickly
+    // (65371 / 128 ≈ 510).
```
kyakdan, line 540: The value bit map size is 65536 bytes

agent/src/main/java/com/code_intelligence/jazzer/runtime/TraceDataFlowNativeCallbacks.java
```diff
@@ -57,7 +57,7 @@
 
   /* indirect-calls */
   // Calls: void __sanitizer_cov_trace_pc_indir(uintptr_t Callee);
-  private static native void tracePcIndir(int callee, int caller);
+  public static native void tracePcIndir(int callee, int caller);
```
kyakdan, line 60: Could we switch the order of the parameters to keep them in sync with the libFuzzer's implementation `void TracePC::HandleCallerCallee(uintptr_t Caller, uintptr_t Callee)`?

agent/src/main/java/com/code_intelligence/jazzer/api/Jazzer.java
```diff
@@ -509,6 +515,54 @@ public static void guideTowardsContainment(String haystack, String needle, int i
...
+    //    upper 3 bits are fixed, which leaves a total of 21 variable bits on x86_64.
+    // 2. On arm64 macOS, where every instruction is aligned to 4 bytes, the lower 2 bits of the
+    //    caller address will always be zero, further reducing the number of variable bits in the
+    //    caller parameter to 7.
+    // https://github.com/llvm/llvm-project/blob/c12d49c4e286fa108d4d69f1c6d2b8d691993ffd/compiler-rt/lib/fuzzer/FuzzerTracePC.cpp#L121
+    // Even taking these restrictions into consideration, we pass state in the lowest bits of the
+    // caller address, which is used to form the lowest bits of the bitmap index. This should result
+    // in the best caching behavior as state is expected to change quickly in consecutive runs and
+    // in this way all its bitmap entries would be located close to each other in memory.
+    int lowerBits = state | (id << 7);
```
kyakdan, line 557: The first comment mentions that we only use the lower 7 bits of the state, but here the most significant bit of the state affects the calculation of `lowerBits`. Shouldn't we set that bit to zero `int lowerBits = (0x7f & state) | (id << 7);`?

agent/src/main/java/com/code_intelligence/jazzer/api/Jazzer.java
```diff
@@ -509,6 +515,54 @@ public static void guideTowardsContainment(String haystack, String needle, int i
...
+    int upperBits = id >>> 12;
```
kyakdan, line 558: Could you clarify this shift operation? IISC, we effectively use the first five bytes of the `id` for `lowerBits` (because of the OR operation done by libfuzzer). In order to use the next 12 bytes, we need to shift the original `id` by the 5 to do that. So shouldn't that code be as follows?
```java
int lowerBits = (state & 0x7f) | (id << 7);
int upperBits = id >>> 5;
```

examples/src/main/java/com/example/MazeFuzzer.java
```diff
@@ -0,0 +1,147 @@
...
+
+  private static void executeCommands(byte[] commands, Consumer3<Byte, Byte, Boolean> callback) {
+    byte x = 0;
+    byte y = 0;
+    callback.accept(x, y, false);
+
+    for (byte command : commands) {
+      byte nextX = x;
+      byte nextY = y;
+      switch (command % 4) {
```
kyakdan: I just added one commit to make it a bit more challenging for the fuzzer. WDYT?

agent/src/main/java/com/code_intelligence/jazzer/api/Jazzer.java
fmeum: For this use case the effective size is 65371, since that is the prime used to reduce the index. But that level of detail only adds to the confusion here, modified the comment.

agent/src/main/java/com/code_intelligence/jazzer/api/Jazzer.java
fmeum: Good catch, done.

agent/src/main/java/com/code_intelligence/jazzer/api/Jazzer.java
fmeum: Fixed

examples/src/main/java/com/example/MazeFuzzer.java
```diff
@@ -0,0 +1,149 @@
...
+        case 'L':
+          nextX--;
+          break;
+        case 'R':
+          nextX++;
+          break;
+        case 'D':
+          nextY--;
+          break;
+        case 'U':
```
fmeum, line 108: I amended your commit to align the meaning of the commands with their effect on the printed maze.

agent/src/main/java/com/code_intelligence/jazzer/runtime/TraceDataFlowNativeCallbacks.java
fmeum: I thought about that at first, but it would mean breaking consistency with all the other Java PC hooks, which pass `id` in last. Should I perhaps also move those `id` arguments to the beginning of the argument list?

examples/src/main/java/com/example/MazeFuzzer.java
fmeum: Looks good, made one change.

examples/src/main/java/com/example/MazeFuzzer.java
kyakdan: Thanks!

agent/src/main/java/com/code_intelligence/jazzer/runtime/TraceDataFlowNativeCallbacks.java
kyakdan: I don't think it is that urgent but caused some confusion for me during the review. Moving the `id` arguments at the beginning of the argument list is a good idea and guarantees that the remaining arguments can be kept in sync with the corresponding ones for the libFuzzer hooks. As I said, it is nice to have but not very urgent.

agent/src/main/java/com/code_intelligence/jazzer/api/Jazzer.java
kyakdan: have you pushed the fixes?

fmeum:
@kyakdan Do you think we should add some way to combine multiple states into one to the `Jazzer` class, similar to what we use in the `MazeFuzzer`?

fmeum:
@kyakdan I think I forgot to push earlier, should be fixed now.

kyakdan:
> @kyakdan Do you think we should add some way to combine multiple states into one to the `Jazzer` class, similar to what we use in the `MazeFuzzer`?

@fmeum I've already thought about it. Do you already have an idea how we would support that?

kyakdan (approved):

PR #329, opened 2022-02-21
Allow custom hooks to statically reference classes to hook

fmeum:
By retransforming classes that are subject to hook instrumentation after
all hooks have been loaded, we ensure that classes don't fail to be
instrumented with hooks because they are statically referenced by hooks.
This makes it easier to write efficient and readable hooks as it removes
the need for using lazy reflection to access members of classes that
should be hooked.


fmeum:
@bertschneider I added a test, which led me to fix a separate bug. I will also submit a follow-up refactoring to properly resolve the FIXME in the test.

bertschneider (approved):
Good catch!

PR #331, opened 2022-02-21
Support hooks on interfaces

bertschneider:
If the target type of a hook is abstract or an interface, it is applied on all its implementations.

### DIFF
--- a/agent/src/main/java/com/code_intelligence/jazzer/api/MethodHook.java
+++ b/agent/src/main/java/com/code_intelligence/jazzer/api/MethodHook.java
@@ -143,6 +143,11 @@ public @interface MethodHook {
    * The name of the class that contains the method that should be hooked,
    * as returned by {@link Class#getName()}.
    * <p>
+   * If an interface or abstract class is specified, also calls to all
+   * implementations and subclasses available on the classpath during startup
+   * are hooked, respectively. Interfaces and subclasses are not taken into
+   * account for concrete classes.
+   * <p>
    * Examples:
    * <p><ul>
    * <li>{@link String}: {@code "java.lang.String"}
--- a/agent/src/main/java/com/code_intelligence/jazzer/instrumentor/Hooks.kt
+++ b/agent/src/main/java/com/code_intelligence/jazzer/instrumentor/Hooks.kt
@@ -1,0 +1,111 @@
+// Copyright 2021 Code Intelligence GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.code_intelligence.jazzer.instrumentor
+
+import com.code_intelligence.jazzer.api.MethodHook
+import com.code_intelligence.jazzer.api.MethodHooks
+import com.code_intelligence.jazzer.utils.ClassNameGlobber
+import io.github.classgraph.ClassGraph
+import io.github.classgraph.ScanResult
+import java.lang.reflect.Method
+
+data class Hooks(
+    val hooks: List<Hook>,
+    val hookClasses: Set<Class<*>>,
+    val additionalHookClassNameGlobber: ClassNameGlobber
+) {
+
+    companion object {
+        fun loadHooks(vararg hookClassNames: Set<String>): List<Hooks> {
+            return ClassGraph()
+                .enableClassInfo()
+                .enableSystemJarsAndModules()
+                .rejectPackages("jaz.*", "com.code_intelligence.jazzer.*")
+                .scan()
+                .use { scanResult ->
+                    // Capture scanResult in HooksLoader field to not pass it through
+                    // all internal hook loading methods.
+                    val loader = HooksLoader(scanResult)
+                    hookClassNames.map(loader::load)
+                }
+        }
+
+        private class HooksLoader(private val scanResult: ScanResult) {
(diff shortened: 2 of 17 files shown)


bertschneider:
The annotation property is called `targetClassName` which would be misleading, renaming it to `targetTypeName` would be a breaking change, and supporting both would be hard to maintain.
Extending the API docs sounds sufficient for now. @fmeum WDYT?

fmeum:
> The annotation property is called `targetClassName` which would be misleading, renaming it to `targetTypeName` would be a breaking change, and supporting both would be hard to maintain. Extending the API docs sounds sufficient for now. @fmeum WDYT?

Yes, I would keep the current name. If we clarify the docs, it won't be too confusing.

examples/BUILD.bazel
```diff
@@ -145,7 +145,7 @@ java_fuzz_target_test(
         "JAVA_OPTS": "-Dfoo=not_foo -Djava_opts=1",
     },
     fuzzer_args = [
-        "-fork=5",
+        "-fork=3",
```
fmeum, line 148: This change is generally fine if it's just that the CI machines do not have enough RAM to sustain that many forks. But I would like to know by how much this PR increases the peak JVM Heap usage. Do you have any data on that?

bertschneider: That was only a test to see if the JVMs would produce OOMs even if the overall memory is sufficient. GitHub Action runners have 7GB memory, which should cause problems when starting 5 JVM with `-Xmx1800m` plus libfuzzer.

Still, the used memory should not increase this dramatically. I finally found the problem, at least in my local tests. ClassGraph uses multiple threads for scanning, reducing those to one should reduce the used memory back to the old amount. Running the failing test locally requires ~130MB heap memory without and ~260MB with ClassGraph with default settings. Using only one scan thread reduces it back to ~130MB. I didn't investigate why the number of scan threads makes such a huge difference.

bertschneider: I analyzed heap dumps generated at the end of the test and found that the already closed `ScanResult` still holds a reference to a quite large map containing class name to `ClassInfo` relations. Explicitly removing the reference to the `ScanResult` (what should have been done to begin with) enables GCing it and aligns the heap dumps of both versions.

agent/src/main/java/com/code_intelligence/jazzer/utils/InstrumentationResources.kt
```diff
@@ -0,0 +1,50 @@
...
+@OptIn(ExperimentalTime::class)
+object InstrumentationResources {
+
+    // Class graph scan result to use in class lookups for hooks.
+    // This scan result includes class info and system jars and modules,
+    // no need for inter class dependencies (quite expensive).
+    // ScanResult should be closed manually, otherwise it's handled in a
+    // finalizer. To not introduce complicated state management this is done
+    // by the agent after instrumentation is finished.
+    var hookClassGraphResult: ScanResult?
```
fmeum, line 31: Given that this is only used from `Hook`, could we move into the companion object of that class? Since it is more narrow in scope then, we could close it right after the last call to `loadHooks` in `Agent`, potentially allowing memory to be freed sooner.

bertschneider: I moved it to `Hook` and it looks ok in `Agent`. On the other hand I would like to reduce responsibilities of `Agent`.

examples/BUILD.bazel
bertschneider: As discussed, there should be no memory increase after startup and the machine is simply overcommitted. Reduced the fork count to `3`.

agent/src/main/java/com/code_intelligence/jazzer/utils/InstrumentationResources.kt
bertschneider: As discussed, I refactored the hook loading to only call it from one place within `Agent`, so that `ScanResult` can be closed properly. This turned out to be a bigger refactoring.

agent/src/main/java/com/code_intelligence/jazzer/api/MethodHook.java
```diff
@@ -143,6 +143,11 @@
    * The name of the class that contains the method that should be hooked,
    * as returned by {@link Class#getName()}.
    * <p>
+   * If an interface or abstract class is specified, also calls to all
+   * implementations and subclasses available during startup are hooked,
```
fmeum, line 147: Could we clarify `available during startup`? Perhaps `available on the classpath during startup`?

bertschneider: Changed.

fmeum (approved):

agent/src/main/java/com/code_intelligence/jazzer/instrumentor/Hooks.kt
```diff
@@ -0,0 +1,110 @@
...
+                val additionalHookClassNameGlobber = ClassNameGlobber(
+                    hooks.flatMap(Hook::additionalClassesToHook),
+                    emptyList()
+                )
+                return Hooks(hooks, hookClasses, additionalHookClassNameGlobber)
+            }
+
+            private fun loadHooks(hookClassName: String): List<Pair<Hook, Class<*>>> {
+                return try {
+                    // Kotlin interferes with the caller sensitive Class.forName and no
```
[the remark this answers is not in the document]
bertschneider: I assumed Kotlin did some class loading trickery here because some "agent classes" were loaded by an application class loader (e.g. `AgentJarFinder`) and others by the bootstrap class loader.  `BootstrapClassLoader` is not able to access classes on the classpath, so custom hooks can not be found. This has nothing to do with Kotlin, correct. I updated the comment accordingly.  Thx for questioning this!
Still, I wonder why and where `Agent`, `AgentJarFinder` and some shaded Kotlin classes are loaded.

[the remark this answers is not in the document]
fmeum: I think that static dependencies of the `Agent` are loaded by an application class loader since that happens before I got a chance to append the agent JAR to the `BootstrapClassLoader`.

PR #335, opened 2022-02-23
Enable BuildBuddy integration

### LINKED ISSUE #294: Reenable BuildBuddy
We are currently not using BuildBuddy in the CI since the required secret isn't available in PR pipelines initiated from forks and BuildBuddy chokes on an empty authorization header. See #293 for more context.

A possible fix could be to set the remote header conditionally only when the secret is not empty.

bertschneider:
Enable BuildBuddy integration when the API key is available in the
workflow. Forks should still be able to execute the workflow without
piping logs to BuildBuddy.

This is done by setting the environment variable BUILD_BUDDY_CONFIG
based on the API secret. The script echoBuildBuddyConfig returns an
environment config stating the needed bazel settings if the API token is
passed in as first argument, otherwise an empty string is returned. The
return value is used to update the workflow environment settings.

The passed around secret is filtered out in the GitHub logs.

Fixes #294

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -33,6 +33,12 @@ run:windows --noincompatible_strict_action_env
 # Since the toolchain is conditional on OS and architecture, set it on the particular GitHub Action.
 build:toolchain --//third_party:toolchain
 
+# CI tests (not using the toolchain to test OSS-Fuzz & local compatibility)
+build:ci --bes_results_url=https://app.buildbuddy.io/invocation/
+build:ci --bes_backend=grpcs://cloud.buildbuddy.io
+build:ci --remote_cache=grpcs://cloud.buildbuddy.io
+build:ci --remote_timeout=3600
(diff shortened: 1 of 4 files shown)


bertschneider:
Another way could be to (conditionally) write the BuildBuddy config to a file and use `try-import`, as described on https://www.buildbuddy.io/docs/guide-auth#separate-auth-file.

fmeum:
> Another way could be to (conditionally) write the BuildBuddy config to a file and use `try-import`, as described on https://www.buildbuddy.io/docs/guide-auth#separate-auth-file.

I think I like the current approach better as it makes the flag passing more visible. Thanks!

fmeum (approved):

fmeum:
@bertschneider Edited the description as "Solves" doesn't seem to be a recognized auto link keyword.

PR #337, opened 2022-02-23
Refactor JNI usage to be more Java-centric

### LINKED ISSUE #336: `UnsatisfiedLinkError` if there is no class to instrument
```
❯ bazel run //:jazzer -- --autofuzz=java.util.regex.Pattern::compile 
INFO: Invocation ID: 1fd163aa-afe7-4555-b345-b1d2d9533867
INFO: Analyzed target //:jazzer (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //driver:jazzer_driver up-to-date:
  bazel-bin/driver/jazzer_driver
INFO: Elapsed time: 0.113s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
OpenJDK 64-Bit Server VM warning: Sharing is only supported for boot loader classes because bootstrap classpath has been appended
INFO: Loaded 8 hooks from com.code_intelligence.jazzer.sanitizers.Deserialization
INFO: Loaded 3 hooks from com.code_intelligence.jazzer.sanitizers.ExpressionLanguageInjection
INFO: Loaded 8 hooks from com.code_intelligence.jazzer.sanitizers.NamingContextLookup
INFO: Loaded 1 hooks from com.code_intelligence.jazzer.sanitizers.OsCommandInjection
INFO: Loaded 1 hooks from com.code_intelligence.jazzer.sanitizers.ReflectiveCall
INFO: Loaded 1 hooks from com.code_intelligence.jazzer.sanitizers.RegexInjection
Exception in thread "main" java.lang.UnsatisfiedLinkError: 'void com.code_intelligence.jazzer.runtime.CoverageMap.initialize(long)'
	at com.code_intelligence.jazzer.runtime.CoverageMap.initialize(Native Method)
	at com.code_intelligence.jazzer.runtime.CoverageMap.<clinit>(CoverageMap.java:55)
	at com.code_intelligence.jazzer.instrumentor.CoverageRecorder.updateCoveredIdsWithCoverageMap(CoverageRecorder.kt:60)
terminate called after throwing an instance of 'std::runtime_error'
  what():  Java exception occurred in CoverageTracker JNI code
```

fmeum:
Fixes #336, simplifies the code and unlocks callback optimizations (e.g. with JavaCritical).


agent/src/main/java/com/code_intelligence/jazzer/runtime/TraceDataFlowNativeCallbacks.java
```diff
@@ -19,47 +19,90 @@
 
 @SuppressWarnings("unused")
 final public class TraceDataFlowNativeCallbacks {
+  // Making this static final ensures that the JIT will eliminate the dead branch of a construct
+  // such as:
+  // if (USE_FAKE_PCS) ... else ...
+  private static final boolean USE_FAKE_PCS = useFakePcs();
```
kyakdan, line 25: Are there use cases where we don't want to use fake PCs?

agent/src/main/java/com/code_intelligence/jazzer/runtime/TraceDataFlowNativeCallbacks.java
```diff
@@ -19,47 +19,90 @@
...
+
   /* trace-cmp */
-  // Calls: void __sanitizer_cov_trace_cmp4(uint32_t Arg1, uint32_t Arg2);
-  public static native void traceCmpInt(int arg1, int arg2, int pc);
+  public static void traceCmpInt(int arg1, int arg2, int pc) {
+    if (USE_FAKE_PCS) {
+      traceCmpIntWithPc(arg1, arg2, pc);
+    } else {
+      traceCmpInt(arg1, arg2);
```
kyakdan, line 32: Are the hooks really useful without fake PCs? How about only calling the CMP hooks only when `USE_FAKE_PCS` is set?

fmeum: I think that for integer compares the hooks don't even let the PC influence the way they add entries to the TORC, so I would expect them to still be useful (even without value profiling).

kyakdan: You are right. The index is calculated by xoring the two arguments.

kyakdan (approved):

PR #338, opened 2022-02-24
Verify specific findings

fmeum:


bazel/fuzz_target.bzl
```diff
@@ -28,8 +28,14 @@ def java_fuzz_target_test(
         env = None,
         verify_crash_input = True,
         verify_crash_reproducer = True,
-        execute_crash_reproducer = False,
+        execute_crash_reproducer = True,
+        # Can't default to the empty string as it is passed as an argument and Bazel tokenizes the
+        # args.
+        expected_finding = "none",
         **kwargs):
+    if not verify_crash_reproducer:
```
bertschneider, line 36: Create an error when `execute_crash_reproducer` is set but no `expected_finding` is given?

bazel/fuzz_target.bzl
```diff
@@ -28,8 +28,14 @@ def java_fuzz_target_test(
         env = None,
         verify_crash_input = True,
         verify_crash_reproducer = True,
-        execute_crash_reproducer = False,
+        execute_crash_reproducer = True,
```
bertschneider, line 31: Could this be combined with `expected_finding`?

bazel/fuzz_target.bzl
```diff
@@ -62,13 +68,20 @@ def java_fuzz_target_test(
     else:
         fail("Invalid sanitizer: " + sanitizer)
 
+    if type(expected_finding) == type(""):
```
bertschneider, line 71: If used, is it possible to evaluate `select` here and only use `string` afterwards? Could not find anything in the documentation on that. 
If so, we could use `string` values in `select` instead of arrays and move `expected_findings` in `args` a few lines below.

bazel/tools/java/com/code_intelligence/jazzer/tools/FuzzTargetTestWrapper.java
```diff
@@ -92,13 +96,13 @@ public static void main(String[] args) {
         System.exit(4);
       }
       // Verify that libFuzzer dumped a crashing input.
-      if (verifyCrashInput
+      if (JAZZER_CI && verifyCrashInput
```
bertschneider, line 99: Why should this only be verified on CI? Also checks below.

examples/src/main/java/com/example/JpegImageParserFuzzer.java
```diff
@@ -23,6 +23,9 @@
 // Found https://issues.apache.org/jira/browse/IMAGING-275.
 public class JpegImageParserFuzzer {
   public static void fuzzerInitialize() {
+    // When running through a Java reproducer, do not check the Java opts.
```
bertschneider, line 26: Time to extract the Java opts checks into a dedicated test in `tests`?

fmeum: Good point, will do.

bazel/fuzz_target.bzl
fmeum: Unfortunately that is not possible: `select` is a bit of magic in that it is not evaluated when called, but rather returns an opaque object that encodes the alternatives. This (currently) makes it impossible to unwrap select and manipulate its values, rather the only supported operation on it is list concatenation. That makes it difficult to improve the logic here.

bazel/tools/java/com/code_intelligence/jazzer/tools/FuzzTargetTestWrapper.java
bertschneider: So the assumption is that someone executing tests is primarily interested in reproducing the findings and verifying the reproducers is a CI or regression topic. In that case let's leave the name as is.

fmeum:
@simonresch We now check for the precise expected finding in CI. Locally, running 
```
bazel test --config=ci //sanitizers/src/test/java/com/example:ExpressionLanguageInjection
```
fails with `java.lang.ClassNotFoundException: com.sun.el.ExpressionFactoryImpl` some time into fuzzing. Is that expected or is something missing from the classpath?

simonresch:
@bertschneider Thanks for already looking into this! There was indeed a dependency missing in the test. I pushed a commit that should fix it.

examples/src/main/java/com/example/JpegImageParserFuzzer.java
bertschneider: Much cleaner now, thx!

maven_install.json
```diff
@@ -147,6 +147,17 @@
                 "sha256": "43fdef0b5b6ceb31b0424b208b930c74ab58fac2ceeb7b3f6fd3aeb8b5ca4393",
                 "url": "https://repo1.maven.org/maven2/javax/activation/javax.activation-api/1.2.0/javax.activation-api-1.2.0.jar"
             },
+            {
+                "coord": "javax.el:el-api:2.2",
```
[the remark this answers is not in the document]
bertschneider: Updated the commit to use a matching version.

bertschneider (approved):

PR #341, opened 2022-02-25
Simplify registering PC tables

kyakdan:
libFuzzer only checks the value of a PC when the corresponding flag is
not zero. Therefore, it is safe to set the whole PC-Table array to zero.

### DIFF
--- a/driver/coverage_tracker.cpp
+++ b/driver/coverage_tracker.cpp
@@ -69,22 +69,16 @@ void CoverageTracker::RegisterNewCounters(JNIEnv &env, jint old_num_counters,
   }
   std::size_t diff_num_counters = new_num_counters - old_num_counters;
   // libFuzzer requires an array containing the instruction addresses associated
-  // with the coverage counters registered above. Given that we are
-  // instrumenting Java code, we need to synthesize addresses that are known not
-  // to conflict with any valid instruction address in native code. Just like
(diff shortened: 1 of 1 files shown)


fmeum (commented):
While it is true that Jazzer and libFuzzer don't use the contents of the PC array during regular fuzzing, the coverage logic at https://github.com/CodeIntelligenceTesting/jazzer/blob/fc4fb6d84de41e55db76650b42af91e780ff54c3/driver/coverage_tracker.cpp#L118 does. I don't immediately see a way to get rid of the fake instructions without breaking this.

kyakdan:
> While it is true that Jazzer and libFuzzer don't use the contents of the PC array during regular fuzzing, the coverage logic at
> 
> https://github.com/CodeIntelligenceTesting/jazzer/blob/fc4fb6d84de41e55db76650b42af91e780ff54c3/driver/coverage_tracker.cpp#L118
> 
> does. I don't immediately see a way to get rid of the fake instructions without breaking this.

I've overseen that! We can set the PC value to the loop counter so that each PC directly matches the corresponding edge id. This way we can use the array returned by `__sanitizer_cov_get_observed_pcs` directly without having to get those ids based on the addresses of the fake instruction.

fmeum (approved):
Looks good to me now. Could you update the commit message before merging?

PR #343, opened 2022-02-28
Minor refactoring and cleanup

kyakdan:
Few minor refactorings I did when going through the code during reviews

### DIFF
--- a/agent/src/main/java/com/code_intelligence/jazzer/runtime/CoverageMap.java
+++ b/agent/src/main/java/com/code_intelligence/jazzer/runtime/CoverageMap.java
@@ -14,7 +14,6 @@
 
 package com.code_intelligence.jazzer.runtime;
 
-import java.nio.ByteBuffer;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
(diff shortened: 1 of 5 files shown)


agent/src/main/java/com/code_intelligence/jazzer/instrumentor/CoverageRecorder.kt
```diff
@@ -106,7 +106,7 @@ object CoverageRecorder {
 
     private fun Double.format(digits: Int) = "%.${digits}f".format(this)
 
-    fun dumpJacocoCoverage(coveredIds: Set<Int>): ByteArray? {
+    private fun dumpJacocoCoverage(coveredIds: Set<Int>): ByteArray? {
```
fmeum, line 109: This is needed for OOP, same below.

fmeum:
Thanks, looks good except for the changed visibility modifiers.

agent/src/main/java/com/code_intelligence/jazzer/instrumentor/CoverageRecorder.kt
kyakdan: Fixed!

fmeum (approved):

PR #345, opened 2022-03-01
Integrate results of the callback benchmark

fmeum:


kyakdan:
@fmeum Could you clarify why the `JavaCritical` variant is not integrated here?

kyakdan (approved):

fmeum:
> @fmeum Could you clarify why the `JavaCritical` variant is not integrated here?

I simply haven't managed to get the `JavaCritical` implementations invoked in the real, non-benchmark fuzzer callback implementation. I will investigate why, but wanted to get this merged as is since it already improves the performance.

PR #347, opened 2022-03-01
Extend class loading detector

bertschneider:
This PR extends the hooked class loading methods and triggers a finding whenever the honeypot class is loaded and initialized.

### DIFF
--- a/sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ReflectiveCall.kt
+++ b/sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ReflectiveCall.kt
@@ -17,18 +17,36 @@ package com.code_intelligence.jazzer.sanitizers
 import com.code_intelligence.jazzer.api.HookType
 import com.code_intelligence.jazzer.api.Jazzer
 import com.code_intelligence.jazzer.api.MethodHook
+import com.code_intelligence.jazzer.api.MethodHooks
 import java.lang.invoke.MethodHandle
 
 /**
- * Detects unsafe reflective calls that lead to attacker-controlled method calls.
+ * Detects unsafe calls that lead to attacker-controlled class loading.
+ *
+ * Guide the fuzzer to load honeypot class via [Class.forName] or [ClassLoader.loadClass].
  */
 @Suppress("unused_parameter", "unused")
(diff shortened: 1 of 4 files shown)


sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ReflectiveCall.kt
```diff
@@ -17,18 +17,36 @@ package com.code_intelligence.jazzer.sanitizers
...
 @Suppress("unused_parameter", "unused")
 object ReflectiveCall {
 
-    @MethodHook(type = HookType.BEFORE, targetClassName = "java.lang.Class", targetMethod = "forName")
+    @MethodHooks(
+        MethodHook(type = HookType.BEFORE, targetClassName = "java.lang.Class", targetMethod = "forName", targetMethodDescriptor = "(Ljava/lang/String;)Ljava/lang/Class;"),
+        MethodHook(type = HookType.BEFORE, targetClassName = "java.lang.Class", targetMethod = "forName", targetMethodDescriptor = "(Ljava/lang/String;ZLjava/lang/ClassLoader;)Ljava/lang/Class;"),
+        MethodHook(type = HookType.BEFORE, targetClassName = "java.lang.ClassLoader", targetMethod = "loadClass", targetMethodDescriptor = "(Ljava/lang/String;)Ljava/lang/Class;"),
+        MethodHook(type = HookType.BEFORE, targetClassName = "java.lang.ClassLoader", targetMethod = "loadClass", targetMethodDescriptor = "(Ljava/lang/String;Z)Ljava/lang/Class;"),
+    )
```
kyakdan, line 36: should we also hook `java.lang.ClassLoader.findClass` methods?

bertschneider: `findClass` is (or should be only) invoked internally by `loadClass`. And calls to `loadClass` are now already hooked for all `ClassLoader`s. So, I think it's not necessary.

kyakdan: I see. Thanks for the clarification.

kyakdan (approved):

PR #349, opened 2022-03-03
Add a sanitizer for general regex injections

fmeum:


fmeum:
@bertschneider I fixed the test failure by not applying replace hooks to Java 6 class files. That is not ideal (especially since Java 6 is a common Kotlin target version), but I currently don't see any other way.

bertschneider:
The current [Kotlin version targets](https://kotlinlang.org/docs/compiler-reference.html#jvm-target-version) `1.8` by default and `1.6` is deprecated. IMO it's ok to only support replace hooks on `1.8+`.

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/RegexInjection.kt
```diff
@@ -33,34 +33,76 @@ object RegexInjection {
...
-        val pattern = args[0] as? String ?: return
-        val flags = args[1] as? Int ?: return
-        if (flags and PATTERN_CANON_EQ == 0) return
-        if (pattern.contains(CANON_EQ_ALMOST_EXPLOIT)) {
-            Jazzer.reportFindingFromHook(
-                FuzzerSecurityIssueLow(
-                    """Regular Expression Injection with CANON_EQ
+    fun patternHook(method: MethodHandle, alwaysNull: Any?, args: Array<Any?>, hookId: Int): Any? {
+        val pattern = args[0] as String
+        val hasCanonEqFlag = args.size >= 2 && (args[1] as? Int ?: 0) and Pattern.CANON_EQ != 0
```
bertschneider, line 71: Another layer of parentheses could help to mentally parse this check.

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/RegexInjection.kt
```diff
@@ -33,34 +33,76 @@ object RegexInjection {
...
     // Note: The fuzzer might trigger an OutOfMemoryError or NegativeArraySizeException (if the size
     // of the array overflows an int) by chance before it correctly emits this "exploit". In that
     // case, we report the original exception instead.
-    @MethodHook(
-        type = HookType.BEFORE,
-        targetClassName = "java.util.regex.Pattern",
-        targetMethod = "compile",
-        targetMethodDescriptor = "(Ljava/lang/String;I)Ljava/util/regex/Pattern;"
+    @MethodHooks(
+        MethodHook(
```
bertschneider, line 49: Should we also hook `String::replaceFirst`, `String::replaceAll`, `String::split` as those are probably often used.

fmeum: Thanks, I added hooks for them. An alternative would be to add `java.lang.String` to `additionalClassesToHook`, as it simply goes through the existing hooks on `Pattern.compile` and `Pattern.matches`. But I don't really feel comfortable opening `java.lang.String` up to all hooks...

bertschneider (approved):

PR #351, opened 2022-03-04
Make exceptions in LDAP hooks more obvious

bertschneider:


fmeum:
I agree it looks more obviously correct this way, but I think there isn't much of a difference: `Jazzer.reportFindingFromHook` throws an `Error` and Jazzer gives precedence to the reported finding over any potential uncaught exceptions.

bertschneider:
True, not much of a difference. It's a little bit more obvious that an exception is thrown in all cases without having to dig through the code. 
We should just use a consistent pattern when calling `reportFindingFromHook`. Handle it like it could return or like it will always throw an exceptions. Currently that's a little bit mixed. 
Sadly it's implemented in Java and we can not use the return type `Nothing` to indicate the thrown exception.

fmeum (approved):

PR #352, opened 2022-03-06
Fix regex injection hook invocation for String functions

fmeum:
Fixes the incorrect invocation of the MethodHandle in the case where the
regex injection hook is applied to a String function, which do require
passing in the this object:

== Java Exception: java.lang.invoke.WrongMethodTypeException: cannot convert MethodHandle(String,String,String)String to (Object,Object)Object
 at java.base/java.lang.invoke.MethodHandle.asTypeUncached(MethodHandle.java:881)
 at java.base/java.lang.invoke.MethodHandle.asType(MethodHandle.java:866)
 at java.base/java.lang.invoke.MethodHandle.invokeWithArguments(MethodHandle.java:729)
 at com.code_intelligence.jazzer.sanitizers.RegexInjection.hookInternal(RegexInjection.kt:126)
 at com.code_intelligence.jazzer.sanitizers.RegexInjection.patternHook(RegexInjection.kt:101)
 ...

Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=45246

### DIFF
--- a/sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/RegexInjection.kt
+++ b/sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/RegexInjection.kt
@@ -29,7 +29,7 @@ object RegexInjection {
      * Part of an OOM "exploit" for [java.util.regex.Pattern.compile] with the
      * [java.util.regex.Pattern.CANON_EQ] flag, formed by three consecutive combining marks, in this
      * case grave accents: ◌̀.
-     * See [patternCompileWithFlagsHook] for details.
+     * See [compileWithFlagsHook] for details.
      */
     private const val CANON_EQ_ALMOST_EXPLOIT = "\u0300\u0300\u0300"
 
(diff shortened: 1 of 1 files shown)


sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/RegexInjection.kt
```diff
@@ -49,7 +49,7 @@ object RegexInjection {
     fun compileWithFlagsHook(method: MethodHandle, alwaysNull: Any?, args: Array<Any?>, hookId: Int): Any? {
```
bertschneider, line 49: Link in `CANON_EQ_ALMOST_EXPLOIT` docs to this method is broken.

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/RegexInjection.kt
```diff
@@ -49,7 +49,7 @@ object RegexInjection {
...
         val pattern = args[0] as String?
         val hasCanonEqFlag = ((args[1] as Int) and Pattern.CANON_EQ) != 0
-        return hookInternal(method, pattern, hasCanonEqFlag, args, hookId)
+        return hookInternal(method, pattern, hasCanonEqFlag, hookId, args)
```
bertschneider, line 52: Spread `args`?

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/RegexInjection.kt
fmeum: Fixed.

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/RegexInjection.kt
fmeum: Fixed, thanks. I think it's not strictly necessary, but definitely more readable that way.

bertschneider (approved):
Looks good. Thx!

PR #354, opened 2022-03-08
Fix VerifyError and simplify debugging

fmeum:


tests/src/test/java/com/example/HookDependenciesFuzzer.java
```diff
@@ -39,10 +39,11 @@
   }
 
   @MethodHook(type = HookType.AFTER, targetClassName = "java.util.regex.Matcher",
-      targetMethod = "matches", additionalClassesToHook = {"java.util.regex.Pattern"})
+      targetMethod = "matches", targetMethodDescriptor = "()Z",
+      additionalClassesToHook = {"java.util.regex.Pattern"})
   public static void
   matcherMatchesHook(MethodHandle method, Object alwaysNull, Object[] alwaysEmpty, int hookId,
-      boolean returnValue) {
+      Boolean returnValue) {
```
bertschneider, line 46: `HookMethodVisitor` always passes return values as `Object`s. We could unwrap them to support primitive types as well, but that's probably not necessary. 
To prevent this error `Hook` should validate that the expected return value is not primitive. Currently the type is only checked if `targetMethodDescriptor` is present.

bertschneider (approved):
Nice improvement!

PR #356, opened 2022-03-10
Add JavaCritical implementations of fuzzer callbacks

fmeum:


fmeum:
@sebastianpoeplau I don't see a good way to test that the `JavaCritical` functions are actually picked up correctly. I verified that they are by inserting `abort()`s.

fmeum:
@sebastianpoeplau I decided not to bother with the Windows linker too much for something that is only a performance optimization using an obsoleted feature. Should be reviewable now.

sebastianpoeplau (approved):
Looks good! :blush:

PR #360, opened 2022-03-16
Add support for `compile_command.json` generation

bertschneider:
`compile_command.json`, as specified at [JSONCompilationDatabase](https://clang.llvm.org/docs/JSONCompilationDatabase.html), is used by many tools to provide C++ support. This file can easily be created using the added bazel rule.

C++ support is particular helpful when changing code in the `driver` module.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 /bazel-*
+/compile_commands.json
 .ijwb
 .clwb
+
--- a/WORKSPACE.bazel
+++ b/WORKSPACE.bazel
@@ -116,3 +116,18 @@ maven_install(
 load("@maven//:defs.bzl", "pinned_maven_install")
 
 pinned_maven_install()
+
+# Hedron's Compile Commands Extractor for Bazel
+# https://github.com/hedronvision/bazel-compile-commands-extractor
+# Enables C++ support in many editors. To generate "compile_commands.json" execute:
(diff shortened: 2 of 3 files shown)


bertschneider:
@fmeum Do you think it's useful to add this bazel rule? I really liked using a language server in my editor of choice to work on the C++ code in `driver`. I really don't like the additional linke in the root dir, though.

fmeum:
Not that I'm against this addition, but I want to make sure that we don't miss even better approaches: Have you tried CLion and/or is it not working for you? It works very well for me.

bertschneider:
CLion is working well for me. Supporting open tools could be helpful for possible contributors, though.

bertschneider:
From my side it would also be fine if we close this PR and point to it if a C++ support questions comes up. I just didn't want to throw away my C++/bazel investigation.

fmeum:
> CLion is working well for me. Supporting open tools could be helpful for possible contributors, though.

That's true, but if that is the focus, we should add a `Contributing` section to the README that explains how to use IntelliJ/CLion and the new `compile_command.json`. Could we do this as part of this PR?

bertschneider:
Good point! There was no appropriate section in the readme for this and I didn't want to start a new one. But if we make the scope a little bit broader it should work. I will try to come up with an initial version.

bertschneider:
Let's close this PR for now as it's quite old and I didn't find time to work on it.

PR #369, opened 2022-03-29
Add a Starlark rule for jar stripping

fmeum:
This rule is easier to use (and thus potentially upstreamable) and
correctly forwards coverage information.


agent/BUILD.bazel
```diff
@@ -16,15 +17,12 @@ java_binary(
     ],
 )
 
-genrule(
+strip_jar(
     name = "jazzer_agent_deploy",
-    srcs = [":jazzer_agent_shaded_deploy"],
-    outs = ["jazzer_agent_deploy.jar"],
```
bertschneider, line 22: `Dockerfile` still uses this target name.

bertschneider (approved):

PR #371, opened 2022-03-29
Generate combined coverage report for all tests

fmeum:
`bazel run //bazel/coverage` generates a combined LCOV coverage report for all tests via `bazel coverage` and then renders it to HTML. The report combines C++ and Java coverage information, including JNI code.


fmeum:
@bertschneider It would be great if you could generate the command locally for some tests and verify that the results look correct. Coverage support is still a pretty new feature in `rules_jni` and I discovered a number of issues with `bazel coverage` along the way that require workarounds.

bertschneider:
Makes sense. Thx for the clarification.

bertschneider (approved):

fmeum:
@bertschneider The ASan tests fail because there are actually two `RulesJni` classes: The one used by Jazzer and shaded to `com.code_intelligence.jazzer.third_party.**` and the usually one used by the examples. But `RulesJni` currently assumes that it is a singleton - it modifies files in the coverage directory. I will probably have to move all coverage files to a temporary directory and then move the report back, but that requires a new rules_jni release.

PR #372, opened 2022-03-31
Verify OSS-Fuzz build in CI

fmeum:


bertschneider (approved):
I really like the new check. Should prevent this in the future.

PR #373, opened 2022-03-31
Remove redundant class path entry

fmeum:
Java agent jars are automatically added to the system class loader,
which means that an explicit class path entry for them is both redundant
and potentially confusing.


bertschneider (approved):

Puma2511a (commented):
- image

Puma2511a:
**********_[image]()_

PR #374, opened 2022-04-01
Exclude OpenJ9 internal classes

bertschneider:
OpenJ9 was not mentioned in the documentation. Using version 11 and 17 to execute the `examples` causes random JVM seg-faults, though.

Nevertheless, the hook instrumentation should exclude OpenJ9 internal classes, analogous to Sun/Oracle internals.


bertschneider:
We should investigate the crashes further before merging. Closing the PR for now.

PR #376, opened 2022-04-02
Link jazzer_fuzzer_callbacks dynamically on Windows

fmeum:


fmeum:
@sebastianpoeplau Just FYI: Removing all checks in Java that require sanitizer hooks to function does let the fuzzer find the crash, so figuring out the weak symbol weirdness on Windows seems to be the last blocker.

sebastianpoeplau:
> @sebastianpoeplau Just FYI: Removing all checks in Java that require sanitizer hooks to function does let the fuzzer find the crash, so figuring out the weak symbol weirdness on Windows seems to be the last blocker.

I'll reserve time to have another look at the file we found :+1:

PR #381, opened 2022-04-04
Print JVM stack traces and exit cleanly on a fatal sanitizer finding

fmeum:


fmeum:
@sebastianpoeplau The call to `driver_cleanup`, which in turn calls `DestroyJavaVM`, causes unexpected failures with both ASan and UBSan:

```
AddressSanitizer: nested bug in the same thread, aborting.
```

```
double free or corruption (out)
==186512== ERROR: libFuzzer: deadly signal
```

Exiting the JVM directly from the Java method called by `DumpJvmStackTraces` does not lead to any issues. Do you see a possible cause for this?

sebastianpoeplau (approved):
The change looks good to me. I'm afraid I don't know enough about how Jazzer overrides the sanitizers' callback mechanism. Why do we store the function in `jazzer::AbstractLibfuzzerDriver::libfuzzer_print_crashing_input_` in addition to giving it to the sanitizer?

fmeum:
@sebastianpoeplau I haven't been able to get the `std::thread` idea to work and also ran into issues with `use_sigaltstack=0`, so this might be indicative of an issue on our end. I will merge this with just the improved stack traces and left a comment with what we know so far.

sebastianpoeplau:
> @sebastianpoeplau I haven't been able to get the `std::thread` idea to work and also ran into issues with `use_sigaltstack=0`, so this might be indicative of an issue on our end. I will merge this with just the improved stack traces and left a comment with what we know so far.

Sounds good! Maybe there's something else about the crash-processing context that we're not aware of...

PR #382, opened 2022-04-04
Use go install instead of go get in format CI job

fmeum:
go get for binaries is no longer supported with Go 1.18.


fmeum:
@bertschneider This causes the CI to be broken.

bertschneider (approved):

PR #383, opened 2022-04-04
Support autofuzz of inner classes

bertschneider:
Method references use `.` as separator between outer and inner classes. Class loaders expect a `$`. 
Autofuzz should support the proper method reference format.


agent/src/main/java/com/code_intelligence/jazzer/autofuzz/FuzzTarget.java
```diff
@@ -84,19 +84,28 @@ public static void fuzzerInitialize(String[] args) {
...
-          "Failed to find class %s for autofuzz, please ensure it is contained in the classpath "
-              + "specified with --cp and specify the full package name%n",
-          className);
-      e.printStackTrace();
-      System.exit(1);
-      return;
-    }
+    Class<?> targetClass = null;
+    String targetClassName = className;
+    do {
```
fmeum, line 89: What do you think of the alternative of replacing all `.` that precede a capital letter (except the first) with `$` and then only trying once? The current approach isn't bad, I'm just wondering whether it could cause any ambiguities.

bertschneider: Class names _could_ be lowercase and package names _could_ be uppercase. Probably no one is going to do that, though. I think the current approach is a little bit more on the safe side.

fmeum: True, especially for other JVM-based languages.

fmeum (approved):

PR #389, opened 2022-04-13
Prevent `ConcurrentModificationException` in `TraceCmpHooks`

bertschneider:
Applying `TraceCmpHooks::mapGet` to `LinkedHashMaps` can lead to `ConcurrentModificationExceptions`. 
As stated in its JavaDoc: 

> In access-ordered linked hash maps, merely querying the map with get is a structural modification.

This can easily cause concurrent modification exceptions as seen during startup of Spring Boot applications.
Strack trace of such on exception:
```
11:25:12.402 [main] ERROR org.springframework.boot.SpringApplication - Application run failed
java.util.ConcurrentModificationException: null
        at java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java:719)
        at java.util.LinkedHashMap$LinkedKeyIterator.next(LinkedHashMap.java:742)
        at com.code_intelligence.jazzer.runtime.TraceCmpHooks.mapGet(TraceCmpHooks.java:308)
        at org.springframework.boot.loader.jar.JarFileEntries.getEntry(JarFileEntries.java:312)
        at org.springframework.boot.loader.jar.JarFileEntries.getEntry(JarFileEntries.java:300)
        at org.springframework.boot.loader.jar.JarFileEntries.doGetEntry(JarFileEntries.java:288)
        at org.springframework.boot.loader.jar.JarFileEntries.getEntry(JarFileEntries.java:243)
        at org.springframework.boot.loader.jar.JarFileEntries.getEntry(JarFileEntries.java:203)
        at org.springframework.boot.loader.jar.JarFile.getEntry(JarFile.java:221)
        at org.springframework.boot.loader.LaunchedURLClassLoader.lambda$definePackage$0(LaunchedURLClassLoader.java:137)
        at java.security.AccessController.doPrivileged(Native Method)
        at org.springframework.boot.loader.LaunchedURLClassLoader.definePackage(LaunchedURLClassLoader.java:129)
        at org.springframework.boot.loader.LaunchedURLClassLoader.definePackageIfNecessary(LaunchedURLClassLoader.java:111)
        at org.springframework.boot.loader.LaunchedURLClassLoader.loadClass(LaunchedURLClassLoader.java:80)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
[...]
```

### DIFF
--- a/agent/src/main/java/com/code_intelligence/jazzer/runtime/TraceCmpHooks.java
+++ b/agent/src/main/java/com/code_intelligence/jazzer/runtime/TraceCmpHooks.java
@@ -18,6 +18,7 @@ import com.code_intelligence.jazzer.api.HookType;
 import com.code_intelligence.jazzer.api.MethodHook;
 import java.lang.invoke.MethodHandle;
 import java.util.Arrays;
+import java.util.ConcurrentModificationException;
 import java.util.Map;
 import java.util.TreeMap;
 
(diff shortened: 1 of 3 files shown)


agent/src/main/java/com/code_intelligence/jazzer/runtime/TraceCmpHooks.java
```diff
@@ -301,6 +303,19 @@ public static void mapGet(
         // compared to the maps keys.
       }
     } else if (currentKey instanceof Comparable) {
+      if (map instanceof LinkedHashMap) {
```
fmeum, line 306: Is this specific to `LinkedHashMap` or are there possibly other implementations with this behavior? Should we wrap the entire hook into a try...catch block?

agent/src/test/java/com/code_intelligence/jazzer/runtime/TraceCmpHooksTest.java
```diff
@@ -0,0 +1,58 @@
...
+      ES.submit(put.apply(i));
+    }
+
+    // Call hook
+    try {
+      for (int i = 0; i < 1_000; i++) {
+        TraceCmpHooks.mapGet(null, map, new Object[] {arg}, 1, null);
+      }
+    } catch (UnsatisfiedLinkError ignored) {
+      // Expected error due to missing native library,
```
fmeum, line 50: This should be fixable with `MockDriver`. If it isn't, I would like to look into why that's the case.

agent/src/test/java/com/code_intelligence/jazzer/runtime/BUILD.bazel
```diff
@@ -10,3 +12,16 @@ java_test(
         "@maven//:junit_junit",
     ],
 )
+
+java_test(
+    name = "TraceCmpHooksTest",
+    srcs = [
+        "TraceCmpHooksTest.java",
+    ],
+    target_compatible_with = SKIP_ON_WINDOWS,
```
fmeum, line 21: Is this needed because of the mock driver?

bertschneider: Yes, my understanding is that the mock driver is not properly running on windows. Did I get that wrong?
The test failed on windows only.

fmeum: You are right. Let's keep this as is and track the proper fix to that problem in #376.

fmeum (approved):

PR #396, opened 2022-04-26
Hierarchically budget fuzzer bytes with Autofuzz 

fmeum:
Currently, whenever Autofuzz needs to decide how many bytes from the
fuzzer input to spend on the construction of an object, it uses half of
the remaining bytes. While this ensures that primitives with non-trivial
values can be generated with arbitrarily many nested object
constructions, this strategy does not make good use of the fuzzer input
in the case of flat call graphs (e.g. a foo(String, int) method).
Experiments show that ignoring half of the fuzzer input hurts efficiency
quite a bit.

This commit switches to a new strategy that hierarchically assigns a
"budget" for every consume or autofuzz call:

1. Every call to consume or autofuzz is allocated a fraction of the
   remaining bytes and assigns equal parts of this budget to the
   recursive invocations it makes.
2. By constructing the objects that consume a bounded number of bytes
   first in autofuzz, the remaining budget for an autofuzz invocation
   can be evenly distributed among the variable-length objects without
   risking that primitive arguments are forced to trivial values.

### DIFF
--- a/agent/BUILD.bazel
+++ b/agent/BUILD.bazel
@@ -22,7 +22,9 @@ strip_jar(
     out = "jazzer_agent_deploy.jar",
     jar = ":jazzer_agent_shaded_deploy",
     paths_to_strip = [
+        "LICENSE",
         "module-info.class",
+        "THIRD-PARTY",
     ],
     visibility = ["//visibility:public"],
 )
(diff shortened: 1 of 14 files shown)


fmeum:
To be superseded by the upcoming structured mutation engine.

PR #402, opened 2022-04-28
Update windows runner in release workflow

bertschneider:
`windows-2016` runners were removed (https://github.com/actions/virtual-environments/issues/4312).


fmeum:
The idea behind the choice of release runners is that the binaries we generate with them should be as compatible as possible. Ideally, this would mean using the oldest still supported version of each OS.

fmeum (approved):

PR #403, opened 2022-04-28
Remove JDK debug output in workflows

bertschneider:


fmeum:
How about we instead delete the entire "Print JDK home" step, both here and in the test workflow? We essentially only introduced it because I messed up the Bazel 5 migration once, but haven't seen any issues with it since. Saves us two places that we would otherwise have to keep things in sync.

bertschneider:
Sure, even better. I didn't know if the output still provided some value but removing it altogether is much cleaner.

fmeum (approved):

PR #407, opened 2022-05-05
Move honeypot class to API

bertschneider:
The honeypot class jaz.Zer is needed to successfully execute reproducers
of deserialization issues. It's moved to the API artifact to be
available on the classpath.

Furthermore, reproducers not necessarily reproduce the found issue,
e.g. due to global state that leads to different behavior. To take that
into account the reproducer check is relaxed.


agent/src/main/java/com/code_intelligence/jazzer/api/BUILD.bazel
```diff
@@ -1,5 +1,14 @@
-java_library(
+java_import(
```
fmeum, line 1: We should be able to get around this `java_library`/`java_import` "hack" by exporting the `jaz.Zer` and `jaz.Ter` sources via a `filegroup` and referencing that in the sources of `java_library`.

bertschneider: Ah, yes, it's much better that way! Done.

fmeum:
@bertschneider The tests that previously couldn't reproduce findings now can and need updating. Worth it :-)

bertschneider:
Great, but also strange, because everything worked locally(tm). I will take a look.

fmeum:
> Great, but also strange, because everything worked locally(tm). I will take a look.

You were probably missing `--config=ci`.

fmeum:
@bertschneider The remaining failure is interesting for two reasons: It seems that our sanitizers weren't compatible with Java 8, but we only noticed now. And the line `INFO: Instrumented com.azul.tooling.Handler$1 (took 36 ms, size +0%)` indicates that we should probably revisit our exclusion list.

bertschneider:
Yes, indeed. I wrote a function mimicking `readNBytes` to make it work on Java 8 again. The excludes should be handled in a different PR, though.

PR #414, opened 2022-06-03
Remove redundant -ldl linkopt

fmeum:
We no longer use dlsym in the driver and thus don't have to link with
-ldl.


fmeum:
CI should be fixed by https://github.com/CodeIntelligenceTesting/jazzer/pull/415

bertschneider (approved):

PR #417, opened 2022-06-17
Add hook to detect SQL injections

simonresch:
This PR adds a hook that detects SQL injections by validating the query syntax.


sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/SqlInjection.kt
```diff
@@ -0,0 +1,165 @@
...
+            }
+            throw throwable
+        }
+    }
+
+    private fun isValidSql(sql: String): Boolean =
+        try {
+            CCJSqlParserUtil.parseStatements(sql)
+            true
+        } catch (e: JSQLParserException) {
```
bertschneider, line 162: Should we add another catch block handling `Exception`, or even `Throwable`, to prevent exceptions thrown in the library from bubbling up into application code?

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/SqlInjection.kt
```diff
@@ -0,0 +1,165 @@
...
+        MethodHook(type = HookType.REPLACE, targetClassName = "java.sql.Statement", targetMethod = "executeBatch"),
+        MethodHook(type = HookType.REPLACE, targetClassName = "java.sql.Statement", targetMethod = "executeLargeBatch"),
+        MethodHook(
+            type = HookType.REPLACE,
+            targetClassName = "java.sql.Statement",
+            targetMethod = "executeLargeUpdate"
+        ),
+        MethodHook(type = HookType.REPLACE, targetClassName = "java.sql.Statement", targetMethod = "executeQuery"),
+        MethodHook(type = HookType.REPLACE, targetClassName = "java.sql.Statement", targetMethod = "executeUpdate"),
+
```
bertschneider, line 69: Remove empty line?

simonresch: Yes, fixed.

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/SqlInjection.kt
simonresch: Good point! I added a catch block that catches all `Throwable`.

bertschneider (approved):

PR #425, opened 2022-07-27
Prepare for the FuzzTargetRunner Java rewrite

fmeum:


agent/src/main/java/com/code_intelligence/jazzer/runtime/FuzzedDataProviderImpl.java
```diff
@@ -19,6 +19,17 @@
 public class FuzzedDataProviderImpl implements FuzzedDataProvider {
   public FuzzedDataProviderImpl() {}
 
+  private static native void nativeInit();
```
bertschneider, line 22: I like that `FuzzedDataProviderImpl` registers the native functions itself now.

bertschneider (approved):

PR #434, opened 2022-08-08
driver: Split libfuzzer_{driver,fuzz_target} into main and library

fmeum:
By using LLVMFuzzerRunDriver, we can remove the LibfuzzerDriver class
and replace it by:

* a real main function that preprocesses arguments for libFuzzer and
  starts a JVM;
* everything else turned into a proper library with no dependency on our
  custom JVM class.

This will allow us to turn everything except the main function into a
JNI shared library, opening up the possibility to launch Jazzer from an
already running JVM.


driver/fuzz_target_runner.cpp
```diff
@@ -12,75 +12,179 @@
...
+void DumpJvmStackTraces() {
+  JavaVM *vm;
+  jsize num_vms;
+  JNI_GetCreatedJavaVMs(&vm, 1, &num_vms);
+  if (num_vms != 1) {
+    return;
+  }
+  JNIEnv *env = nullptr;
+  if (vm->AttachCurrentThread(reinterpret_cast<void **>(&env), nullptr) !=
+      JNI_OK) {
```
simonresch, line 75: Are there cases where this is expected to fail? Do you think it makes sense log a warning here?

simonresch:
Very nice! The restructuring makes sense to me and using `LLVMFuzzerRunDriver` instead of the `LibfuzzerDriver` classes is a good idea.

simonresch (approved):

driver/fuzz_target_runner.cpp
fmeum: I don't know, I added a warning to be on the safe side.

PR #439, opened 2022-08-11
Add support for Tomcat and Jakarta Expression Language Injection Bug Detectors.

henryrneh:
I added the support for org.apache.el.ExpressionFactoryImpl and jakarta.el.ExpressionFactory in Expression Language Injection Detector. With the modification we can catch Expression Language Injection within Tomcat Applications.

### DIFF
--- a/sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ExpressionLanguageInjection.kt
+++ b/sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ExpressionLanguageInjection.kt
@@ -44,6 +44,16 @@ object ExpressionLanguageInjection {
             targetClassName = "javax.el.ExpressionFactory",
             targetMethod = "createMethodExpression",
         ),
+        MethodHook(
+            type = HookType.BEFORE,
+            targetClassName = "jakarta.el.ExpressionFactory",
+            targetMethod = "createValueExpression",
+        ),
+        MethodHook(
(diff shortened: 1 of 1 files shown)


sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ExpressionLanguageInjection.kt
```diff
@@ -44,6 +44,26 @@ object ExpressionLanguageInjection {
             targetClassName = "javax.el.ExpressionFactory",
             targetMethod = "createMethodExpression",
         ),
+        MethodHook(
+            type = HookType.BEFORE,
+            targetClassName = "org.apache.el.ExpressionFactoryImpl",
```
fmeum, line 49: According to https://tomcat.apache.org/tomcat-8.0-doc/api/org/apache/el/ExpressionFactoryImpl.html, this class inherits from `javax.el.ExpressionFactory`, which we already hook with the annotations above. Hooks descend to subclasses, so this Apache-specific hook shouldn't be needed.

Could you check whether you can drop it and still get findings?

henryrneh: Yes you are right, I can still get the findings without the redundant ExpressionFactoryImpl. I already update my commit, and thank you for your insights for remove the redundant code!

fmeum (approved):

PR #442, opened 2022-08-15
driver: Set a default -rss_limit_mb

fmeum:
This is necessary for a pure Java driver as we can no longer set -Xmx in
that situation. It is also much cleaner than hand-tuning -Xmx, but we
still keep the max heap size in the native driver for backwards
compatibility with existing crashing inputs.


simonresch (approved):
Very nice! :)

PR #443, opened 2022-08-15
Xpath sanitizer

SyrasX:
Added XPath Sanitizer + simple test case

### DIFF
--- a/sanitizers/sanitizers.bzl
+++ b/sanitizers/sanitizers.bzl
@@ -24,6 +24,7 @@ _sanitizer_class_names = [
     "RegexInjection",
     "RegexRoadblocks",
     "SqlInjection",
+    "XPathInjection",
 ]
 
 SANITIZER_CLASSES = [_sanitizer_package_prefix + class_name for class_name in _sanitizer_class_names]
--- a/sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/BUILD.bazel
+++ b/sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/BUILD.bazel
@@ -32,6 +32,7 @@ kt_jvm_library(
         "ReflectiveCall.kt",
         "RegexInjection.kt",
         "Utils.kt",
+        "XPathInjection.kt",
     ],
     visibility = ["//sanitizers:__pkg__"],
     runtime_deps = [
--- a/sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/XPathInjection.kt
+++ b/sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/XPathInjection.kt
@@ -1,0 +1,78 @@
+// Copyright 2022 Code Intelligence GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.code_intelligence.jazzer.sanitizers
+
+import com.code_intelligence.jazzer.api.FuzzerSecurityIssueHigh
+import com.code_intelligence.jazzer.api.HookType
+import com.code_intelligence.jazzer.api.Jazzer
+import com.code_intelligence.jazzer.api.MethodHook
+import com.code_intelligence.jazzer.api.MethodHooks
+import java.lang.invoke.MethodHandle
+import javax.xml.xpath.XPathExpressionException
+
+/**
+ * Detects XPath injections.
+ *
+ * Untrusted input has to be escaped in such a way that queries remain valid, otherwise an injection
+ * could be possible. This sanitizer guides the fuzzer to inject insecure characters. If an exception
+ * is raised during execution the fuzzer was able to inject an invalid pattern, otherwise all input
+ * was escaped correctly.
+ * Checking if the innermost cause of XPathExpressionException is a TransformerException should
+ * indicate injection instead of a false positive.
+ */
+@Suppress("unused_parameter", "unused")
+object XPathInjection {
+
+    // Characters that should be escaped in user input.
+    // https://owasp.org/www-community/attacks/XPATH_Injection
+    private const val CHARACTERS_TO_ESCAPE = "'\""
+
+    private val XPATH_SYNTAX_ERROR_EXCEPTIONS = "javax.xml.transform.TransformerException"
+
+    @MethodHooks(
+        MethodHook(type = HookType.REPLACE, targetClassName = "javax.xml.xpath.XPath", targetMethod = "compile"),
+        MethodHook(type = HookType.REPLACE, targetClassName = "javax.xml.xpath.XPath", targetMethod = "evaluate"),
+        MethodHook(type = HookType.REPLACE, targetClassName = "javax.xml.xpath.XPath", targetMethod = "evaluateExpression"),
+    )
+    @JvmStatic
+    fun checkXpathExecute(method: MethodHandle, thisObject: Any?, arguments: Array<Any>, hookId: Int): Any {
+        if (arguments.isNotEmpty() && arguments[0] is String) {
+            val query = arguments[0] as String
+            Jazzer.guideTowardsContainment(query, CHARACTERS_TO_ESCAPE, hookId)
+        }
+        return try {
+            method.invokeWithArguments(thisObject, *arguments)
+        } catch (exception: XPathExpressionException) {
+            // find innermost cause
+            var innerCause = exception.cause
+            while (innerCause?.cause != null && innerCause.cause != innerCause) {
+                innerCause = innerCause.cause
+            }
+
+            if (innerCause != null && XPATH_SYNTAX_ERROR_EXCEPTIONS.equals(innerCause.javaClass.name)) {
+                Jazzer.reportFindingFromHook(
+                    FuzzerSecurityIssueHigh(
+                        """
+                    XPath Injection
+                    Injected query: ${arguments[0]}
+                        """.trimIndent(),
+                        exception,
+                    ),
+                )
+            }
+            throw exception
+        }
+    }
+}
--- a/sanitizers/src/test/java/com/example/BUILD.bazel
+++ b/sanitizers/src/test/java/com/example/BUILD.bazel
@@ -197,3 +197,16 @@ java_test(
         "//src/main/java/com/code_intelligence/jazzer/api:hooks",
     ],
 )
+
+java_fuzz_target_test(
+    name = "XPathInjection",
+    srcs = [
+        "XPathInjection.java",
+    ],
+    allowed_findings = [
+        "com.code_intelligence.jazzer.api.FuzzerSecurityIssueHigh",
+    ],
+    target_class = "com.example.XPathInjection",
+    # Fuzz target catches the syntax exception triggered by the reproducer without the sanitizer.
+    verify_crash_reproducer = False,
+)
--- a/sanitizers/src/test/java/com/example/XPathInjection.java
+++ b/sanitizers/src/test/java/com/example/XPathInjection.java
@@ -1,0 +1,53 @@
+// Copyright 2022 Code Intelligence GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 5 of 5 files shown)


sanitizers/sanitizers.bzl
```diff
@@ -24,6 +24,7 @@ _sanitizer_class_names = [
     "RegexInjection",
     "RegexRoadblocks",
     "SqlInjection",
+    "XpathInjection",
```
fmeum, line 27: You can either format via `./format.sh` locally or apply the diff printed by the [failing CI job](https://github.com/CodeIntelligenceTesting/jazzer/runs/7841422789?check_suite_focus=true#step:5:15).

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/XpathInjection.kt
```diff
@@ -0,0 +1,95 @@
...
+import com.code_intelligence.jazzer.api.Jazzer
+import com.code_intelligence.jazzer.api.MethodHook
+import com.code_intelligence.jazzer.api.MethodHooks
+import java.lang.invoke.MethodHandle
+
+/**
+ * Detects Xpath injections.
+ *
+ * Untrusted input has to be escaped in such a way that queries remain valid otherwise an injection
+ * could be possible. This sanitizer guides the fuzzer to inject insecure characters. If an exception
```
fmeum, line 28: I'm not familiar with Xpaths - could you explain why an injection would be problematic? What would be the possible impact?

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/XpathInjection.kt
```diff
@@ -0,0 +1,95 @@
...
+ * is raised during execution the fuzzer was able to inject an invalid pattern, otherwise all input
+ * was escaped correctly.
+
+ * False positives can be possible but should bugs in the code just not an injection. e.g.
+ * "If expression contains any XPathFunctions, they must be available via the XPathFunctionResolver. An XPathExpressionException will be thrown if the XPathFunction cannot be resovled with the XPathFunctionResolver."
+ * "The parameter item represents the context the XPath expression will be operated on. The type of the context is implementation-dependent. If the value is null, the operation must have no dependency on the context, otherwise an XPathExpressionException will be thrown."
+ * https://docs.oracle.com/en/java/javase/17/docs/api/java.xml/javax/xml/xpath/XPath.html
+ * 
+ * Checking the number of quotation marks to be uneven should indicate injection instead of a false positive
```
fmeum, line 37: Could you explain what such a false positive would look like? Why are quotation marks relevant for this?

fmeum (commented):
Thanks for the interesting contribution!

I left some general questions as I am not very familiar with the subject this sanitizer applies to.

sanitizers/sanitizers.bzl
SyrasX: Yeah sorry for that, somehow format.sh changes something in pretty much every file of the project so I assumed it does not work correctly for me. Applied the diff changes hope that it's correct.

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/XpathInjection.kt
SyrasX: Xpath is used to query XML. Kinda similar to SQL Injection. If quotes in the user input are not sanitized they can be used to inject into the Xpath expression. This allows for example to crawl the XML file or skip password checks.
Examples here:
https://owasp.org/www-community/attacks/XPATH_Injection
https://owasp.org/www-community/attacks/Blind_XPath_Injection

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/XpathInjection.kt
SyrasX: The XPathExpressionException is kinda generall and can have many reasons. Like for example "javax.xml.xpath.XPathExpressionException: java.io.IOException: InputStreamReader is closed" Since this is hopefully expected and dealt with in the fuzzed app and also is no injection, that would be a false positive. To prevent them I thought of just checking the number of single and double quotes in the expression. From my understanding they should always be even, so if they are not it is an possible injection. The alternative is that the app adds an additional quote somewhere which throws the exception. While that would not be an injection it still is a bug and should only happen rarely.

I checked some small test examples and it works correctly for them. Naturally there can be cases that I have not considered. Since I started reading about Xpath just for this sanitizer so I would not call myself an expert ^^

sanitizers/sanitizers.bzl
fmeum: Oops, that's not nice. Could you give me an example? Then I will try to fix it.

SyrasX: I'm sure my configuration is at fault. It for example removed every comma of the last element e.g. 

 "RegexRoadblocks",
 "SqlInjection",
-"XpathInjection",
+"XpathInjection"

I now installed everything without homebrew and it seems not work.

fmeum:
@SyrasX Sorry for the long wait, I have been quite busy finishing up some larger refactorings and feature additions. I will have to brush up my knowledge of XPath and XML in general for a proper review. I hope to get to that by next week.

SyrasX:
@fmeum No worries take as much time as you need. Maybe some additional info: 

The current approch will not detect an injection if quotation marks in input are sanitized but the programm allows the input to directly build a querry instead of using somethin like this "//input[@id='firstName']". I personnaly would rate it rather unlikely.

The original plan was to use the Xpath.compile() function to check  for correct syntax in a similar way to the SQL Injection Sanitizer. I just fear that since the XPathExpressionException  is thrown because of reason not strictly releted to syntax, that we would report all the normally catched  and handled exceptions without them being an injection.

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/XpathInjection.kt
```diff
@@ -0,0 +1,92 @@
...
+    fun checkXpathExecute(method: MethodHandle, thisObject: Any?, arguments: Array<Any>, hookId: Int): Any {
+
+        if (arguments.isNotEmpty() && arguments[0] is String) {
+            val query = arguments[0] as String
+            Jazzer.guideTowardsContainment(query, CHARACTERS_TO_ESCAPE, hookId)
+        }
+        return try {
+            method.invokeWithArguments(thisObject, *arguments)
+        } catch (throwable: Throwable) {
+            if (XPATH_SYNTAX_ERROR_EXCEPTIONS.contains(throwable.javaClass.name) && !isEvenNumberOfQuotationMarks(arguments[0] as String)) {
```
fmeum, line 66: I delved into the `java.xml` source code to find a better signal that a given `XPathExpressionException` actually arises from a syntax error.

The actual parsing happens in https://github.com/openjdk/jdk/blob/master/src/java.xml/share/classes/com/sun/org/apache/xpath/internal/compiler/XPathParser.java, which seems to always call the `error` function in case of a parsing error. This function creates a `TransformerException` at https://github.com/openjdk/jdk/blob/master/src/java.xml/share/classes/com/sun/org/apache/xpath/internal/compiler/XPathParser.java#L644, which code higher up the call stack wraps into an `XPathExpressionException`, together with essentially any other kind of exception (for example at https://github.com/openjdk/jdk/blob/e5edb10dc56d9edac8e050e0f8e6c116743975d6/src/java.xml/share/classes/com/sun/org/apache/xpath/internal/jaxp/XPathImpl.java#L174).

I briefly looked through the other occurrences of `new TransformerException` and it seems that we could just catch `XPathExpressionException` instead of `Throwable` and report a finding if the innermost cause of that exception is a `TransformerException`.

fmeum (commented):
Apologies for the long delay. I finally got to read up on XPaths and delve into the JDK implementation.

I rebased your branch and pushed a few small commits to get the tests to pass after the recent Jazzer changes.

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/XpathInjection.kt
SyrasX: While I did not take a closer look at this approach yet, the reason I had before to not dive deeper into OpenJDKs code was that I am not sure how standardized differend implimentation of the JDK are and if we could trust them to internally be/stay the same. If that is not a problem this approach seems to be much nicer than my solution.

fmeum: It looks like the OpenJDK implementation is just a vendored version of an Apache library, so I would assume other JDKs use the same one. While Jazzer itself tries to stay out of JDK internals, hook implementations can't realistically do this (see our regex hooks, which make extensive assumptions about the implementation). 

If something works well with OpenJDK I would say go with that and revisit once it leads to problems with other JDKs. In this particular case the exception type we would be testing for is part of the public API surface, so there is a good chance this will even work across JDKs.

SyrasX: Ok so after playing around with it, your proposed approach seems to work fine. The only problem I see right now is that the error() func which than creates the TransformerException is called from several places that don't necessary imply an injection. So there still could be some false positives. 

From my point of view the isEvenNumberOfQuotationMarks() approach before will prevent all false positives but not detect an injection if quotation marks in input are sanitized and the programm allows the input to directly build a querry.
The new approach probably detects all injections but has some false positives. I can't really assess how probable these are in production code.

What is your opinion ?

bertschneider: I also tried to figure out where exactly `TransformerException` are thrown, and as far as I can see they are always related to parsing, at least in `XPathParser`. Other classes throw `TransformException`s as well, though.

IMHO it's ok to assume that queries are generally valid and would pass internal checks. In that case we could check the exception type as @fmeum suggested. 

Which one would you prefer?

SyrasX: I would vote for the TransformerException approach then. It seems like a nicer solution. If there are some unexpected false positives, we could still fall back to the other approach.

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/XpathInjection.kt
```diff
@@ -0,0 +1,79 @@
...
+            var innerCause = exception.cause
+            while (innerCause?.cause != null && innerCause.cause != innerCause) {
+                innerCause = innerCause.cause
+            }
+
+            if (innerCause != null && XPATH_SYNTAX_ERROR_EXCEPTIONS.contains(innerCause.javaClass.name)) {
+                Jazzer.reportFindingFromHook(
+                    FuzzerSecurityIssueHigh(
+                        """
+                    Xpath Injection
```
bertschneider, line 69: Uppercase `P` in `XPath`.

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/XpathInjection.kt
```diff
@@ -0,0 +1,79 @@
...
+    )
+    @JvmStatic
+    fun checkXpathExecute(method: MethodHandle, thisObject: Any?, arguments: Array<Any>, hookId: Int): Any {
+        if (arguments.isNotEmpty() && arguments[0] is String) {
+            val query = arguments[0] as String
+            Jazzer.guideTowardsContainment(query, CHARACTERS_TO_ESCAPE, hookId)
+        }
+        return try {
+            method.invokeWithArguments(thisObject, *arguments)
+        } catch (exception: javax.xml.xpath.XPathExpressionException) {
```
bertschneider, line 58: Use a static import for the exception type?

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/XpathInjection.kt
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * Untrusted input has to be escaped in such a way that queries remain valid, otherwise an injection
+ * could be possible. This sanitizer guides the fuzzer to inject insecure characters. If an exception
+ * is raised during execution the fuzzer was able to inject an invalid pattern, otherwise all input
+ * was escaped correctly.
+ * Checking if the innermost cause of XPathExpressionException is a TransformerException should 
+ * indicate injection instead of a false positive.
+ */
+@Suppress("unused_parameter", "unused")
+object XpathInjection {
```
bertschneider, line 35: Uppercase `P` for a consistent naming with the other files/classes?

bertschneider (commented):
Great, thx again for your contribution! Beside minor nit-picks this PR is good to go :+1:

bertschneider:
You probably have to rebase on the latest `main` version and resolve the conflict to trigger a new pipeline run.

SyrasX:
> You probably have to rebase on the latest `main` version and resolve the conflict to trigger a new pipeline run.

I don't think I can even trigger a pipeline run myself. Also getting to run the format.sh correctly last time took some time, so I just hope it's fine. If not please just run it yourself.

0xricksanchez:
@SyrasX Pipeline runs now :).

bertschneider (approved):

SyrasX:
Done. Does it work now @bertschneider ?

bertschneider:
Merged. Thx again for the PR and your patience.

PR #444, opened 2022-08-16
driver: Load fuzz target with Driver's class loader 

fmeum:
This ensures that the fuzz target is found even if Driver is loaded by
a custom class loader.


fmeum:
@bertschneider This made my head hurt quite a bit. If you see a way we can get around this issue in a less intrusive way, please let me know. I documented why the first approach that comes to mind doesn't work.

bertschneider (approved):

PR #446, opened 2022-08-17
Add a test for expression language injection.

henryrneh:
I added an easy example/test for the expression language injection, not sure if it is needed, thank you!


fmeum:
We already have [this test](https://github.com/CodeIntelligenceTesting/jazzer/blob/dccc3ae7e8e213164dc2c8df36088a862878f9ba/sanitizers/src/test/java/com/example/BUILD.bazel#L33). If your test covers some other part of the sanitizer, you could add it to that directory instead - we don't want to grow the number of examples too large, hence the separate location.

You can run `./format.sh` locally to fix formatting issues or apply the diff printed by the CI job [here](https://github.com/CodeIntelligenceTesting/jazzer/runs/7874255687?check_suite_focus=true#step:5:10).

henryrneh:
Sorry I did not know about that repo, my test is generally the same as the one you already have there, so I will close this pull request, sorry for the inconvenience and thank you!

PR #452, opened 2022-08-19
all: Make global state in FuzzedDataProviderImpl local

fmeum:
An instance of FuzzedDataProviderImpl can now be one of the following:

* backed by an automatically created native copy of a Java array that is
  freed by `close()`;
* backed by a preexisting native array whose lifetime is managed
  elsewhere.

All instances now maintain their own internal, non-global state and can
thus be used concurrently.


driver/src/main/native/com/code_intelligence/jazzer/driver/fuzzed_data_provider.cpp
```diff
@@ -131,15 +121,21 @@ template <typename T>
...
-         gRemainingBytes != 0) {
-    --gRemainingBytes;
-    result = (result << 8u) | gDataPtr[gRemainingBytes];
+         remainingBytes != 0) {
+    --remainingBytes;
+    result = (result << 8u) | dataPtr[remainingBytes];
     offset += 8;
   }
 
+  env.SetIntField(self, gRemainingBytesField, remainingBytes);
```
bertschneider, line 137: I think I don't fully understand what's happening here. Don't we have to set the new position in `gDataPtrField` as well?

bertschneider:
Is the idea to eventually move the complete FuzzDataProvider functionality to Java or would the performance impact be too much?

agent/src/main/java/com/code_intelligence/jazzer/runtime/FuzzedDataProviderImpl.java
```diff
@@ -26,18 +27,107 @@ public class FuzzedDataProviderImpl implements FuzzedDataProvider {
...
+  /**
+   * Replaces the current native backing array.
+   *
+   * <p><b>Must not</b> be called on instances created with {@link #withJavaData(byte[])}.
+   *
+   * @param dataPtr a native pointer to the new backing array
+   * @param dataLength the length of the new backing array
+   */
+  public void setNativeData(long dataPtr, int dataLength) {
+    this.originalDataPtr = dataPtr;
```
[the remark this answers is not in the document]
fmeum: As this class is not part of the public API and the method is in the critical path for fuzz target executions, I would prefer to solve this via the javadocs only - if we throw an exception, the JVM has to generate code that handles it.

agent/src/main/java/com/code_intelligence/jazzer/runtime/FuzzedDataProviderImpl.java
```diff
@@ -26,18 +27,107 @@ public class FuzzedDataProviderImpl implements FuzzedDataProvider {
...
+   */
+  public void reset() {
+    dataPtr = originalDataPtr;
+    remainingBytes = originalRemainingBytes;
+  }
+
+  /**
+   * Releases native memory allocated for this instance (if any).
+   *
+   * <p>While the instance should not be used after this method returns, no usage if {@link
```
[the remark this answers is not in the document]
fmeum: Fixed the comment, thanks.

agent/src/main/java/com/code_intelligence/jazzer/runtime/FuzzedDataProviderImpl.java
```diff
@@ -26,18 +27,107 @@ public class FuzzedDataProviderImpl implements FuzzedDataProvider {
...
+    }
+    // Prevent double-frees and use-after-frees by effectively making all methods no-ops after
+    // close() has been called.
+    originalDataPtr = 0;
+    originalRemainingBytes = 0;
+    dataPtr = 0;
+    remainingBytes = 0;
+  }
+
+  private static final Unsafe UNSAFE = UnsafeProvider.getUnsafe();
```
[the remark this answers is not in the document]
fmeum: As it's only used in this one place, I would lean towards keeping it close to it. It is also somewhat tightly coupled to `fuzzed_data_provider.cpp`.

driver/src/main/native/com/code_intelligence/jazzer/driver/fuzzed_data_provider.cpp
fmeum: The loop above only reads from `dataPtr`, but doesn't modify it, so we don't have to update the field. I will add a comment.

fmeum (commented):
This makes it easier to port `FuzzedDataProvider` to Java in the future, but I don't know yet whether that would be a performance win - it certainly isn't necessary yet.

Rather, the aim of this PR is simply to make it safe for multiple `FuzzedDataProviderImpl` instances to exist in parallel, which is important for the upcoming JUnit integration.

bertschneider (approved):

driver/src/main/native/com/code_intelligence/jazzer/driver/fuzzed_data_provider.cpp
bertschneider: So we take the bytes we need from the end of the array and just modify `remainingBytes` accordingly? But why is this handled differently to the other methods?

fmeum: Bulk data is taken from the front, bytes used to generate short primitive types are taken from the end. The idea is that of you are fuzzing a jpeg parser, your Corpus entries should start with jpeg data even if you consume some additional bytes to e.g. select parser settings. The original FuzzedDataProvider did this and I just copied the logic

PR #455, opened 2022-08-24
JUnit regression test

fmeum:


fmeum:
@bertschneider You can try out the example in VSC and IntelliJ. I will flesh out the example documentation some more when the actual fuzzing engine has been implemented.

fmeum:
@bertschneider I can't get my `jazzer.instrument` configuration parameter to be picked up. This worked before at some point, but doesn't anymore. If you have any idea why or get it working in some IDE, let me know.

fmeum:
I pushed one more commit to ensure that regression tests report the same findings as Jazzer, but didn't touch any of the other commits.

driver/src/main/java/com/code_intelligence/jazzer/junit/RegressionTestArgumentProvider.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.jar.JarInputStream;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.ArgumentsProvider;
+import org.junit.jupiter.params.support.AnnotationConsumer;
+
+class RegressionTestArgumentProvider implements ArgumentsProvider, AnnotationConsumer<FuzzTest> {
+  FuzzTest annotation;
```
bertschneider, line 49: Is this visibility required? Otherwise I would prefer `private`.

driver/src/main/java/com/code_intelligence/jazzer/junit/RegressionTestArgumentProvider.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+  }
+
+  private Stream<Map.Entry<String, byte[]>> walkSeedCorpus(Class<?> testClass) throws IOException {
+    String seedCorpusArg = annotation.seedCorpus();
+    String seedCorpusPath =
+        seedCorpusArg.isEmpty() ? Utils.defaultSeedCorpusPath(testClass) : seedCorpusArg;
+    URL seedCorpusUrl = testClass.getResource(seedCorpusPath);
+    if (seedCorpusUrl == null) {
+      return Stream.of();
```
bertschneider, line 87: Use `Stream.empty()`?

driver/src/main/java/com/code_intelligence/jazzer/junit/RegressionTestArgumentProvider.java
```diff
@@ -0,0 +1,135 @@
...
+    Stream<Map.Entry<String, byte[]>> rawSeeds =
+        Stream.concat(Stream.of(new SimpleEntry<>("<empty input>", new byte[] {})),
+            walkSeedCorpus(extensionContext.getRequiredTestClass()));
+    return adaptSeedsForFuzzTest(extensionContext.getRequiredTestMethod(), rawSeeds);
+  }
+
+  private Stream<? extends Arguments> adaptSeedsForFuzzTest(
+      Method fuzzTestMethod, Stream<Map.Entry<String, byte[]>> rawSeeds) {
+    if (fuzzTestMethod.getParameterCount() != 1) {
+      throw new IllegalArgumentException();
```
bertschneider, line 68: Use a descriptive message like in the case of a wrong parameter type? Also a test for this case would be great :-)

driver/src/main/java/com/code_intelligence/jazzer/junit/RegressionTestExtensions.java
```diff
@@ -0,0 +1,77 @@
...
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.InvocationInterceptor;
+import org.junit.jupiter.api.extension.ReflectiveInvocationContext;
+
+class RegressionTestExtensions
+    implements BeforeEachCallback, AfterEachCallback, InvocationInterceptor {
+  private static Field lastFindingField;
+
+  @Override
+  public void beforeEach(ExtensionContext extensionContext) {
```
bertschneider, line 31: Wouldn't it make sense to use `BeforeAll` as these functions can only be executed effectively ones?

driver/src/main/java/com/code_intelligence/jazzer/junit/RegressionTestExtensions.java
```diff
@@ -0,0 +1,77 @@
...
+  @Override
+  public void afterEach(ExtensionContext extensionContext) {
+    extensionContext.publishReportEntry(
+        "No fuzzing has been performed, the fuzz test has only been executed on the fixed set of inputs in the "
+        + "seed corpus.\n"
+        + "To start fuzzing, run a test with the environment variable JAZZER_FUZZ set to a non-empty value.");
+  }
+
+  private static Field getLastFindingField() throws Throwable {
+    if (lastFindingField == null) {
```
bertschneider, line 70: Why not simply initialize the field in a static block?

driver/src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetFinder.java
```diff
@@ -0,0 +1,193 @@
...
+          clazz.getName()));
+    }
+
+    // Both the constructor and the method may not be accessible - JUnit test classes and methods
+    // are usually declared without access modifiers and thus package-private.
+    method.setAccessible(true);
+    constructor.setAccessible(true);
+
+    return Optional.of(new FuzzTarget(parameter == FuzzedDataProvider.class, method,
+        constructor::newInstance, targetPublicStaticMethod(clazz, FUZZER_TEAR_DOWN)));
```
bertschneider, line 140: Mixing annotation based configuration and "name based" configuration with `fuzzerTearDown` feels strange. Should we skip the tear down function in this case, or use `@AfterAll` (multiple methods could be annotated)?

driver/src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetFinder.java
```diff
@@ -0,0 +1,193 @@
...
+                    .map(init -> (Callable<Object>) () -> {
+                      init.invoke(null, (Object) Opt.targetArgs.toArray(new String[] {}));
+                      return null;
+                    }),
+                targetPublicStaticMethod(clazz, FUZZER_INITIALIZE)
+                    .map(init -> (Callable<Object>) () -> {
+                      init.invoke(null);
+                      return null;
+                    }),
+                Optional.of((Callable<Object>) () -> null))
```
bertschneider, line 170: Remove "Null-Optional" and `getOrElse` as termination operation? Took me some time to understand the IntelliJ warning.

bertschneider (commented):
Great PR!

At some places it already hints at JUnit fuzzing support but that's not included yet. Looking forward to the next review :smile:

driver/src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetFinder.java
fmeum: Agreed, we should use JUnit concepts here in the future. I have rarely seen uses for `fuzzerTearDown` (if any at all?), so I will leave it out for now.

driver/src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetFinder.java
fmeum: Good catch, simplified this a bit, but maybe you see a way to simplify it even further.

driver/src/main/java/com/code_intelligence/jazzer/junit/FuzzTest.java
```diff
@@ -0,0 +1,68 @@
...
+ * href="https://github.com/CodeIntelligenceTesting/jazzer">Jazzer</a>.
+ *
+ * When executed as a JUnit Jupiter test, the fuzz test is executed on all files in the seed corpus
+ * (see documentation for the {@code seedCorpus} parameter.
+ *
+ * <p>Methods annotated with {@link FuzzTest} must take a single parameter of type {@code byte[]} or
+ * {@link com.code_intelligence.jazzer.api.FuzzedDataProvider}. The latter provides convenience
+ * methods to generate common Java types from the raw fuzzer input.
+ *
+ * <p>Only a single method per class can be annotated with {@link FuzzTest}.
```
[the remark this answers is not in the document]
fmeum: It will be enforced by the Jazzer engine, but it would be better if we could also enforce it on the JUnit level - I didn't find a good way to do that though.

I will take a stab at making multiple `@FuzzTest` work instead - shouldn't be that difficult.

driver/src/main/java/com/code_intelligence/jazzer/junit/RegressionTestArgumentProvider.java
fmeum: I think it's not, added.

driver/src/main/java/com/code_intelligence/jazzer/junit/RegressionTestArgumentProvider.java
fmeum: Added the same message and a test.

driver/src/main/java/com/code_intelligence/jazzer/junit/RegressionTestArgumentProvider.java
fmeum: Done

driver/src/main/java/com/code_intelligence/jazzer/junit/RegressionTestExtensions.java
fmeum: I tried that before and found it not to work. I think it's because we would have to specify the `BeforeAllCallback` on the class rather than the method, but we don't control that.

driver/src/main/java/com/code_intelligence/jazzer/junit/RegressionTestExtensions.java
fmeum: That doesn't work because the field lives in the bootstrap jar, which is only available after `beforeEach` has been called once. I added a comment since that's rather non-obvious.

driver/src/test/java/com/code_intelligence/jazzer/junit/RegressionTestTest.java
```diff
@@ -0,0 +1,81 @@
...
+  @Test
+  public void regressionTest() {
+    EngineExecutionResults results =
+        EngineTestKit.engine("junit-jupiter")
+            .selectors(selectPackage("com.example"))
+            .configurationParameter("jazzer.instrument",
+                "com.other.package.**,com.example.**,com.yet.another.package.*")
+            .execute();
+
+    if (!System.getenv("JAZZER_FUZZ").isEmpty()) {
```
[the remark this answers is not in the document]
fmeum: It doesn't quite work that way since Bazel's test runner is still based on JUnit 4 (although there is https://github.com/bazel-contrib/rules_jvm#java_junit5_test), but the split also makes sense with JUnit 4.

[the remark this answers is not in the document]
fmeum: It's funny that I've read the entire JUnit user guide looking for features to use for `@FuzzTest` and then forget about them right when I start writing tests myself.

fmeum:
@bertschneider Thanks for the review! I pushed the commits that should address your comments as new fix-up commits. I will look into supporting multiple `@FuzzTest` annotations.

driver/src/main/java/com/code_intelligence/jazzer/junit/FuzzTest.java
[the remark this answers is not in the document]
fmeum: Was easier than expected, done in a follow-up commit. This will also simplify the Jazzer test engine design.

fmeum:
> @bertschneider I can't get my `jazzer.instrument` configuration parameter to be picked up. This worked before at some point, but doesn't anymore. If you have any idea why or get it working in some IDE, let me know.

This was just me being stupid: The `junit-platform.properties` file has to be in the classpath root, not the project root.

driver/src/main/java/com/code_intelligence/jazzer/junit/RegressionTestExtensions.java
bertschneider: Ah, I see. Thx for the comment.

driver/src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetFinder.java
bertschneider: By the way, shouldn't `@BeforeXX` and `AfterXX` work out of the box?

driver/src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetFinder.java
bertschneider: `Optional`s don't work well with `Stream`s. In other functional languages a `flatMap` should work nicely. I think it's good enough now :smile:

bertschneider:
> This was just me being stupid: The junit-platform.properties file has to be in the classpath root, not the project root.

Ah, yes :+1:

bertschneider (approved):

driver/src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetFinder.java
fmeum: They work out of the box for the regression test, but not if Jazzer executes the fuzz test - after all `@Before/AfterXXX` is part of the Jupiter engine, not the JUnit platform.

driver/src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetFinder.java
fmeum: That would be possible in Java as well, but both https://docs.oracle.com/javase/9/docs/api/java/util/Optional.html#stream-- and https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Optional.html#or(java.util.function.Supplier) require Java 9.

PR #458, opened 2022-08-29
driver: Find @FuzzTests even if @FuzzTest is not on the classpath

fmeum:
This makes it much easier to execute fuzz tests with Jazzer from build
systems such as Maven in which it is rather difficult to build a "fat"
JAR for tests.

### DIFF
--- a/driver/src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetFinder.java
+++ b/driver/src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetFinder.java
@@ -16,17 +16,29 @@
 
 package com.code_intelligence.jazzer.driver;
 
+import static java.util.stream.Collectors.toList;
+
 import com.code_intelligence.jazzer.api.FuzzedDataProvider;
 import com.code_intelligence.jazzer.utils.ManifestUtils;
+import java.io.IOException;
+import java.io.InputStream;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
+import java.util.AbstractMap;
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.Callable;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import org.objectweb.asm.AnnotationVisitor;
+import org.objectweb.asm.ClassReader;
+import org.objectweb.asm.ClassVisitor;
+import org.objectweb.asm.MethodVisitor;
+import org.objectweb.asm.Opcodes;
+import org.objectweb.asm.Type;
 
 class FuzzTargetFinder {
   static class FuzzTarget {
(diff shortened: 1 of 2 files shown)


driver/src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetFinder.java
```diff
@@ -209,4 +213,39 @@ private static Optional<Method> targetPublicStaticMethod(
...
   }
+
+  // Returns a list of all methods annotated with @FuzzTest without requiring @FuzzTest to be on the
+  // class path.
+  private static List<Method> findFuzzTests(Class<?> clazz) {
+    // Stores pairs of the form (method name, method descriptor).
+    HashSet<Map.Entry<String, String>> annotatedMethods = new HashSet<>();
+    // If an annotation is not on the classpath, it is silently ignored when the annotated element
+    // is loaded. We thus need to load and parse the class file ourselves.
+    try (InputStream stream = FuzzTargetFinder.class.getResourceAsStream(
```
bertschneider, line 224: Is the `FuzzTargetFinder` class loader always able to lookup the given `Class` itself? If not the exception is thrown and the test simply fails, right?

driver/src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetFinder.java
```diff
@@ -209,4 +213,39 @@ private static Optional<Method> targetPublicStaticMethod(
       return Optional.empty();
     }
   }
+
+  // Returns a list of all methods annotated with @FuzzTest without requiring @FuzzTest to be on the
```
bertschneider, line 217: In which case is the annotation missing on the classpath?

fmeum: For example, when users package their tests into a JAR in Maven and pass it to the Jazzer CLI via `--cp`, they would be surprised to find that their fuzz tests aren't executing - they would have to list our JUnit integration as well as all its JUnit dependencies as well. This would make it very difficult to execute a `@FuzzTest`-based fuzz test on the command-line if not for this change.

driver/src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetFinder.java
fmeum: At this point it should be: `FuzzTargetRunner` looked up the class with its own class loader and found it, then passed the `Class` object to this method. Since `FuzzTargetFinder` is loaded by `FuzzTargetRunner`, I would expect it to find the class. That probably breaks with really weird class loaders (e.g. class loaders not based on `.class` files), but then I'm not sure we would support them that well elsewhere in Jazzer.

fmeum: I added the previous behavior as a fallback for this case.

bertschneider (approved):

PR #461, opened 2022-08-30
junit: Add support for coverage runs

fmeum:
When a known coverage agent is present, also let the regression test run
over the generated corpus, if it exists.


bertschneider (approved):
Looks good!

PR #470, opened 2022-09-06
ci: Add BuildBuddy metadata

fmeum:


fmeum:
Reading the setup guide again, this should really be picked up automatically from the GitHub actions env. I will investigate why it isn't.

bertschneider:
Yes, I also tried that out once in the past but couldn't get it to work and then other things got in the way. It would be great if you could find the problem.

Issue #465, opened 2022-09-01
Some reproducers do not compile


McLarney:
Hello, I'm fuzzing an old version of Jsoup as an exercise. I'm using the autofuzzer and outputting reproducers. I'm trying to run the reproducers individually and noticed that some of them don't compile. For example:

```
public class Crash_c8b1d033b996bea197a6c1d8e839fbdf99584532 {
  public static void main(String[] args) throws Throwable {
    org.jsoup.Jsoup.parse((java.lang.String) null, "\\b\\n\\n\\n", new org.jsoup.parser.Parser((org.jsoup.parser.TreeBuilder) ((java.util.function.Supplier<org.jsoup.parser.HtmlTreeBuilder>) (() -> {org.jsoup.parser.HtmlTreeBuilder autofuzzVariable0 = new org.jsoup.parser.HtmlTreeBuilder(); return autofuzzVariable0;})).get()));
  }
}
```
gives the error

> error: TreeBuilder is not public in org.jsoup.parser; cannot be accessed from outside package
    org.jsoup.Jsoup.parse((java.lang.String) null, "\\b\\n\\n\\n", new org.jsoup.parser.Parser((org.jsoup.parser.TreeBuilder) ((java.util.function.Supplier<org.jsoup.parser.HtmlTreeBuilder>) (() -> {org.jsoup.parser.HtmlTreeBuilder autofuzzVariable0 = new org.jsoup.parser.HtmlTreeBuilder(); return autofuzzVariable0;})).get()));

The target jar being fuzzed is included in the classpath when I try to compile reproducer. I am building from source and the target is jsoup's 1.13.1 release, before they fixed the bugs this tool found. 

If I manually change that TreeBuilder to XmlTreeBuilder it compiles. XmlTreeBuilder extends the abstract class TreeBuilder if that helps.

PR #471, opened 2022-09-06
autofuzz: Do not cast to non-public types

fmeum:
Fixes https://github.com/CodeIntelligenceTesting/jazzer/issues/465

### DIFF
--- a/driver/src/main/java/com/code_intelligence/jazzer/autofuzz/Meta.java
+++ b/driver/src/main/java/com/code_intelligence/jazzer/autofuzz/Meta.java
@@ -51,6 +51,9 @@ import net.jodah.typetools.TypeResolver;
 import net.jodah.typetools.TypeResolver.Unknown;
 
 public class Meta {
+  public static final boolean IS_DEBUG = isDebug();
+
+  private static final boolean IS_TEST = isTest();
   private static final WeakHashMap<Class<?>, List<Class<?>>> implementingClassesCache =
       new WeakHashMap<>();
   private static final WeakHashMap<Class<?>, List<Class<?>>> nestedBuilderClassesCache =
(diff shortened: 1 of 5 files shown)


driver/src/main/java/com/code_intelligence/jazzer/autofuzz/Meta.java
```diff
@@ -229,6 +233,10 @@ public static Object consume(FuzzedDataProvider data, Class<?> type) {
   // the genericType Class<java.lang.String> should lead to the generated code
   // "(java.lang.String) null", not just "null". This makes it possible to safely use consume in
   // recursive argument constructions.
+  // Exception: Some Java libraries offer public methods that take private interfaces or abstract
+  // classes as parameters. In this case, a cast to the parent type would case an
```
bertschneider, line 237: Typo at `case`.

bertschneider (approved):
Fixes the problem and describes the current limitations, approve :+1: 

I still wonder if we could relax the restriction. Even though the current approach should work in most cases, I could think about situations where someone wants to test internal functionality and has to use package visible classes. Moving the fuzz test into that package should still work if package visible classes are allowed. This opens up other problems, e.g. how should we know the "target" package. In writing this I realized we should probably leave it like it is now :smile:

fmeum:
> Fixes the problem and describes the current limitations, approve +1
> 
> I still wonder if we could relax the restriction. Even though the current approach should work in most cases, I could think about situations where someone wants to test internal functionality and has to use package visible classes. Moving the fuzz test into that package should still work if package visible classes are allowed. This opens up other problems, e.g. how should we know the "target" package. In writing this I realized we should probably leave it like it is now smile

That's a valid point once we enable `@FuzzTest` for Autofuzz (which I'm actually thinking about doing soon). I will need to think about it. We can certainly relax visibility checks, but would then have to emit code that makes certain classes or methods available via reflection.

PR #473, opened 2022-09-06
BREAKING: autofuzz: Do not enable --keep_going by default

fmeum:
Users reported that the implicit --keep_going of Autofuzz results in
poor UX as findings fly by unnoticed and it is not clear whether fuzzing
will ever stop by itself.

Instead, stop after the first finding as usual but print helpful
follow-up commands to skip a particular finding or a finding of a
particular type.

Stacked on #472


bertschneider (approved):
I like this approach. It fits to using autofuzz as an exploratory tool.

PR #476, opened 2022-09-07
junit: Make `@FuzzTest` methods autofuzzable

fmeum:


fmeum:
@bertschneider If you have time, you can already take a look. I am planning to refactor the static state out of `Meta` & `Utils` before I merge this, but that change will be boring.

driver/src/main/java/com/code_intelligence/jazzer/autofuzz/Meta.java
```diff
@@ -445,27 +454,31 @@ static Object consume(FuzzedDataProvider data, Type genericType, AutofuzzCodegen
...
         }
         try (ScanResult result = classGraph.scan()) {
           ClassInfoList children =
               type.isInterface() ? result.getClassesImplementing(type) : result.getSubclasses(type);
-          implementingClasses = children.getStandardClasses()
-                                    .filter(cls -> cls.isPublic() && !cls.isAbstract())
-                                    .loadClasses();
+          implementingClasses =
+              children.getStandardClasses()
+                  .filter(info -> isAccessible(info.getPackageName(), info.getModifiers()))
```
bertschneider, line 480: `isAccessible` does not check `isAbstract` but `children` could still include `abstract` classes, couldn't it?

examples/junit/src/test/java/com/example/AutofuzzFuzzTests.java
```diff
@@ -0,0 +1,34 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.example;
+
+import com.code_intelligence.jazzer.api.FuzzerSecurityIssueLow;
```
bertschneider, line 17: Unused

examples/junit/src/test/java/com/example/AutofuzzFuzzTests.java
```diff
@@ -0,0 +1,34 @@
...
+import com.code_intelligence.jazzer.junit.FuzzTest;
+
+class AutofuzzFuzzTests {
```
bertschneider, line 20: I tried to add an inner class, to check the package visibility checks, and use that as test parameter, but directly get an error that the parameter is null. Could you please add a similar test case?

bertschneider (commented):
`@FuzzTest` gets more and more useful!

bertschneider:
>  I am planning to refactor the static state out of `Meta` & `Utils`

Yes, please :smile:

driver/src/main/java/com/code_intelligence/jazzer/autofuzz/Meta.java
```diff
@@ -445,27 +454,31 @@ static Object consume(FuzzedDataProvider data, Type genericType, AutofuzzCodegen
...
       }
-      return data.pickValue(sortExecutables(YourAverageJavaClass.class.getConstructors()));
+      return data.pickValue(getAccessibleMethods(YourAverageJavaClass.class));
     } else if (type.isInterface() || Modifier.isAbstract(type.getModifiers())) {
       List<Class<?>> implementingClasses = implementingClassesCache.get(type);
       if (implementingClasses == null) {
-        ClassGraph classGraph =
-            new ClassGraph().enableClassInfo().enableInterClassDependencies().rejectPackages(
-                "jaz.*");
+        ClassGraph classGraph = new ClassGraph()
```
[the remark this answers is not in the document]
fmeum: That's true. Due to the cache, we will probably not scan that often, but it would still be useful to only scan once. I don't know whether there is a good way to wrap `ScanResult` in something like a `SoftReference` though - I will need to read up on this first.

driver/src/main/java/com/code_intelligence/jazzer/autofuzz/Meta.java
fmeum: True, added another `filter` call.

examples/junit/src/test/java/com/example/AutofuzzFuzzTests.java
fmeum: Could you describe the failure in a bit more detail? I added a private static class and verified and the test passes.

bertschneider: In an autofuzz fuzzing run the method parameter has a good chance of being `null`. Like in this example:

```
  static class InnerClass {
    private String str;

    InnerClass(String str) {
      this.str = str;
    }

    void validate() {
      if ("jazzer".equals(str)) {
        throw new RuntimeException();
      }
    }
  }

  @FuzzTest
  void innerClass(InnerClass innerClassParam) {
    // Ignore assumptions exceptions during fuzzing run?
    // assumeTrue(innerClassParam != null);
    // if (innerClassParam == null) {
    //   return;
    // }
    innerClassParam.validate();
  }
```

Of course null-checking the parameter works, but is that what users expect? On the other hand, providing `null`, from time to time, is also valuable.

I noticed that exceptions raised in assumptions let a fuzzing run fail, should we ignore those while fuzzing?

bertschneider (approved):

examples/junit/src/test/java/com/example/AutofuzzFuzzTests.java
fmeum: I pushed a commit that ignores `TestAbortedException`. I really like `assumeTrue(innerClassParam != null)`. Given the chaos surrounding `@NonNull`-like annotations and the fact that it may be impossible for the fuzzer to generate a non-`null` value given its current input, I think that `assumeX` models this quite well and should be our preferred way to handle this situation.

PR #477, opened 2022-09-08
Generate docs and source JAR for jazzer-junit

### LINKED ISSUE #474: JUnit source JAR on Maven Central is invalid
The new JUnit integration is awesome, but it looks like the source JAR uploaded to Maven Central (jazzer-junit-0.12.0-sources.jar) is invalid. My IDE (Eclipse) is unable to use it to browse the sources, and 7Zip is unable to open it (which is usually possible).

fmeum:
The automatic docs and source JAR generation provided by `rules_jvm_external`'s `java_export` couldn't handle our shading and stripping.

Fixes #474

### DIFF
--- a/driver/src/main/java/com/code_intelligence/jazzer/junit/BUILD.bazel
+++ b/driver/src/main/java/com/code_intelligence/jazzer/junit/BUILD.bazel
@@ -1,10 +1,25 @@
 load("@com_github_johnynek_bazel_jar_jar//:jar_jar.bzl", "jar_jar")
 load("@fmeum_rules_jni//jni:defs.bzl", "java_jni_library")
+load("@rules_jvm_external//:defs.bzl", "javadoc")
 load("//bazel:compat.bzl", "SKIP_ON_WINDOWS")
 load("//bazel:jar.bzl", "strip_jar")
 load("//:maven.bzl", "JAZZER_JUNIT_MAVEN_DEPS")
 load("//sanitizers:sanitizers.bzl", "SANITIZER_CLASSES")
 
+javadoc(
+    name = "junit-docs",
+    visibility = ["//deploy:__pkg__"],
+    deps = [
+        ":agent_configurator",
+        ":fuzz_test",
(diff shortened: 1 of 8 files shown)


fmeum:
@bertschneider I pushed a new commit that I hope fixes the flakiness for good. I observed that the test always passed on retry, so running it by itself should work well.

driver/src/main/java/com/code_intelligence/jazzer/junit/BUILD.bazel
```diff
@@ -1,24 +1,42 @@
...
+# The Jazzer JUnit integration without its dependencies declared in the pom.xml, together with the unshaded sources.
+java_import(
     name = "junit",
+    jars = ["junit_shaded.jar"],
+    srcjar = "junit_unshaded_deploy-src.jar",
+    visibility = ["//visibility:public"],
+)
+
+javadoc(
+    name = "junit-docs",
```
bertschneider, line 18: The `@FuzzTest` Javadoc seems outdated.

bertschneider (commented):
Changing `JAZZER_VALUE_PROFILE` to `True` and `False`, as suggested by the error message, fixes the tests for me.

driver/src/main/java/com/code_intelligence/jazzer/junit/BUILD.bazel
fmeum: Yes, I want to make one more breaking change to the API before I update it.

fmeum:
@bertschneider Do you know whether Javadocs for one Maven dependency can reference classes from another one? Otherwise we should probably include the docs for `jazzer-api` in those for `jazzer-junit`.

bertschneider:
Not that I'm aware of.

bertschneider (approved):

fmeum:
> Not that I'm aware of.

I have inlined the docs for the non-hook part of the API into the `jazzer-junit` docs.

PR #479, opened 2022-09-09
Fix JaCoCo coverage measurements

fmeum:
Fixes https://github.com/google/oss-fuzz/issues/8241


fmeum:
@sebastianpoeplau The last commit is optional, the tests already pass without it. If you have any concerns, I would err on the safe side and drop it.

sebastianpoeplau (approved):
Looks good to me!

sebastianpoeplau:
> @sebastianpoeplau The last commit is optional, the tests already pass without it. If you have any concerns, I would err on the safe side and drop it.

I think it's fine :thinking: I'm more concerned about returning from libfuzzer's main function instead of exiting. Are we sure that that's safe? I seem to remember a comment somewhere in libfuzzer code saying something like "let's make sure the destructor for this object is never called", but maybe that's in a different place.

fmeum:
> > @sebastianpoeplau The last commit is optional, the tests already pass without it. If you have any concerns, I would err on the safe side and drop it.
> 
> I think it's fine thinking I'm more concerned about returning from libfuzzer's main function instead of exiting. Are we sure that that's safe? I seem to remember a comment somewhere in libfuzzer code saying something like "let's make sure the destructor for this object is never called", but maybe that's in a different place.

That place was [here](https://github.com/CodeIntelligenceTesting/llvm-project-jazzer/commit/987fcadbb22a4f7ac90e17af6fa922c66f5eb02b#diff-597daa25b5da15ceaa6dedf9673583d5e7f4a0573227b9ea36771f1fc0e3e4afL919) - it's already patched out on `main` even before this PR. @kyakdan and I briefly looked into it and couldn't find anything problematic - all the data structures powering e.g. the sanitizer callbacks weren't affected. But it would be great if we could confirm this once and for all (maybe next week?) as we could then try to upstream these patches.

sebastianpoeplau:
> > > @sebastianpoeplau The last commit is optional, the tests already pass without it. If you have any concerns, I would err on the safe side and drop it.
> > 
> > 
> > I think it's fine thinking I'm more concerned about returning from libfuzzer's main function instead of exiting. Are we sure that that's safe? I seem to remember a comment somewhere in libfuzzer code saying something like "let's make sure the destructor for this object is never called", but maybe that's in a different place.
> 
> That place was [here](https://github.com/CodeIntelligenceTesting/llvm-project-jazzer/commit/987fcadbb22a4f7ac90e17af6fa922c66f5eb02b#diff-597daa25b5da15ceaa6dedf9673583d5e7f4a0573227b9ea36771f1fc0e3e4afL919) - it's already patched out on `main` even before this PR. @kyakdan and I briefly looked into it and couldn't find anything problematic - all the data structures powering e.g. the sanitizer callbacks didn't seem to be affected. But it would be great if we could confirm this once and for all (maybe next week?) as we could then try to upstream these patches.

Sure, I'll have a look :blush:

PR #480, opened 2022-09-09
sanitizers: Rewrite SqlInjection sanitizer in Java

fmeum:


fmeum:
> Looks good.
> 
> If wonder if `org.h2.jdbc.JdbcSQLSyntaxErrorException` could still occur during fuzzing or if we should remove it from the expected findings list. If the fuzzer is able to produce such an exception the SQL is invalid and a SQL injection finding should be generated, shouldn't in?

It occurs when the reproducer is executed. At some point we could split the `allowed_findings` into one for fuzzing and one for hooks, but our test setup is already pretty complicated.

bertschneider:
Ah, yes, thx for clarification.

PR #485, opened 2022-09-13
Clean up and verify stack traces

fmeum:


bazel/tools/java/com/code_intelligence/jazzer/tools/FuzzTargetTestWrapper.java
```diff
@@ -185,6 +210,65 @@ private static String rlocationPath(String rootpath) {
...
+    List<String> findings =
+        stackTrace.stream()
+            .filter(line -> line.startsWith(EXCEPTION_PREFIX))
+            .map(line -> line.substring(EXCEPTION_PREFIX.length()).split(":", 2)[0])
+            .collect(toList());
+    if (findings.isEmpty()) {
+      throw new IllegalStateException("Expected a crash, but did not get a stack trace");
+    }
+    for (String finding : findings) {
+      if (!expectedFindings.contains(finding)) {
```
bertschneider, line 249: It's not checked if all `expectedFindings` are actually found.

fmeum: This is meant to be a superset of the findings that are allowed rather than the exact set that must be reproduced - different OSes and slightly different OOM semantics make it very difficult to reliably reproduce the exact same set of findings every time.

I would like to clarify the naming and comments though. Do you like `allowedFindings` better?

bertschneider: Yes, that clarifies it. Thx.

fmeum: Pushed that as a separate commit to make it a pure refactoring.

bertschneider (approved):

bazel/tools/java/com/code_intelligence/jazzer/tools/FuzzTargetTestWrapper.java
bertschneider: I like the new name, makes it clear.

PR #488, opened 2022-09-15
Ignore regex pattern syntax exceptions caused by stack overflows

bertschneider:
Compiling regex patterns can cause stack overflows and so the Pattern class handles this case and throws a PatternSyntaxException after unwinding the stack.

We encountered a situation where a library doesn't unwind the stack and uses a regex during handling of a stack overflow. This in turn causes another stack overflow error and the RegexInjection bug detector wrongly reports a finding.

It is very unlikely that the fuzzer triggers a stack overflow while trying to provoke a PatternSyntaxException. Hence, the RegexInjection bug detector now ignores PatternSyntaxExceptions caused by stack overflows.

Apparently stack overflow handling was introduced in JDK 14 and ported back to older versions. The exception message seems to be stable across all versions.

### DIFF
--- a/sanitizers/src/test/java/com/example/BUILD.bazel
+++ b/sanitizers/src/test/java/com/example/BUILD.bazel
@@ -161,6 +161,21 @@ java_fuzz_target_test(
     verify_crash_reproducer = False,
 )
 
+# Catching StackOverflowErrors doesn't work reliably across all systems and JDK versions.
+# It may lead to a native crash before we can handle the exception in Java, therefore the
+# test is set to manual execution.
+java_fuzz_target_test(
+    name = "StackOverflowRegexInjection",
+    srcs = ["StackOverflowRegexInjection.java"],
+    allowed_findings = ["java.util.regex.PatternSyntaxException"],
+    fuzzer_args = [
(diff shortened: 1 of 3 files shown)


fmeum:
The fact that fuzzer output is going into the pipe in CI now is confusing, let me try to improve that situation.

sanitizers/src/test/java/com/example/BUILD.bazel
```diff
@@ -161,6 +161,19 @@ java_fuzz_target_test(
     verify_crash_reproducer = False,
 )
 
+# Due to system dependant behavior of JVMs this test is set to manual execution.
```
fmeum, line 164: Could we expand this comment a bit more? We should at least say that catching `StackOverflowError` doesn't work reliably since it may have lead to a native crash before we can handle it in Java.

fmeum (approved):

sanitizers/src/test/java/com/example/BUILD.bazel
bertschneider: Right, I extended the comment. Is it better now?

PR #490, opened 2022-09-19
Extract launcher

fmeum:


bertschneider (approved):
Is this change transparent for OSS-Fuzz? I don't know how Jazzer is set up there.

fmeum:
It requires changes, but we can make those next time we update the pinned version.

PR #492, opened 2022-09-20
driver: Add a Java launcher

fmeum:


bazel/tools/java/com/code_intelligence/jazzer/tools/FuzzTargetTestWrapper.java
```diff
@@ -98,14 +101,29 @@ public static void main(String[] args) {
...
-        hookJarActualPath == null
-            ? jarActualPath
-            : String.join(System.getProperty("path.separator"), jarActualPath, hookJarActualPath)));
     if (System.getenv("JAZZER_NO_EXPLICIT_SEED") == null) {
       command.add("-seed=2735196724");
     }
+    if (usesJavaLauncher) {
+      System.setProperty("jazzer.reproducer_path", outputDir);
+      // Forward all jazzer.* properties to the JVM used by the Java driver via the java_binary's
+      // --wrapper_script_flag argument.
```
bertschneider, line 110: Could you please explain a little bit more where `--wrapper_script_flag` is coming from and that `driverActualPath` can point to the wrapper script, if `usesJavaLauncher`? Perhaps also a link to the bazel docs.

driver/src/main/java/com/code_intelligence/jazzer/Jazzer.java
```diff
@@ -0,0 +1,89 @@
...
+ * <p>Arguments to libFuzzer are passed as command-line arguments.
+ */
+public class Jazzer {
+  public static void main(String[] args) throws IOException {
+    List<String> libFuzzerArgs =
+        Stream.concat(Stream.of(prepareArgv0()), Arrays.stream(args)).collect(toList());
+    System.exit(Driver.start(libFuzzerArgs));
+  }
+
+  private static final FileAttribute<?>[] LAUNCHER_SCRIPT_ATTRIBUTES = isPosix()
```
bertschneider, line 53: Any particular reason for the fields to be `static`? I would tend to move them into `prepareArgv0`.

bertschneider (approved):
Really nice! No native binaries needed anymore.
Just two minor remarks.

driver/src/main/java/com/code_intelligence/jazzer/Jazzer.java
fmeum: They are true `static final` constants (since `isPosix()` is as well). I could move them into `prepareArgv0`, but wouldn't that somewhat obscure this fact? I will definitely move them to the end of the class.

bazel/tools/java/com/code_intelligence/jazzer/tools/FuzzTargetTestWrapper.java
fmeum: Added, please take a look.

driver/src/main/java/com/code_intelligence/jazzer/Jazzer.java
bertschneider: True, but they are only needed once inside `prepareArgv0` and there's no need to keep them around. But, as mentioned, only a minor difference in taste.

fmeum: I made them local.

PR #494, opened 2022-09-27
BREAKING: junit: Remove seedCorpus parameter on @fuzztest 

fmeum:
It is not clear that the default seed corpus location needs to be
overridden - we should stick to conventions for as long as possible.


fmeum:
@bertschneider The two commits don't build individually, so I will squash them before the merge. They should help with the review though.

driver/src/main/java/com/code_intelligence/jazzer/junit/JazzerFuzzTestExecutor.java
```diff
@@ -78,47 +78,37 @@ public TestExecutionResult execute() throws IOException, URISyntaxException {
...
+                         -> libFuzzerArgs.add(
+                             String.format("-artifact_prefix=%s%c", path, File.separatorChar)));
     } else {
       // We can't directly use the seed corpus from resources as it's packaged into a JAR. Instead,
       // try to get the path to the seed corpus in the source tree.
-      Optional<Path> seedCorpusSourceDirectory = findSeedCorpusDirectoryInSourceTree();
-      if (seedCorpusSourceDirectory.isPresent()) {
-        libFuzzerArgs.add(seedCorpusSourceDirectory.get().toString());
+      Optional<Path> seedCorpus = seedCorpusSourcePath(fuzzTestClass, baseDir);
+      if (seedCorpus.isPresent()) {
```
bertschneider, line 106: Should we use the same pattern to check if `seedCorpus` is present as in the other if-branch? So either `seedCorpus.ifPresent()` or `if(seedCorpus.isPresent())`.

bertschneider (approved):

driver/src/main/java/com/code_intelligence/jazzer/junit/JazzerFuzzTestExecutor.java
fmeum: That case would require `ifPresentOrElse`, which unfortunately isn't available in Java 8. :-(

PR #498, opened 2022-09-27
Process arguments in Java

fmeum:


fmeum:
@bertschneider I haven't added any actual help texts as I would like to get your feedback on the general approach first.

driver/src/main/java/com/code_intelligence/jazzer/driver/Opt.java
```diff
@@ -41,51 +47,75 @@
...
-  public static final String coverageReport = stringSetting("coverage_report", "");
-  public static final List<String> customHookIncludes = stringListSetting("custom_hook_includes");
-  public static final List<String> customHookExcludes = stringListSetting("custom_hook_excludes");
-  public static final List<String> customHooks = stringListSetting("custom_hooks");
-  public static final List<String> disabledHooks = stringListSetting("disabled_hooks");
-  public static final String dumpClassesDir = stringSetting("dump_classes_dir", "");
-  public static final boolean hooks = boolSetting("hooks", true);
-  public static final String idSyncFile = stringSetting("id_sync_file", null);
+  // All supported arguments are added to this set by the individual *Setting methods.
+  private static final Map<String, OptDetails> knownArgs = new TreeMap<>();
```
bertschneider, line 51: Hide `OptDetails` and the different type methods in a dedicated class?

bertschneider:
I really like the approach to handle arguments in Java and get rid of `gflags`. The parsing code is still quite simple but we need to watch out that it stays this way :smile:

driver/src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetRunner.java
```diff
@@ -251,7 +251,7 @@ private static int runOne(long dataPtr, int dataLength) {
       dumpReproducer(data);
     }
 
-    if (Long.compareUnsigned(ignoredTokens.size(), Opt.keepGoing) >= 0) {
+    if (!Opt.dedup || Long.compareUnsigned(ignoredTokens.size(), Opt.keepGoing) >= 0) {
```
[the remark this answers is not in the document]
fmeum: I explained the situation in which this could occur in more detail in the commit message. It required combining two undocumented features and, with the upcoming changes that found this, was caught by an existing test.

driver/src/main/java/com/code_intelligence/jazzer/driver/Opt.java
fmeum: Done, I refactored out all the parsing logic.

bertschneider (approved):

PR #506, opened 2022-10-10
Clean up FuzzTargetTestWrapper

fmeum:


bazel/tools/java/com/code_intelligence/jazzer/tools/FuzzTargetTestWrapper.java
```diff
@@ -172,8 +175,7 @@ public static void main(String[] args) {
 
     if (JAZZER_CI && verifyCrashReproducer) {
       try {
-        verifyCrashReproducer(
-            outputDir, driverActualPath, apiActualPath, jarActualPath, allowedFindings);
+        verifyCrashReproducer(outputDir, apiActualPath, targetJarActualPath, allowedFindings);
```
kyakdan, line 178: Both the boolean variable and method have the same names. It would be more readable to rename boolean variable to something like `shouldVerifyCrashReproducer`

kyakdan: Should we also make `outputDir` a `Path` instead of `String`?

fmeum: Good point, both done!

kyakdan (approved):

PR #507, opened 2022-10-10
Prepare release of the stand-alone Jazzer JAR

fmeum:


agent/src/main/java/com/code_intelligence/jazzer/runtime/BUILD.bazel
```diff
@@ -16,6 +16,18 @@ java_binary(
...
 )
 
+# These classes with public Bazel visibility are contained in jazzer_bootstrap.jar
+# and will thus be available on the bootstrap class path.
+java_binary(
+    name = "jazzer_bootstrap_env",
+    create_executable = False,
+    visibility = ["//driver/src/main/java/com/code_intelligence/jazzer:__pkg__"],
+    runtime_deps = [
+        "//agent/src/main/java/com/code_intelligence/jazzer/api:hooks",
```
kyakdan, line 32: This would mean that these classes will be included in both `jazzer.jar` and `jazzer_bootstrap.jar`. Does this mean that duplicate classes might be loaded?

bazel/tools/java/com/code_intelligence/jazzer/tools/JarStripper.java
```diff
@@ -78,35 +94,55 @@ public static void main(String[] args) {
...
-            throw new RuntimeException(e);
-          }
-        });
+        walk.sorted(Comparator.reverseOrder())
+            .filter(path
+                -> (pathsToKeep != null && !pathsToKeep.matches(path))
+                    || (pathsToDelete != null && pathsToDelete.matches(path)))
+            .forEach(path -> {
+              try {
+                Files.delete(path);
```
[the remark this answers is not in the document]
fmeum: That would be possible, but I don't think we could use this convenient interface then. The aim of this tool is mostly to be simple and deterministic. If performance were a concern (our jars are <20MB), we would be better off using something like Bazel's ijar as a base.

bertschneider (approved):

agent/src/main/java/com/code_intelligence/jazzer/runtime/BUILD.bazel
fmeum: I added a comment to explain the purpose of this target better: It is used as a `deploy_env` of the actual Jazzer binary precisely to prevent this issue ;-)

kyakdan: Great, thanks for clarifying this!

PR #508, opened 2022-10-11
Test and improve durationStringToSeconds

fmeum:


driver/src/main/java/com/code_intelligence/jazzer/junit/JazzerFuzzTestExecutor.java
```diff
@@ -158,10 +158,9 @@ public TestExecutionResult execute() throws IOException, URISyntaxException {
     }
   }
 
-  private static long durationStringToSeconds(String duration) {
+  static long durationStringToSeconds(String duration) {
     // Convert the string to ISO 8601 (https://en.wikipedia.org/wiki/ISO_8601#Durations). We do not
-    // allow for duration
-    // units longer than hours, so we can always prepend PT.
+    // allow for duration units longer than hours, so we can always prepend PT.
     String isoDuration = "PT" + duration.replace("min", "m").replace(" ", "");
```
kyakdan, line 164: Why do minutes receive special handling?

fmeum: Good point. Instead of specifying supported units, I just wanted to support all the common abbreviation. I added `hr` and `sec`, which means we are now compatible with [NPM's parse-duration](https://www.npmjs.com/package/parse-duration) as well as Go duration strings.

kyakdan: NPM's parse-duration also handles plural as well, e.g., `'1 hr 20 mins'`, but I don't think we need to support this as long as  we don't say that we are compatible.

kyakdan (approved):

PR #515, opened 2022-10-19
Improve jar stripping to fix various packaging issues

fmeum:


deploy/BUILD.bazel
```diff
@@ -34,6 +38,40 @@ alias(
     actual = "//driver/src/main/java/com/code_intelligence/jazzer/junit:junit.publish",
 )
 
+sh_test(
+    name = "jazzer-api_artifact_test",
+    srcs = ["jazzer-api_artifact_test.sh"],
```
bertschneider, line 43: I wonder if we should move the `junit` artifact and `junit_artifact_test` also into `deploy` to keep exports more closely together.

driver/src/main/java/com/code_intelligence/jazzer/junit/RegressionTestArgumentProvider.java
```diff
@@ -43,6 +45,15 @@
 class RegressionTestArgumentProvider implements ArgumentsProvider, AnnotationConsumer<FuzzTest> {
   private static final String INCORRECT_PARAMETERS_MESSAGE =
       "Methods annotated with @FuzzTest must take at least one parameter";
+  private static final AtomicBoolean agentInstalled = new AtomicBoolean(false);
+
+  private static void configureAndInstallAgent(ExtensionContext extensionContext) {
+    if (agentInstalled.compareAndSet(false, true)) {
```
bertschneider, line 51: Is this check necessary, since both methods are idempotent / check their installation state themselves?

bertschneider (approved):
Great PR :+1:  

Special thanks for the detailed commit messages, those made the review much easier.

deploy/BUILD.bazel
fmeum: Good idea, done. Got quite some code savings that way.

driver/src/main/java/com/code_intelligence/jazzer/junit/RegressionTestArgumentProvider.java
fmeum: It's not necessary now, but I'm not sure I won't break idempotency at some point without noticing - there is just too much global state in these functions. Not rerunning them for each test also seems less wasteful than the explicit check.

PR #531, opened 2022-11-21
Add script engine injection sanitizer with real life example

gdemarcsek:
This merge request adds a sanitiser to detect data flows of user input into the `javax.script.ScriptEngine#eval` sink which often results in arbitrary code execution (CWE-94) using a registered scripting engine.

JVM script engines (JSR-223 implementations) can be registered under multiple arbitrary names and they can implement any language (although ECMAScript is probably the most widely used one in real applications).

This sanitizer implementation only confirms data flow, it does not require successful script execution with a registered script engine (similarly to how the `OsCommandInjection` sanitizer does not guide the fuzzer into running the actual command). As soon as an input produces an execution that reaches the evaluation of the “1+1” expression by a `javax.script.ScriptEngine` implementation, we report a finding.

I have added a test that demonstrates the sanitizer by rediscovering the CVE-2022-42889 in Apache Commons Text using minimal assumptions of the library and the finding. (It does however limit the length of the input taken from the fuzzer and it adds a hook to avoid the `${const:…}` interpolation which can be used to read public static members of any loadable class - not remotely as promising as the `${script:…}` payload.)

```
bazel clean && bazel run --local_cpu_resources=8 //examples:CommonsTextFuzzer
...
INFO: Build completed successfully, 443 total actions
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
Executing tests from //examples:CommonsTextFuzzer
-----------------------------------------------------------------------------
INFO: Loaded 125 hooks from com.code_intelligence.jazzer.runtime.TraceCmpHooks
INFO: Loaded 4 hooks from com.code_intelligence.jazzer.runtime.TraceDivHooks
INFO: Loaded 2 hooks from com.code_intelligence.jazzer.runtime.TraceIndirHooks
INFO: Loaded 4 hooks from com.code_intelligence.jazzer.runtime.NativeLibHooks
INFO: Loaded 1 hooks from com.example.CommonsTextFuzzerHooks
INFO: Loaded 8 hooks from com.code_intelligence.jazzer.sanitizers.Deserialization
INFO: Loaded 5 hooks from com.code_intelligence.jazzer.sanitizers.ExpressionLanguageInjection
INFO: Loaded 70 hooks from com.code_intelligence.jazzer.sanitizers.LdapInjection
INFO: Loaded 46 hooks from com.code_intelligence.jazzer.sanitizers.NamingContextLookup
INFO: Loaded 1 hooks from com.code_intelligence.jazzer.sanitizers.OsCommandInjection
INFO: Loaded 52 hooks from com.code_intelligence.jazzer.sanitizers.ReflectiveCall
INFO: Loaded 8 hooks from com.code_intelligence.jazzer.sanitizers.RegexInjection
INFO: Loaded 16 hooks from com.code_intelligence.jazzer.sanitizers.RegexRoadblocks
INFO: Loaded 19 hooks from com.code_intelligence.jazzer.sanitizers.SqlInjection
INFO: Loaded 5 hooks from com.code_intelligence.jazzer.sanitizers.ScriptEngineInjection
INFO: Instrumented com.example.CommonsTextFuzzer (took 17 ms, size +10%)
INFO: libFuzzer ignores flags that start with '--'
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 3229358036
INFO: Loaded 1 modules   (512 inline 8-bit counters): 512 [0x160278000, 0x160278200),
INFO: Loaded 1 PC tables (512 PCs): 512 [0x15a660e00,0x15a662e00),
INFO: -fork=8: fuzzing in separate process(s)
INFO: -fork=8: 0 seed inputs, starting to fuzz in /var/folders/j7/bry6w71d1jl97rd7s3cy3yrm0000gn/T//libFuzzerTemp.FuzzWithFork68681.dir
#24988: cov: 0 ft: 0 corp: 0 exec/s 12494 oom/timeout/crash: 0/0/0 time: 7s job: 1 dft_time: 0
#73551: cov: 278 ft: 843 corp: 109 exec/s 16187 oom/timeout/crash: 0/0/0 time: 11s job: 2 dft_time: 0
#177135: cov: 308 ft: 1068 corp: 241 exec/s 25896 oom/timeout/crash: 0/0/0 time: 13s job: 3 dft_time: 0
#295530: cov: 338 ft: 1314 corp: 391 exec/s 23679 oom/timeout/crash: 0/0/0 time: 16s job: 4 dft_time: 0
#464996: cov: 341 ft: 1365 corp: 436 exec/s 28244 oom/timeout/crash: 0/0/0 time: 19s job: 5 dft_time: 0
#668737: cov: 415 ft: 1617 corp: 547 exec/s 29105 oom/timeout/crash: 0/0/0 time: 22s job: 6 dft_time: 0
#895854: cov: 637 ft: 2369 corp: 625 exec/s 28389 oom/timeout/crash: 0/0/0 time: 26s job: 7 dft_time: 0
...
#453559	REDUCE cov: 745 ft: 4257 corp: 1474/25Kb lim: 661 exec/s: 5335 rss: 924Mb L: 21/68 MS: 1 EraseBytes-
Warning: Nashorn engine is planned to be removed from a future JDK release
Warning: Nashorn engine is planned to be removed from a future JDK release
#455106	REDUCE cov: 745 ft: 4257 corp: 1474/25Kb lim: 670 exec/s: 5354 rss: 924Mb L: 12/68 MS: 2 CopyPart-EraseBytes-
#455172	REDUCE cov: 745 ft: 4257 corp: 1474/25Kb lim: 670 exec/s: 5354 rss: 924Mb L: 13/68 MS: 1 EraseBytes-
Warning: Nashorn engine is planned to be removed from a future JDK release
Warning: Nashorn engine is planned to be removed from a future JDK release
#456349	REDUCE cov: 745 ft: 4257 corp: 1474/25Kb lim: 679 exec/s: 5368 rss: 924Mb L: 14/68 MS: 2 CopyPart-EraseBytes-
#456545	REDUCE cov: 745 ft: 4257 corp: 1474/25Kb lim: 679 exec/s: 5371 rss: 924Mb L: 12/68 MS: 1 EraseBytes-
Warning: Nashorn engine is planned to be removed from a future JDK release
Warning: Nashorn engine is planned to be removed from a future JDK release

== Java Exception: com.code_intelligence.jazzer.api.FuzzerSecurityIssueCritical: Possible script execution
	at com.code_intelligence.jazzer.sanitizers.ScriptEngineInjection.checkScriptEngineExecute(ScriptEngineInjection.java:114)
	at org.apache.commons.text.lookup.ScriptStringLookup.lookup(ScriptStringLookup.java:86)
	at org.apache.commons.text.lookup.InterpolatorStringLookup.lookup(InterpolatorStringLookup.java:135)
	at org.apache.commons.text.StringSubstitutor.resolveVariable(StringSubstitutor.java:1067)
	at org.apache.commons.text.StringSubstitutor.substitute(StringSubstitutor.java:1433)
	at org.apache.commons.text.StringSubstitutor.substitute(StringSubstitutor.java:1308)
	at org.apache.commons.text.StringSubstitutor.replace(StringSubstitutor.java:816)
	at com.example.CommonsTextFuzzer.fuzzerTestOneInput(CommonsTextFuzzer.java:24)
DEDUP_TOKEN: f10483b167fba3d6
== libFuzzer crashing input ==
MS: 2 ChangeASCIIInt-CMP- DE: "1+1"-; base unit: 6931eff51adba2b80c902e62511ee20b7fe8b3ae
0x24,0x7b,0x73,0x63,0x72,0x69,0x70,0x74,0x3a,0x6a,0x73,0x3a,0x31,0x2b,0x31,0x7d,0x4,0xe7,
${script:js:1+1}\004\347
artifact_prefix='/private/var/tmp/_bazel_gyorgydemarcsek/071a3ae2fb95e15eb05d82d6a2a2dd28/execroot/jazzer/bazel-out/darwin_arm64-opt/testlogs/examples/CommonsTextFuzzer/test.outputs/'; Test unit written to /private/var/tmp/_bazel_gyorgydemarcsek/071a3ae2fb95e15eb05d82d6a2a2dd28/execroot/jazzer/bazel-out/darwin_arm64-opt/testlogs/examples/CommonsTextFuzzer/test.outputs/crash-7b8a90d41dff830e912476dbe9304d2b56e9235e
Base64: JHtzY3JpcHQ6anM6MSsxfQTn
stat::number_of_executed_units: 456787
stat::average_exec_per_sec:     5373
stat::new_units_added:          3165
stat::slowest_unit_time_sec:    0
stat::peak_rss_mb:              924
Warning: Nashorn engine is planned to be removed from a future JDK release
reproducer_path='/private/var/tmp/_bazel_gyorgydemarcsek/071a3ae2fb95e15eb05d82d6a2a2dd28/execroot/jazzer/bazel-out/darwin_arm64-opt/testlogs/examples/CommonsTextFuzzer/test.outputs'; Java reproducer written to /private/var/tmp/_bazel_gyorgydemarcsek/071a3ae2fb95e15eb05d82d6a2a2dd28/execroot/jazzer/bazel-out/darwin_arm64-opt/testlogs/examples/CommonsTextFuzzer/test.outputs/Crash_7b8a90d41dff830e912476dbe9304d2b56e9235e.java
INFO: exiting: 19712 time: 2686s
```

I have managed to reproduce this a couple of times in reasonable running times on a laptop.


examples/BUILD.bazel
```diff
@@ -177,6 +177,36 @@ java_fuzz_target_test(
...
+# WARNING: This fuzz target uses a vulnerable version of Apache Commons Text, which could result in the execution
+# of arbitrary code during fuzzing if executed with an older JDK. Use at your own risk.
+java_fuzz_target_test(
+    name = "CommonsTextFuzzer",
+    srcs = [
+        "src/main/java/com/example/CommonsTextFuzzer.java",
+    ],
+    fuzzer_args = [
+        "-fork=8",
+        "-use_value_profile=1",
```
fmeum, line 199: Did you happen to check whether these arguments, especially `-use_value_profile=1`, actually shorten the time until the bug is found on average?

examples/BUILD.bazel
```diff
@@ -177,6 +177,36 @@ java_fuzz_target_test(
...
+    srcs = ["src/main/java/com/example/CommonsTextFuzzerHooks.java"],
+    create_executable = False,
+    deploy_manifest_lines = ["Jazzer-Hook-Classes: com.example.CommonsTextFuzzerHooks"],
+    deps = [
+        "//agent/src/main/java/com/code_intelligence/jazzer/api:hooks",
+    ],
+)
+
+# WARNING: This fuzz target uses a vulnerable version of Apache Commons Text, which could result in the execution
+# of arbitrary code during fuzzing if executed with an older JDK. Use at your own risk.
```
fmeum, line 191: We added the "if executed with an older JDK" bit to the log4j fuzzer because of [the information in this blog post](https://www.lunasec.io/docs/blog/log4j-zero-day/#who-is-impacted). Does something similar hold for the Apache Commons Text issue?

examples/src/main/java/com/example/CommonsTextFuzzer.java
```diff
@@ -0,0 +1,28 @@
...
+package com.example;
+
+import com.code_intelligence.jazzer.api.FuzzedDataProvider;
+import org.apache.commons.text.StringSubstitutor;
+import com.code_intelligence.jazzer.api.Jazzer;
+
+public class CommonsTextFuzzer {
+    public static void fuzzerTestOneInput(FuzzedDataProvider data) {
+        try {
+            StringSubstitutor.createInterpolator().replace(data.consumeAsciiString(20));
```
fmeum, line 24: `data.consumeRemainingAsAsciiString()` would be slightly more idiomatic. Do you know whether that makes the bug harder to find?

examples/src/main/java/com/example/CommonsTextFuzzerHooks.java
```diff
@@ -0,0 +1,28 @@
...
+package com.example;
+
+import com.code_intelligence.jazzer.api.HookType;
+import com.code_intelligence.jazzer.api.MethodHook;
+import java.lang.invoke.MethodHandle;
+
+public class CommonsTextFuzzerHooks {
+    @MethodHook(type = HookType.REPLACE, targetClassName = "org.apache.commons.text.lookup.StringLookupFactory", targetMethod = "constantStringLookup")
+    public static Object hookConstStringLookup(MethodHandle handle, Object thisObject, Object[] args, int hookId) {
+        // Remove the ConstantStringLookup singleton to avoid the ReflectiveCall
```
fmeum, line 24: Could you explain in a bit more detail why this is needed and why this additional hook makes it work? Could `--disabled_hooks`, which can be used to disable a hook by class name, be added to `fuzzer_args` instead?

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ScriptEngineInjection.java
```diff
@@ -0,0 +1,112 @@
...
+ * Detects Script Engine injection
+ *
+ * <p>
+ * The hooks in this class attempt to detect user input flowing into
+ * {@link javax.script.ScriptEngine.eval} that might lead to
+ * remote code executions depending on the scripting engine's capabilities.
+ * Before JDK 15, the Nashorn Engine
+ * was registered by default with ScriptEngineManager under several aliases,
+ * including "js". Nashorn allows
+ * access to JVM classes, for exmaple {@link java.lang.Runtime} allowing the
```
fmeum, line 45:
```suggestion
 * access to JVM classes, for example {@link java.lang.Runtime} allowing the
```

You can also reflow the comment by turning it into a single line and then running the format script.

fmeum (commented):
Thanks for the great contribution, we certainly don't want to miss detection capabilities for this type of issue.

I will read up on scripting engines and comment on the actual sanitizer a bit more. For the meantime, I left some basic review comments.

examples/BUILD.bazel
gdemarcsek: Yeah, in fact it would be more accurate to say that it may be more likely to be exploitable with older JDKs (< 15) because the Oracle Nashorn scripting engine is no longer registered by default starting at 15 (https://openjdk.org/jeps/372). So yeah, maybe it would be best to remove this remark about the JDK version, I will do that, thanks.

examples/BUILD.bazel
gdemarcsek: Good point, will do some measurements. (I assumed that given the library expects some very specific textual input for it to run into the vulnerable path, it might be better to use value profiling but I never checked how it does without it - maybe the instrumentation overhead makes it not worthwhile, I lack the XP to tell, so yeah, happy to make some measurements :) )

examples/src/main/java/com/example/CommonsTextFuzzer.java
gdemarcsek: I did a couple of runs and it seemed that convergence was better with a limited length, but again, I will try and do some measurements if that's interesting.

examples/src/main/java/com/example/CommonsTextFuzzerHooks.java
gdemarcsek: Yeah, I might even add this as a comment to the hook, but the reason is the following.

Without the hooks, the fuzzer would always first discover the following vector:

```
$\373const:jaz.Zer.\322\"}\346}\012\200\177\177\177\177
```

this is a sample inpit the fuzzer finds, but it is enough to be

```
${const:jaz.Zer.e}
```

for it to trigger. This is because the `org.apache.commons.text.lookup.ConstantStringLookup#lookup` method loads a class by name in order to look up a static final member of the class, hence here user input can control class loading and your `ReflectiveCall` sanitizer catches it, the data flow is about this:

https://github.com/apache/commons-text/blob/rel/commons-text-1.9/src/main/java/org/apache/commons/text/lookup/ConstantStringLookup.java#L122 --> 
https://github.com/apache/commons-text/blob/rel/commons-text-1.9/src/main/java/org/apache/commons/text/lookup/ConstantStringLookup.java#L146 --> 
https://github.com/apache/commons-lang/blob/rel/commons-lang-3.11/src/main/java/org/apache/commons/lang3/ClassUtils.java#L1118 --> 
https://github.com/apache/commons-lang/blob/rel/commons-lang-3.11/src/main/java/org/apache/commons/lang3/ClassUtils.java#L1135 --> 
https://github.com/apache/commons-lang/blob/rel/commons-lang-3.11/src/main/java/org/apache/commons/lang3/ClassUtils.java#L1069 --> 
https://github.com/CodeIntelligenceTesting/jazzer/blob/5d91fc6f21490932fca1081b52e3e57e71a3d5fd/sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ReflectiveCall.kt

I wanted to avoid this for the demo / test because I think the script execution sink is a lot more interesting: with this one, the most you can do is read some constant from a class on the classpath if you also get back the substituted string. So to avoid this, I replaced the factory method responsible for creating `ConstantStringLookup` with an implementation that returns `null`. That way I disable the `${const:something}` interpolation so the fuzzer can never run into the above flow.

bertschneider: We changed the behavior of `jaz.Zer` so that it only reports an issue if an instance is created. This should remove the need for the extra hook.

gdemarcsek:
@bertschneider Got a bit drifted away from this, thx for the reminder, will give it a go this weekend. :)

examples/src/main/java/com/example/CommonsTextFuzzer.java
```diff
@@ -0,0 +1,28 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.example;
+
+import com.code_intelligence.jazzer.api.FuzzedDataProvider;
+import org.apache.commons.text.StringSubstitutor;
+import com.code_intelligence.jazzer.api.Jazzer;
```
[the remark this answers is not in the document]
gdemarcsek: `format.sh` seems to be quite non-portable to me, even the one on main (I think it is written to work on Ubuntu maybe)

anyways, I will rather try to address the issues found by the CI job manually to avoid changing other files (although I no longer can see any jobs triggered for this PR, for some reason, they seem to be stuck)

examples/src/main/java/com/example/CommonsTextFuzzerHooks.java
gdemarcsek: I would have tried to do that after a rebase, but after resolving conflicts, I see weird bazel-related errors popping up in unrelated parts. I also tried ot just build the current main with bazelisk both on macOS and an Ubuntu virual machine with no success.

The error on macOS:

```
bazelisk clean && bazelisk build //... && bazelisk test //... && bazelisk run --local_cpu_resources=8 //examples:CommonsTextFuzzer
INFO: Starting clean (this may take a while). Consider using --async if the clean takes more than several minutes.
INFO: Multiplexer process for Javac has closed its output stream
INFO: Repository com_github_jetbrains_kotlin instantiated at:
  /Users/gyorgydemarcsek/projects/jazzer/WORKSPACE.bazel:10:12: in <toplevel>
  /Users/gyorgydemarcsek/projects/jazzer/init.bzl:27:24: in jazzer_init
  /private/var/tmp/_bazel_gyorgydemarcsek/071a3ae2fb95e15eb05d82d6a2a2dd28/external/io_bazel_rules_kotlin/src/main/starlark/core/repositories/initialize.bzl:48:31: in kotlin_repositories
Repository rule kotlin_compiler_repository defined at:
  /private/var/tmp/_bazel_gyorgydemarcsek/071a3ae2fb95e15eb05d82d6a2a2dd28/external/io_bazel_rules_kotlin/src/main/starlark/core/repositories/compiler.bzl:22:45: in <toplevel>
INFO: Repository remotejdk11_macos instantiated at:
  /DEFAULT.WORKSPACE.SUFFIX:109:6: in <toplevel>
  /private/var/tmp/_bazel_gyorgydemarcsek/071a3ae2fb95e15eb05d82d6a2a2dd28/external/bazel_tools/tools/build_defs/repo/utils.bzl:233:18: in maybe
  /private/var/tmp/_bazel_gyorgydemarcsek/071a3ae2fb95e15eb05d82d6a2a2dd28/external/bazel_tools/tools/jdk/remote_java_repository.bzl:48:17: in remote_java_repository
Repository rule http_archive defined at:
  /private/var/tmp/_bazel_gyorgydemarcsek/071a3ae2fb95e15eb05d82d6a2a2dd28/external/bazel_tools/tools/build_defs/repo/http.bzl:372:31: in <toplevel>
INFO: Repository jazzer_libfuzzer instantiated at:
  /Users/gyorgydemarcsek/projects/jazzer/WORKSPACE.bazel:6:20: in <toplevel>
  /Users/gyorgydemarcsek/projects/jazzer/repositories.bzl:177:10: in jazzer_dependencies
  /private/var/tmp/_bazel_gyorgydemarcsek/071a3ae2fb95e15eb05d82d6a2a2dd28/external/bazel_tools/tools/build_defs/repo/utils.bzl:233:18: in maybe
Repository rule http_archive defined at:
  /private/var/tmp/_bazel_gyorgydemarcsek/071a3ae2fb95e15eb05d82d6a2a2dd28/external/bazel_tools/tools/build_defs/repo/http.bzl:372:31: in <toplevel>
ERROR: /private/var/tmp/_bazel_gyorgydemarcsek/071a3ae2fb95e15eb05d82d6a2a2dd28/external/fmeum_rules_jni/jni/tools/libjvm_stub/BUILD.bazel:72:21: While resolving toolchains for target @fmeum_rules_jni//jni/tools/libjvm_stub:current_java_runtime: No matching toolchains found for types @bazel_tools//tools/cpp:toolchain_type.
To debug, rerun with --toolchain_resolution_debug='@bazel_tools//tools/cpp:toolchain_type'
If platforms or toolchains are a new concept for you, we'd encourage reading https://bazel.build/concepts/platforms-intro.
ERROR: Analysis of target '//:jazzer' failed; build aborted:
INFO: Elapsed time: 2.278s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (214 packages loaded, 1808 targets\
 configured)
    Fetching repository @local_config_cc; Building xcode-locator
    Fetching https://github.com/.../v1.7.22/kotlin-compiler-1.7.22.zip
    Fetching repository @jazzer_jacoco; starting
    Fetching ...d82d6a2a2dd28/external/jazzer_jacoco; Extracting v0.8.8.tar.gz
    Fetching https://mirror.bazel.build/....19-ca-jdk11.0.15-macosx_x64.tar.gz
```

Ubuntu 22.04:

```
bazelisk clean && bazelisk build //...
INFO: Starting clean (this may take a while). Consider using --async if the clean takes more than several minutes.
ERROR: /home/gdemarcsek/jazzer/src/main/java/com/code_intelligence/jazzer/runtime/BUILD.bazel:59:8: no such target '@local_jdk//:bin/jar': target 'bin/jar' not declared in package '' defined by /home/gdemarcsek/.cache/bazel/_bazel_gdemarcsek/b7804ce787ab40bb5bfdfe307e852b25/external/local_jdk/BUILD.bazel (Tip: use `query "@local_jdk//:*"` to see all the targets in that package) and referenced by '//src/main/java/com/code_intelligence/jazzer/runtime:jazzer_bootstrap_shading_test'
INFO: Repository remote_coverage_tools instantiated at:
  /DEFAULT.WORKSPACE.SUFFIX:4:6: in <toplevel>
  /home/gdemarcsek/.cache/bazel/_bazel_gdemarcsek/b7804ce787ab40bb5bfdfe307e852b25/external/bazel_tools/tools/build_defs/repo/utils.bzl:233:18: in maybe
Repository rule http_archive defined at:
  /home/gdemarcsek/.cache/bazel/_bazel_gdemarcsek/b7804ce787ab40bb5bfdfe307e852b25/external/bazel_tools/tools/build_defs/repo/http.bzl:372:31: in <toplevel>
INFO: Repository remote_java_tools instantiated at:
  /DEFAULT.WORKSPACE.SUFFIX:363:6: in <toplevel>
  /home/gdemarcsek/.cache/bazel/_bazel_gdemarcsek/b7804ce787ab40bb5bfdfe307e852b25/external/bazel_tools/tools/build_defs/repo/utils.bzl:233:18: in maybe
Repository rule http_archive defined at:
  /home/gdemarcsek/.cache/bazel/_bazel_gdemarcsek/b7804ce787ab40bb5bfdfe307e852b25/external/bazel_tools/tools/build_defs/repo/http.bzl:372:31: in <toplevel>
INFO: Repository com_github_jetbrains_kotlin instantiated at:
  /home/gdemarcsek/jazzer/WORKSPACE.bazel:10:12: in <toplevel>
  /home/gdemarcsek/jazzer/init.bzl:27:24: in jazzer_init
  /home/gdemarcsek/.cache/bazel/_bazel_gdemarcsek/b7804ce787ab40bb5bfdfe307e852b25/external/io_bazel_rules_kotlin/src/main/starlark/core/repositories/initialize.bzl:48:31: in kotlin_repositories
Repository rule kotlin_compiler_repository defined at:
  /home/gdemarcsek/.cache/bazel/_bazel_gdemarcsek/b7804ce787ab40bb5bfdfe307e852b25/external/io_bazel_rules_kotlin/src/main/starlark/core/repositories/compiler.bzl:22:45: in <toplevel>
ERROR: Analysis of target '//src/main/java/com/code_intelligence/jazzer/runtime:jazzer_bootstrap_shading_test' failed; build aborted: Analysis failed
INFO: Elapsed time: 1.010s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (121 packages loaded, 83 targets configured)
    currently loading: @rules_pkg//pkg/private ... (4 packages)
    Fetching repository @rules_python; starting
    Fetching ...rcsek/b7804ce787ab40bb5bfdfe307e852b25/external/rules_python; Extracting 0.16.1.tar.gz
    Fetching https://mirror.bazel.build/.../releases/coverage_output_generator-v2.6.zip
    Fetching https://mirror.bazel.build/bazel_java_tools/releases/java/v11.9/java_tools-v11.9.zip
    Fetching https://github.com/JetBrains/kotlin/releases/download/v1.7.22/kotlin-compiler-1.7.22.zip

```

I have no clue so far what to do with this.

bertschneider: The first one, `No matching toolchains found for types @bazel_tools//tools/cpp:toolchain_type`, indicates that Bazel can not pick up the C/C++ toolchain. Is XCode installed?

The second one is similar, but can not find a local JDK.

bertschneider: Could you take another look at the build issue?

gdemarcsek: Yep, it seems a full XCode installation and switching to XCode from the CLI tools solves this issue. I guess in the meantime having the cli tools became insufficient to build the main branch. for some reason.

examples/src/main/java/com/example/CommonsTextFuzzer.java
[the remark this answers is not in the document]
gdemarcsek: i did run the clang formatter on the new Java files, not sure if it fixes it on CI because workflows are blocked on approval - please reopen if still problematic

examples/BUILD.bazel
gdemarcsek: I made some measurements based on which my assumption was right. I observed that with value profiling on, the overall convergence is better, even though execution speed gradually decreases. Without value profile, exec / sec is sort of consistently high, but I expect the fuzzer explores a lot more uninteresting branches and inputs and overall delivers slower progress.

In my environment, the Commons Text example averaged about 50 minutes but with large deviation, the quickest run I have seen was 15 mins. Meanwhile, without value profiling, the average runtime was about 2 hours and even the quickest I have seen was 75 mins. I would keep the value profiling on.

examples/src/main/java/com/example/CommonsTextFuzzer.java
gdemarcsek: Yeah, this seems to make a big difference in runtime as far as I could see. I am not sure exactly why, I don't know nearly enough about the internals, but my educated guess was that because it has to fuzz `${script:js:(...)}` it is best to minimize the search space (the `js:(...)` part is injected by the hook, but the rest has to be found by the fuzzer and every input that doesn't end with a `}` won't cut it, no matter that it begins correctly.

bertschneider:
I thought about the false positive rate of raising a finding whenever the scripting engine is invoked and ways to reduce that possibility.

As I understand the [documentation](https://docs.oracle.com/javase/10/nashorn/toc.htm) and [other blog posts](https://codeutopia.net/blog/2009/01/02/sandboxing-rhino-in-java/) one would restrict the script engine to only access allowed classes. Using a properly secured script engine would still be fine.

To take that into account we could try to invoke `Jazzer::reportFindingFromHook` and/or block the execution thread (e.g. with an endless loop) to detect denial of service attacks. This would require some code and would not be appropriate for the `guideTowardsContainment` approach anymore.

But it should still be possible to guide the fuzzer to invoke the internal `load` function to execute a provided script, which could trigger the finding. E.g. `load('classpath:script.js')`.

What do you think about this approach?

bertschneider:
I could also give it a try, if you like :smile:

bertschneider:
I don't want to annoy, but any update on this sanitizer? I think it could be a great addition to the project.

gdemarcsek:
Sorry about the delay. Some thoughts (tl;dr I would ignore sandboxing and would avoid relying on any specific script language or engine but some smaller changes might make sense like checking containment instead of equality and randomizing the injected string to avoid detecting constants that happen to be the same):

> I thought about the false positive rate of raising a finding whenever the scripting engine is invoked and ways to reduce that possibility.

If this raises a finding whenever the scripting engine is invoked, then I think I misunderstood the entire project. :( I thought the point of this is to prove that user-supplied input flows into the script engine sink, i.e. no constant string would trigger the detector provided it doesn't happen to be the test script - having said that, it might be better to randomize it, as it can be any string really given we are not checking for actual execution). 

In fact, it's already quite low FP IMHO because it checks for equality, not even containment, so it detects only if you can control the entire script - maybe `guideTowardsContainment` would be more realistic as controlling even some parts can be dangerous, but it's a question of balancing FP and FN. I am happy to make any of these two changes ofc (randomized payload and/or containment instead of equality). 


> As I understand the [documentation](https://docs.oracle.com/javase/10/nashorn/toc.htm) and [other blog posts](https://codeutopia.net/blog/2009/01/02/sandboxing-rhino-in-java/) one would restrict the script engine to only access allowed classes. Using a properly secured script engine would still be fine.

As far as I understand, such sandboxing capabilities would be quite engine-specific and rely on engine-specific APIs (like the Rhino blog post you linked or this guide for Nashorn: https://docs.oracle.com/javase/10/nashorn/nashorn-java-api.htm#JSNUG125). This feature would only detect injections via the Java Scripting API (JSR 223) as it would be quite difficult to write a difficult to every single script engine out there. (Similarly how you already detect SQLi by hooking JDBC classes, not specific SQL driver classes.) On a more principled note, it's impossible to reason about the environment of the application from the vantage point of the detector IMHO. (Sure, there might be a sandbox, but is it really secure? How "far" should I look for it, for example the Java process might run in a container which may be quite restricted. Similarly, should I not report a confirmed command injection if AppArmor/SELinux/Seccomp/Capabilities might be making it difficult (impossible?) to spawn child processes?) You guys know better, but I think the vulnerability is still there - sandboxes mitigate (reduce risk), but do not remediate (remove root cause). I think the only way to avoid script injection while allowing the scripts to access user-input is via bindings maybe, but even then, user-controlled input should not flow to `eval` directly so it might make sense actually to even hook the versions of `eval` that do receive a `ScriptContext` or `Bindings` argument but I didn't want to go that far for a first iteration. 

> To take that into account we could try to invoke Jazzer::reportFindingFromHook and/or block the execution thread (e.g. with an endless loop) to detect denial of service attacks. This would require some code and would not be appropriate for the guideTowardsContainment approach anymore.
> 
> But it should still be possible to guide the fuzzer to invoke the internal load function to execute a provided script, which could trigger the finding. E.g. load('classpath:script.js').
> 
> What do you think about this approach?

I am not sure I fully understand this part, but indeed, my initial approach was to actually execute the script and confirm evaluation, e,g. execute "1 + 1" with the script engine and check that I get 2 or similar POCs (like `load()`). Then I thought the problem is that whatever we write there, it will be specific to the engine and its language which we cannot really anticipate and support universally. I think you might have faced the same dilemma with the command injection detector which also doesn't actually verify command execution, only injection. So yeah, it depends on what you really want to detect - my goal was to detect user-controllable data flow into `javax.script.ScriptEngine.eval` sink and show that it could have found a specific CVE without too much knowledge about the tested library.

bertschneider:
I misphrased that, sorry, and your understanding is correct! I wondered if raising a finding whenever the script engine is invoked _with the fixed string_ would be sufficient or not. 

The argument that we would need script engine specific inputs to trigger findings from within the script engines themselves is quite convincing and I would also try to avoid that. I only thought about JS/Nashorn, as that mainly caused problems in the past, but there are plenty of other script engines out there as well.

I agree that user controlled input should not be part of the script content directly, in that case checking for _containment_ could be even better. It's not very likely but `1+1` could be included in a script content, so we should change that to something Jazzer specific before switching to a containment check. Could you change that part as offered?

If the script content is checked before the actual invocation the hooks an `registerEngineName` and `getEngineByName` are not really necessary anymore, aren't they?

bertschneider:
Friendly ping :smiley:

bertschneider:
@gdemarcsek I took the liberty to rebase your PR an `master` and also incorporate our last discussion. Switching to a `contains` check somewhat complicates the detection. The content of `Reader` parameters can now fully be read, checked if it contains our payload and passed on to the `String` version of `eval`. To do so requires some class loading trickery, though. WDYT of this approach?
I also noticed that there are two overloaded versions of `eval`, which are now hooked as well.

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ScriptEngineInjection.java
```diff
@@ -0,0 +1,146 @@
...
+            "Script Engine Injection: Insecure user input was used in script engine invocation.\n"
+            + "Depending on the script engine's capabilities this could lead to sandbox escape and remote code execution."));
+      }
+      Jazzer.guideTowardsContainment(script, PAYLOAD, hookId);
+    }
+    return script;
+  }
+
+  private static String readAll(Reader reader) {
+    return new BufferedReader(reader).lines().collect(Collectors.joining());
```
fmeum, line 144: This drops line endings. We should probably directly read bytes from the reader instead, possibly limiting the buffet size to something reasonable.

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ScriptEngineInjection.java
```diff
@@ -0,0 +1,146 @@
...
+
+  @MethodHook(type = HookType.REPLACE, targetClassName = "javax.script.ScriptEngine",
+      targetMethod = "eval", targetMethodDescriptor = "(Ljava/lang/String;)Ljava/lang/Object;")
+  @MethodHook(type = HookType.REPLACE, targetClassName = "javax.script.ScriptEngine",
+      targetMethod = "eval", targetMethodDescriptor = "(Ljava/io/Reader;)Ljava/lang/Object;")
+  public static Object
+  checkScriptEngineExecute(MethodHandle method, Object thisObject, Object[] arguments, int hookId)
+      throws Throwable {
+    if (eval == null) {
+      eval = methodFromObject(thisObject, "eval", "java.lang.String");
```
fmeum, line 68: Do we only need this because we have to call the String overload after having read the Reader? If so, it could be simpler to split the hook into one for the String and one for the Reader case. The former is simple, the latter could wrap the String in a `StringReader` and just use the method handle.

fmeum (approved):

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ScriptEngineInjection.java
bertschneider: Right, changed.

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ScriptEngineInjection.java
bertschneider: Ah, why didn't I think of converting the string back to a reader :facepalm:. This makes everything so much simpler. Thanks for the suggestion.

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ScriptEngineInjection.java
```diff
@@ -0,0 +1,107 @@
...
+    return method.invokeWithArguments(thisObject, arguments);
+  }
+
+  private static void checkScriptContent(String content, int hookId) {
+    if (content != null) {
+      if (content.contains(PAYLOAD)) {
+        Jazzer.reportFindingFromHook(new FuzzerSecurityIssueCritical(
+            "Script Engine Injection: Insecure user input was used in script engine invocation.\n"
+            + "Depending on the script engine's capabilities this could lead to sandbox escape and remote code execution."));
+      }
```
fmeum, line 93: Doesn't matter except for readability, but the next line could be wrapped in an `else`.

fmeum (approved):

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ScriptEngineInjection.java
bertschneider: Done.

bertschneider:
@gdemarcsek Do you want to take another look? Otherwise we will merge this PR in a few days. Further changes are always possible, though.

bertschneider:
Merged the PR now. @gdemarcsek, thanks again for your contribution!

PR #533, opened 2022-11-24
Fix sanitizers not being loaded with jazzer_standalone_deploy.jar

fmeum:
I forgot to add the deploy manifest line to `jazzer_standalone_deploy.jar`. Since this is error-prone, use code generation to load the list of built-in hooks instead in a single location.

Also add a check to verify that built-in hooks are loaded. This would have let existing tests catch this regression.

### DIFF
--- a/agent/src/main/java/com/code_intelligence/jazzer/agent/Agent.kt
+++ b/agent/src/main/java/com/code_intelligence/jazzer/agent/Agent.kt
@@ -20,6 +20,7 @@ import com.code_intelligence.jazzer.driver.Opt
 import com.code_intelligence.jazzer.instrumentor.CoverageRecorder
 import com.code_intelligence.jazzer.instrumentor.Hooks
 import com.code_intelligence.jazzer.instrumentor.InstrumentationType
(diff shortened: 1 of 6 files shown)


bertschneider (approved):
Looks good. 

`sanitizers.Constants` is currently located in the main jar file, wheras the actual sanitizers reside in the bootstrap jar. Was that intended?

fmeum:
> `sanitizers.Constants` is currently located in the main jar file, wheras the actual sanitizers reside in the bootstrap jar. Was that intended?

As it is only used by the agent and is loaded before custom hooks have been added to the bootstrap classpath, I think this makes sense, yes. This means both one class less on the bootstrap classpath to worry about and more correct go to definition/source lookup functionality on the jars we ship.

Before this commit, due to the constant being a static final String, it was inlined into usage sites, hence having the class around in the main jar was really the only way that made sense. But given the above, I would still prefer to keep it even after this change. What do you think?

fmeum:
I will try to roll the Jazzer update on OSS-Fuzz forward next week with this fix included.

PR #534, opened 2022-11-24
deploy: Raise Maven version to 0.13.1

fmeum:


fmeum:
I went over the changelog and verified that we only fixed bugs on `main` since the last release.

bertschneider (approved):

PR #548, opened 2022-12-09
Simplify building and testing

fmeum:


fmeum:
I created https://github.com/bazelbuild/bazel/pull/16973 to fix the Bazel bug and added a workaround for the meantime: Windows users will not be able to access the files emitted by our examples unless they figure out that there is a zip file containing them in the same directory. Do you think that's fine for now?

bertschneider (approved):
Looks good!

PR #555, opened 2022-12-15
Preparations for better JUnit support

fmeum:


bertschneider (approved):
LGTM

Apparently provoking the OOM takes too long on some systems. Just add a "bigger chunk of data" in the loop?

fmeum:
> Apparently provoking the OOM takes too long on some systems. Just add a "bigger chunk of data" in the loop?

Tried that, but it leads to crashes in the native instrumentation agent. I am now just allocating a large `long` array, which should fail instantly.

Issue #539, opened 2022-11-28
Junit jupiter not picking up any "jazzer" tagged tests


stephengardiner:
Hi,

I'm trying to retrofit Jazzer to an existing project (often a mistake!), just to get some familiarity with the software. Having seen that the new Junit jupiter integration is in play, I thought I would try and use that.

So, I have a new fuzz test task in my gradle build script:
```
	val fuzzTest = task<Test>("fuzztest") {
		description = "Runs fuzz tests."
		group = "verification"

	dependsOn(tasks.compileTestJava)

	testClassesDirs = sourceSets["test"].output.classesDirs
	classpath = sourceSets["test"].runtimeClasspath + sourceSets.main.get().output
	useJUnitPlatform {
		includeTags("jazzer")
	}
	//shouldRunAfter("test")
}
```

 and I have tagged my fuzz test with the required tag (which is also included in the @FuzzTest annotation) 
```
@Tag("jazzer")
class ByteFuzzTest {
    @FuzzTest(maxDuration = "2m")
    void byteFuzz(byte[] data) {
        assumeFalse(System.getenv("JAZZER_FUZZ").isEmpty());
        if (data.length < 1) {
            return;
        }
        if (data[0] % 2 == 0) {
            fail();
        }
    }
}
```

with JAZZER_FUZZ=1, I always get this error message:
```
org.junit.platform.launcher.core.EngineDiscoveryOrchestrator lambda$logTestDescriptorExclusionReasons$7
INFO: 18 containers and 14 tests were excluded because tags do not match tag expression(s): [jazzer]
```

Probably my configuration or junit jupiter libraries, perhaps? I have these in my build script:
```
    testImplementation(group = "org.junit.jupiter", name = "junit-jupiter", version = "5.9.1")
    testImplementation(group = "org.junit.jupiter", name = "junit-jupiter-api", version = "5.9.1")
    testImplementation(group = "org.junit.jupiter", name = "junit-jupiter-params", version = "5.9.1")
    testImplementation(group = "com.code-intelligence", name="jazzer-junit", version="0.13.1")
```

Pretty sure this is on me, but any suggestions would be gratefully received.

fmeum:
I edited your initial comment to fix the formatting: You have to use three backticks instead of just a single one to properly start and end a code block.

[...]

PR #556, opened 2022-12-16
junit: Implement fuzzing within JUnit Jupiter

### LINKED ISSUE #543: `@FuzzTest` with `JAZZER_FUZZ=1` is not backed by JUnit Jupiter
Firstly, thank you for fixing my formatting of the original issue (seems when you hit insert "Add code" in my browser you only get two backticks inserted - which I assumed you would put your code in-between 🙄 ).

I cannot reopen the original issue as you closed it and I didn't, and you, as collaborator trump me, which is obviously the correct state of affairs.

Also, apologies for not replying sooner, I had a crash on my bicycle the other night and narrowly avoided faceplanting the tarmac (I did hit my head but my helmet saved me from greater injury). Thus, I had a mild concussion and am a bit banged up, so had to curtail my extra curricular jazzer spelunking. 🤕 

Okay, so 0.13.2 gives us some improvement. With JAZZER_FUZZ not set, I get the same output:
```
Dec 02, 2022 7:35:55 PM org.junit.platform.launcher.core.EngineDiscoveryOrchestrator lambda$logTestDescriptorExclusionReasons$7
INFO: 18 containers and 14 tests were excluded because tags do not match tag expression(s): [jazzer]
```
So, it seems it doesn't pay attention to the tags, but obviously only picks up ```"@FuzzTest"``` tagged tests.

It does instrument the project and I suspect, as I haven't setup any resource input, it doesn't do anything for that reason. However right now, I'm more interested in using JAZZER_FUZZ=1 and when I do this I get the same error message above but it does start fuzzing, so thank you!

Something to look at though, especially from my perspective at this stage of jazzer integration and testing, is that your jazzer-specific junit engine doesn't respect the import ```org.junit.jupiter.api.Disabled``` annotation at neither the class level nor the individual test level. This is important because I might have a selection of jazzer tests that I have recorded output for, but then want to create a new jazzer test to stress another part of the api.

Is that possible?

_Originally posted by @stephengardiner in https://github.com/CodeIntelligenceTesting/jazzer/issues/539#issuecomment-1335676027_

fmeum:
Instead of implementing fuzzing runs as a custom JUnit test engine, this commit implements it within the JUnit Jupiter platform instead.

The main benefit of this approach is the full and effortless interoperability with lifecycle hooks such as `@BeforeAll` and `@AfterAll`. The new implementation is also significantly shorter and automatically avoids inconsistencies such as
https://github.com/CodeIntelligenceTesting/jazzer/issues/539 and https://github.com/CodeIntelligenceTesting/jazzer/issues/543.

The new implementation models fuzzing as a single invocation of the `@ParameterizedTest` used for the regression test with fake arguments named `Fuzzing...`. The registered `InvocationInterceptor` is then used to replace the invocation with the actual fuzzing run.
    
In this way, `@FuzzTest` methods invoke `@BeforeEach` hooks once before fuzzing starts. Having the invocation represented as a single invocation parametrized test should make this behavior less surprising.

Fixes #543

Ref: FUZZ-472


src/main/java/com/code_intelligence/jazzer/junit/RegressionTestArgumentProvider.java
```diff
@@ -39,47 +39,70 @@
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Stream;
 import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.ExtensionContext.Namespace;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.ArgumentsProvider;
 import org.junit.jupiter.params.support.AnnotationConsumer;
 
-class RegressionTestArgumentProvider implements ArgumentsProvider, AnnotationConsumer<FuzzTest> {
+class FuzzTestArgumentsProvider implements ArgumentsProvider, AnnotationConsumer<FuzzTest> {
```
bertschneider, line 47: Something went wrong here. The name change and the new file suggest that this one should be removed, right?
Also `RegressionTestExtension` and `JazzerTestEngine`.

src/main/java/com/code_intelligence/jazzer/junit/FuzzTestExtensions.java
```diff
@@ -0,0 +1,97 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.code_intelligence.jazzer.junit;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicReference;
```
bertschneider, line 20: Unused imports.

src/main/java/com/code_intelligence/jazzer/junit/JazzerFuzzTestExecutor.java
```diff
@@ -31,36 +29,33 @@
...
-  private final JazzerFuzzTestDescriptor fuzzTestDescriptor;
-  private final Path baseDir;
+  private final List<String> libFuzzerArgs;
 
-  public JazzerFuzzTestExecutor(
-      ExecutionRequest request, JazzerFuzzTestDescriptor fuzzTestDescriptor, Path baseDir) {
-    this.request = request;
-    this.fuzzTestDescriptor = fuzzTestDescriptor;
-    this.baseDir = baseDir;
+  private JazzerFuzzTestExecutor(List<String> libFuzzerArgs) {
```
bertschneider, line 43: This is the only class in the JUnit integration prefixed with `Jazzer`. Remove that?

bertschneider (changes requested):

src/main/java/com/code_intelligence/jazzer/junit/RegressionTestArgumentProvider.java
bertschneider: I wonder how the build / tests worked in the CI ...

fmeum: This happened when I rebased on the merge of `//driver/...` with `//agent/...`. The files weren't referenced anywhere, which is why CI didn't complain.

src/main/java/com/code_intelligence/jazzer/junit/JazzerFuzzTestExecutor.java
fmeum: Renamed.

src/main/java/com/code_intelligence/jazzer/junit/FuzzTestExtensions.java
fmeum: Removed.

src/main/java/com/code_intelligence/jazzer/junit/RegressionTestArgumentProvider.java
bertschneider: Much nicer now, thx! :smile:

bertschneider (approved):
Looks really good and easier than before.

PR #565, opened 2022-12-21
deploy: Build JNI libraries for Windows arm and arm64

fmeum:


fmeum:
The clang-cl toolchain configured by Bazel doesn't seem to support cross-compilation yet.

PR #567, opened 2022-12-22
Update scripts and docs, and rerun the scripts

kyakdan:
This PR updates the formatting tools and the corresponding comments in the docs.

### DIFF
--- a/.github/workflows/check-formatting.yml
+++ b/.github/workflows/check-formatting.yml
@@ -19,15 +19,15 @@ jobs:
       - name: Setup Go environment
         uses: actions/setup-go@v3
         with:
-          go-version: '^1.19.2'
+          go-version: 'stable'
 
       - name: Install formatters
         run: |
           wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
           sudo apt-get install software-properties-common
-          sudo add-apt-repository 'deb http://apt.llvm.org/focal/ llvm-toolchain-focal-13 main'
-          sudo apt-get install clang-format-13
-          curl -sSLO https://github.com/pinterest/ktlint/releases/download/0.42.1/ktlint && chmod a+x ktlint && sudo mv ktlint /usr/bin/ktlint
+          sudo add-apt-repository 'deb http://apt.llvm.org/focal/ llvm-toolchain-focal-14 main'
+          sudo apt-get install clang-format-14
+          curl -sSLO https://github.com/pinterest/ktlint/releases/download/0.48.0/ktlint && chmod a+x ktlint && sudo mv ktlint /usr/bin/ktlint
           go install github.com/google/addlicense@latest
           go install github.com/bazelbuild/buildtools/buildifier@latest
 
--- a/format.sh
+++ b/format.sh
@@ -1,14 +1,16 @@
+#!/bin/bash
+
 # C++ & Java
-find -name '*.cpp' -o -name '*.c' -o -name '*.h' -o -name '*.java' | xargs clang-format-13 -i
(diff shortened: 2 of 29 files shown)


.github/workflows/check-formatting.yml
```diff
@@ -19,20 +19,21 @@ jobs:
...
-          curl -sSLO https://github.com/pinterest/ktlint/releases/download/0.42.1/ktlint && chmod a+x ktlint && sudo mv ktlint /usr/bin/ktlint
+          sudo add-apt-repository 'deb http://apt.llvm.org/focal/ llvm-toolchain-focal-15 main'
+          sudo apt-get install clang-format-15
+          curl -sSLO https://github.com/pinterest/ktlint/releases/download/0.48.0/ktlint && chmod a+x ktlint && sudo mv ktlint /usr/bin/ktlint
           go install github.com/google/addlicense@latest
           go install github.com/bazelbuild/buildtools/buildifier@latest
 
       - name: Run format.sh and print changes
         run: |
+          shopt -s expand_aliases
```
fmeum, line 36: Didn't know this, could you explain what it does in this context?

.github/workflows/check-formatting.yml
```diff
@@ -19,20 +19,21 @@ jobs:
...
+          go-version: 'stable'
 
       - name: Install formatters
         run: |
           wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
           sudo apt-get install software-properties-common
-          sudo add-apt-repository 'deb http://apt.llvm.org/focal/ llvm-toolchain-focal-13 main'
-          sudo apt-get install clang-format-13
-          curl -sSLO https://github.com/pinterest/ktlint/releases/download/0.42.1/ktlint && chmod a+x ktlint && sudo mv ktlint /usr/bin/ktlint
+          sudo add-apt-repository 'deb http://apt.llvm.org/focal/ llvm-toolchain-focal-15 main'
```
fmeum, line 28: Unless updating to 15 gives us much, could we stick to 14? LLVM 15 isn't available in the official Ubuntu 22.04 sources, which would make it harder to run `./format.sh` locally.

fmeum (commented):
Xcode is back and with it CI flakes, will look into it.

format.sh
```diff
@@ -1,14 +1,16 @@
+#!/bin/bash
+
 # C++ & Java
-find -name '*.cpp' -o -name '*.c' -o -name '*.h' -o -name '*.java' | xargs clang-format-13 -i
+find -name '*.cpp' -o -name '*.c' -o -name '*.h' -o -name '*.java' | xargs clang-format -i
```
fmeum, line 4: I think we should stick to versioned `clang-format` for reproducibility. If the local script picks up `clang-format-11` and CI uses `clang-format-14`, this could become very confusing.

.github/workflows/check-formatting.yml
kyakdan: This was an attempt to make the `clang-format` alias available when executing the format script. I usually install the LLVM toolchain from the GitHub releases, and the tools are not suffixed with a version number. I've adjusted the format script to use `clang-format` and used LLVM 14 in the GitHub action.

.github/workflows/check-formatting.yml
kyakdan: No major gains their that I'm aware of. Changed to LLVM 14

format.sh
kyakdan: It makes sense, but don't we have the same issue with other tools such as `ktlint`?

fmeum: That's a good point. They are easier to install, but I agree we are consistent here. If you think it's easier, then feel free to drop the version suffix here.

fmeum (approved):

PR #574, opened 2022-12-29
Do not throw a security exception in jaz.Zer for disabled sanitizers

kyakdan:


sanitizers/src/test/java/com/example/BUILD.bazel
```diff
@@ -25,6 +25,20 @@ java_fuzz_target_test(
...
 
+java_fuzz_target_test(
+    name = "ReflectiveCallDisabled",
+    srcs = [
+        "ReflectiveCallDisabled.java",
+    ],
+    fuzzer_args = [
+        "-runs=0",
+    ] + select({
+        "@platforms//os:windows": ["--disabled_hooks=com.example.DisabledHook;com.code_intelligence.jazzer.sanitizers.ReflectiveCall"],
```
fmeum, line 36: Since this is a pain and class names never contain `,`, how about using that as a separator? I see relying on `;` or `:` for anything other than `--cp` as a mistake, but back in the day we just didn't think about it as Windows support wasn't on the roadmap at all. For new features, I would like to avoid repeating this mistake.

src/main/java/jaz/Zer.java
```diff
@@ -50,10 +51,33 @@
                Iterator, Runnable, Callable, Function, Collection, List {
   static final long serialVersionUID = 42L;
 
+  private static final Map<String, Set<String>> HOOKS_USING_JAZ_ZER = new HashMap<String,
+      Set<String>>() {
+    {
```
fmeum, line 56: This is silently subclassing `HashMap`, which feels overly intrusive. Since we are iterating over the map anyway, we could just make this a list of map entries or, alternatively, have a small function that initializes the map with statements rather than a single expression.

src/main/java/jaz/Zer.java
```diff
@@ -64,6 +88,25 @@ public Zer(String arg1) {}
 
   public Zer(String arg1, Throwable arg2) {}
 
+  private static boolean isJazZarDisabled() {
+    String allDisabledHooks = System.getProperty("jazzer.disabled_hooks");
```
fmeum, line 92: Let's extract the result of the property lookup into a static final constant so that this can benefit from dead code eliminiation.

src/main/java/jaz/Zer.java
```diff
@@ -50,10 +51,33 @@
...
+          new HashSet<>(Arrays.asList("java.io.ObjectInputStream.readObject",
+              "java.io.ObjectInputStream.readObjectOverride",
+              "java.io.ObjectInputStream.readUnshared")));
+
+      put("com.code_intelligence.jazzer.sanitizers.ExpressionLanguageInjection",
+          new HashSet<>(Arrays.asList("javax.el.ExpressionFactory.createValueExpression",
+              "javax.el.ExpressionFactory.createMethodExpression",
+              "jakarta.el.ExpressionFactory.createValueExpression",
+              "jakarta.el.ExpressionFactory.createMethodExpression",
+              "javax.validation.ConstraintValidatorContext.buildConstraintViolationWithTemplate")));
```
fmeum, line 71: Given the extent of these manual lists, I think that we should add test coverage for all the three sanitizers. Maybe it's possible to reuse the existing Java test code and just add new Bazel test targets with the appropriate changes (set the property and modify `allowed_findings`)?

kyakdan:
@fmeum This PR now implements a new approach to check which sanitizer used `jaz.Zer`  when running in the regression mode. The new approach does not rely on stack traces since the methods that guide the fuzzer do not necessarily end up in the stack trace once the honeypot class is actually executed.

launcher/jvm_tooling.h
```diff
@@ -25,6 +25,15 @@ extern std::string FLAGS_jvm_args;
 extern std::string FLAGS_additional_jvm_args;
 extern std::string FLAGS_agent_path;
 
+#define ARG_SEPARATOR ","
```
fmeum, line 28: Isn't this a breaking change? Do we have to update all OSS-Fuzz build scripts? 

I would prefer to not make any sweeping changes like this to the native launcher as we are only keeping it around for backwards compatibility anyway.

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ExpressionLanguageInjection.kt
```diff
@@ -31,7 +31,7 @@ object ExpressionLanguageInjection {
      * Try to call the default constructor of the honeypot class.
      */
     private const val EXPRESSION_LANGUAGE_ATTACK =
-        "\${\"\".getClass().forName(\"$HONEYPOT_CLASS_NAME\").newInstance()}"
+        "\${\"\".getClass().forName(\"$HONEYPOT_CLASS_NAME\").getMethod(\"el\").invoke(null)}"
```
fmeum, line 34: Why can't we keep `newInstance` here? It is shorter and should also exercise non-trivial functionality (constructor call).

sanitizers/src/test/java/com/example/ClassLoaderLoadClass.java
```diff
@@ -22,9 +22,11 @@ public static void fuzzerTestOneInput(FuzzedDataProvider data) throws Interrupte
     String input = data.consumeRemainingAsAsciiString();
     try {
       // create an instance to trigger class initialization
-      ClassLoaderLoadClass.class.getClassLoader().loadClass(input).getConstructor().newInstance();
-    } catch (ClassNotFoundException | InvocationTargetException | InstantiationException
-        | IllegalAccessException | NoSuchMethodException ignored) {
+      ClassLoaderLoadClass.class.getClassLoader().loadClass(input).newInstance();
+      // TODO(khaled): this fails to reproduce the finding. It seems that this is related not
```
fmeum, line 26: I don't quite understand this comment. If we don't catch all ways a constructor may end up being called, isn't that a sign that something is off with the new (old) approach?

sanitizers/src/test/java/com/example/el/UserData.java
```diff
@@ -0,0 +1,41 @@
...
+}
+
+@Constraint(validatedBy = InsecureEmailValidator.class)
+@Target({ElementType.METHOD, ElementType.FIELD})
+@Retention(RetentionPolicy.RUNTIME)
+@interface ValidEmailConstraint {
+  String message() default "Invalid email address";
+  Class<?>[] groups() default {};
+  Class<? extends Payload>[] payload() default {};
+}
```
fmeum, line 41: Add trailing newline.

src/main/java/com/code_intelligence/jazzer/driver/OptParser.java
```diff
@@ -70,7 +70,7 @@ static String stringSetting(String name, String defaultValue, String description
   }
 
   static List<String> stringListSetting(String name, String description) {
-    return stringListSetting(name, File.pathSeparatorChar, description);
+    return stringListSetting(name, ',', description);
```
fmeum, line 73: Let's keep this breaking change out of this PR and have a separate one for it. I do think we should make it, but we have to think about the implications for users.

src/main/java/jaz/Zer.java
```diff
@@ -50,185 +48,304 @@
...
+  private static final byte EXPRESSION_LANGUAGE_SANITIZER_ID = 2;
+
+  // A byte representing the relevant sanitizer for a given jaz.Zer instance. It is used to check
+  // whether the corresponding sanitizer is disabled and jaz.Zer will not report a finding in this
+  // case. Each sanitizer which relies on this class must set this byte accordingly. We choose a
+  // single byte to represent the sanitizer in order to keep the serialized version of jaz.Zer
+  // objects small (currently 41 bytes) so that it fits in the 64 byte limit of the words that can
+  // be used with Jazzer's methods that guide the fuzzer towards generating inputs that contain or
+  // are equal to target strings. This limit comes from the corresponding libFuzzer hooks that
+  // Jazzer uses under the hoods.
```
fmeum, line 63:
```suggestion
  // Jazzer uses under the hood.
```

src/main/java/jaz/Zer.java
```diff
@@ -50,185 +48,304 @@
...
+  public Zer(byte sanitizer) {
+    this.sanitizer = sanitizer;
+    reportFindingIfEnabled();
+  }
+
+  // A special static method that is called by the expression language injection sanitizer. We
+  // choose a parameterless method to keep the string that the sanitizer guides the fuzzer to
+  // generate within the 64-byte boundary required by the corresponding guiding methods.
+  public static void el() {
+    if (isSanitizerEnabled(EXPRESSION_LANGUAGE_SANITIZER_ID)) {
```
fmeum, line 88: Why can't we use `isSanitizerEnabled(sanitizer)` here?

sanitizers/src/test/java/com/example/el/UserData.java
kyakdan: Done!

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ExpressionLanguageInjection.kt
kyakdan: This was my first attempt, but I couldn't get the string size to be less than 64 bytes while using the constructor that sets the sanitizer ID. The default constructor sets the sanitizer ID to the RCE sanitizer.

Do you think we can reduce the attack string size to less than 64 bytes? `"\${\"\".getClass().forName(\"$HONEYPOT_CLASS_NAME\").getDeclaredConstructor(byte.class).newInstance((byte)2)}"`

sanitizers/src/test/java/com/example/ClassLoaderLoadClass.java
kyakdan: I've reverted this change since it is not used now by the tests for disabled hooks. However, I think the issue is not related to the `jaz.Zer` class per se. If the tested code caches the exceptions we throw and silently handles them, we will not observe a test failure when running in the regression test mode. On the other hand, we throw 
a hard-to-catch error and report a finding. The `FuzzerSecurityIssue[Low|Medium|High|Critical]` exceptions can be handled differently in fuzzing and regression modes.

sanitizers/src/test/java/com/example/BUILD.bazel
kyakdan: Changed to use commas to separate elements of disabled sanitizers

launcher/jvm_tooling.cpp
```diff
@@ -126,10 +118,11 @@ JVM::JVM(const std::string &executable_path) {
...
   const auto class_path_from_env = std::getenv("JAVA_FUZZER_CLASSPATH");
   if (class_path_from_env) {
-    class_path += absl::StrCat(ARG_SEPARATOR, class_path_from_env);
+    class_path += absl::StrCat(CLASSPATH_SEPARATOR, class_path_from_env);
   }
-  class_path +=
-      absl::StrCat(ARG_SEPARATOR, getInstrumentorAgentPath(executable_path));
+  class_path += absl::StrCat(CLASSPATH_SEPARATOR,
+                             getInstrumentorAgentPath(executable_path));
+  std::cerr << "classpath: " << class_path << std::endl;
```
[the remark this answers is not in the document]
kyakdan: As discussed, I removed the commit related to changing the separator in the launcher

launcher/jvm_tooling.h
kyakdan: same here. Dropped the commit related to changing the argument separator in the launcher

src/main/java/jaz/Zer.java
kyakdan: As mentioned before, using the single-argument constructor to create an instance and set the sanitizer ID was not possible since this would require strings that are longer than 64 bytes and cannot be handled in full by libFuzzer. That is why I resorted to defining a static function that should be exclusively called from the expression language injection sanitizers.

sanitizers/src/test/java/com/example/ClassLoaderLoadClass.java
fmeum: I see, that's a problem with our test setup then, right? With JUnit, findings reported by hooks can't be swallowed even in regression test mode since `reportFindingFromHook` stores them in a global variable.

src/main/java/jaz/Zer.java
fmeum: Sounds good. I will codegolf the injected sample for a bit and if I can't find anything that's short enough, this approach is probably the best we can do.

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ExpressionLanguageInjection.kt
fmeum: I tried but didn't find a way. You can get rid of the `(byte)` by changing the constructor parameter type to `int` and also use `getConstructors`, but that's still off by a large margin.

I only found this improvement over the current approach:
```suggestion
        "\${Byte.class.forName(\"$HONEYPOT_CLASS_NAME\").getMethod(\"el\").invoke(null)}"
```

fmeum (approved):

sanitizers/src/test/java/com/example/BUILD.bazel
kyakdan: I've removed this change for now and will create a dedicated PR for changing the separator

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ExpressionLanguageInjection.kt
kyakdan: Good idea. Also added a check for the size of the payloads we use in the deserialization sanitizer

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/Deserialization.kt
```diff
@@ -57,7 +57,7 @@ object Deserialization {
         // We can't instantiate jaz.Zer directly, so we instantiate and serialize jaz.Ter and then
         // patch the class name.
         val baos = ByteArrayOutputStream()
-        ObjectOutputStream(baos).writeObject(jaz.Ter())
+        ObjectOutputStream(baos).writeObject(jaz.Ter(jaz.Ter.EXPRESSION_LANGUAGE_SANITIZER_ID))
```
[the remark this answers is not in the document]
kyakdan: Added!

PR #576, opened 2022-12-30
CI fixes and minor enhancements

fmeum:


src/main/java/com/code_intelligence/jazzer/agent/RuntimeInstrumentor.kt
```diff
@@ -118,6 +121,10 @@ class RuntimeInstrumentor(
             // Throwables raised from transform are silently dropped, making it extremely hard to detect instrumentation
             // failures. The docs advise to use a top-level try-catch.
             // https://docs.oracle.com/javase/9/docs/api/java/lang/instrument/ClassFileTransformer.html
+            if (dumpClassesDir != null) {
```
kyakdan, line 124: This causes the class files to be dumped twice when instrumentation fails. In the `try` block, we call the first `transform` overload, which dumps the class files when instrumentation fails and rethrows the exception.

fmeum: Good catch, thanks. I moved the call of the overload out of the `try` block, which should fix this.

kyakdan (approved):

PR #580, opened 2023-01-02
ci: Do not pin Xcode version in release workflow

fmeum:
We pin the Xcode version in our build & test workflow as detecting host Xcodes is slow and frequently led to timeouts on the slow macOS runners. Since our release workflow uses an older version of macOS, it may not have the pinned Xcode available.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,7 +36,6 @@ jobs:
           echo "build --config=maven" >> .bazelrc
           echo "build:linux --config=toolchain" >> .bazelrc
           echo "build:linux --extra_toolchains=@llvm_toolchain//:cc-toolchain-x86_64-linux" >> .bazelrc
-          echo "build:macos  --xcode_version_config=//.github:host_xcodes" >> .bazelrc
 
       - name: Build
         shell: bash


fmeum:
Should fix https://github.com/CodeIntelligenceTesting/jazzer/actions/runs/3824600555/jobs/6506884018.

.github/workflows/release.yml
```diff
@@ -36,7 +36,6 @@ jobs:
           echo "build --config=maven" >> .bazelrc
           echo "build:linux --config=toolchain" >> .bazelrc
           echo "build:linux --extra_toolchains=@llvm_toolchain//:cc-toolchain-x86_64-linux" >> .bazelrc
-          echo "build:macos  --xcode_version_config=//.github:host_xcodes" >> .bazelrc
```
kyakdan, line 39: Couldn't we unify the macOS versions used in the test and release workflows?

fmeum: The reason we use different versions is supposedly improved backwards compatibility. Seems to be a good time to read up on how Windows and macOS handle this and unify the versions if there is no benefit to using an older one.

kyakdan: Agreed. This is not a blocker to merging this PR.

kyakdan (approved):

PR #585, opened 2023-01-04
junit: Simplify and improve inputs directory discovery

fmeum:
Inputs directory handling was confusing as e.g. Maven doesn't copy over empty resource directories by default. Instead of relying on that, always fall back to locating the inputs directory in the source tree and even try to create it if at least the test resources root is found.

### DIFF
--- a/src/main/java/com/code_intelligence/jazzer/junit/FuzzTestExecutor.java
+++ b/src/main/java/com/code_intelligence/jazzer/junit/FuzzTestExecutor.java
@@ -14,6 +14,8 @@
 
 package com.code_intelligence.jazzer.junit;
 
+import static com.code_intelligence.jazzer.junit.Utils.generatedCorpusPath;
+import static com.code_intelligence.jazzer.junit.Utils.inputsDirectoryResourcePath;
 import static com.code_intelligence.jazzer.junit.Utils.inputsDirectorySourcePath;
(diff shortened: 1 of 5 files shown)


src/main/java/com/code_intelligence/jazzer/junit/FuzzTestExecutor.java
```diff
@@ -69,50 +69,47 @@ public static FuzzTestExecutor prepare(ExtensionContext context, String maxDurat
...
-          inputsDirectoryResourcePath, fuzzTestClass.getPackage().getName());
-      context.publishReportEntry("missing inputs directory", message);
-      libFuzzerArgs.add(String.format("-artifact_prefix=%s%c", baseDir, File.separatorChar));
-    } else if ("file".equals(inputsDirectoryUrl.getProtocol())) {
-      // From the second positional argument on, files and directories are used as seeds but not
-      // modified. Using inputsDirectoryUrl.getFile() fails on Windows.
+    // We can only emit findings into the source tree version of the inputs directory, not e.g. the
+    // copy under Maven's target directory. If it doesn't exist, collect the inputs in the current
+    // working directory, which is usually the project's source root.
+    Optional<Path> findingsDirectory = Utils.inputsDirectorySourcePath(fuzzTestClass, baseDir);
```
bertschneider, line 75: `import static` as the other two functions?

bertschneider (approved):

PR #587, opened 2023-01-05
Add foundations for Android support

TheCoryBarker:
This includes the necessary changes to the native launcher, driver and runtime, but does not yet include a way to instrument the application under test - Android doesn't support Java agents.

Since building the Android targets requires a local installation of the Android SDK and NDK, these targets aren't built by default and only requested in the Linux CI pipeline.


.bazelrc
```diff
@@ -22,6 +22,22 @@ build:ci --features=layering_check
 build --java_language_version=8
 build --tool_java_language_version=8
 
+# Android launcher build config
+build:android_binary --crosstool_top=@androidndk//:toolchain
+build:android_binary --incompatible_enable_cc_toolchain_resolution=false
```
fmeum, line 27: Could you try to remove this line and instead add `--incompatible_enable_android_toolchain_resolution`? All of these flags are [planned to be flipped in Bazel 7](https://github.com/bazelbuild/bazel/issues/7260), so I would prefer to keep using them.

WORKSPACE.bazel
```diff
@@ -146,3 +146,32 @@ http_file(
...
+    strip_prefix = "rules_android-0.1.1",
+    urls = ["https://github.com/bazelbuild/rules_android/archive/v0.1.1.zip"],
+)
+
+load("@build_bazel_rules_android//android:rules.bzl", "android_sdk_repository")
+
+android_sdk_repository(
+    name = "androidsdk",
+    #path = "/usr/local/google/home/cobark/Desktop/Tools/android-sdk",
+    path = "/usr/local/google/home/cobark/Android/Sdk",
```
fmeum, line 162: If `path` isn't set, the `android_sdk_repository` rule will look into `ANDROID_HOME`, which is available in CI (see https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#environment-variables-3).

WORKSPACE.bazel
```diff
@@ -146,3 +146,32 @@ http_file(
...
+    sha256 = "3fa4a58f4df356bca277219763f91c64f33dcc59e10843e9762fc5e7947644f9",
+    strip_prefix = "rules_android_ndk-63fa7637902fb1d7db1bf86182e939ed3fe98477",
+    url = "https://github.com/bazelbuild/rules_android_ndk/archive/63fa7637902fb1d7db1bf86182e939ed3fe98477.zip",
+)
+
+load("@rules_android_ndk//:rules.bzl", "android_ndk_repository")
+
+android_ndk_repository(
+    name = "androidndk",
+    path = "/usr/local/google/home/cobark/Desktop/Tools/android-ndk-r25b",
```
fmeum, line 176: Same as for `android_sdk_repository`.

WORKSPACE.bazel
```diff
@@ -146,3 +146,32 @@ http_file(
...
+    name = "androidsdk",
+    #path = "/usr/local/google/home/cobark/Desktop/Tools/android-sdk",
+    path = "/usr/local/google/home/cobark/Android/Sdk",
+)
+
+http_archive(
+    name = "rules_android_ndk",
+    sha256 = "3fa4a58f4df356bca277219763f91c64f33dcc59e10843e9762fc5e7947644f9",
+    strip_prefix = "rules_android_ndk-63fa7637902fb1d7db1bf86182e939ed3fe98477",
+    url = "https://github.com/bazelbuild/rules_android_ndk/archive/63fa7637902fb1d7db1bf86182e939ed3fe98477.zip",
```
fmeum, line 169: Could you update this to the most recent version?

launcher/jazzer_main.cpp
```diff
@@ -27,6 +27,12 @@
 #include "absl/strings/str_split.h"
 #include "jvm_tooling.h"
 
+#if defined(_ANDROID)
+#define android true
```
fmeum, line 31: Please use uppercase names for macros so that they can't clash with local variables.

src/main/java/com/code_intelligence/jazzer/Jazzer.java
```diff
@@ -85,6 +87,10 @@ private static void start(List<String> args) throws IOException, InterruptedExce
     }
     // No native fuzzing has been requested, fuzz in the current process.
     if (!fuzzNative) {
+      if(isAndroid()){
```
fmeum, line 90: Why is this special case needed?

src/main/java/com/code_intelligence/jazzer/Jazzer.java
```diff
@@ -197,28 +213,53 @@ private static String prepareArgv0(Map<String, String> additionalEnvironment) th
...
     String invocation = env.isEmpty() ? command : env + " " + command;
     String launcherContent = String.format(launcherTemplate, invocation);
-    Path launcher = Files.createTempFile("jazzer-", launcherExtension, launcherScriptAttributes);
+
+    Path launcher;
+    if(!isAndroid()){
+      launcher = Files.createTempFile("jazzer-", launcherExtension, launcherScriptAttributes);
+    }
+    else {
+      launcher = Files.createTempFile(Paths.get("/data/local/tmp/"), "jazzer-", launcherExtension, launcherScriptAttributes);
```
fmeum, line 222: Is `Files.createTempFile` available on Android, but `Files.createTempDirectory` isn't? Similar to the rules_jni PR, can we assume that this directory is always correct and exists?

src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetFinder.java
```diff
@@ -82,6 +86,10 @@ static String findFuzzTargetClassName() {
    * @return a {@link FuzzTarget}
    */
   static FuzzTarget findFuzzTarget(Class<?> clazz) {
+    if (Opt.isAndroid) {
```
fmeum, line 89: Returning early in `findFuzzTargetByAnnotation` instead would be cleaner.

src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetRunner.java
```diff
@@ -55,7 +55,9 @@
  */
 public final class FuzzTargetRunner {
   static {
-    AgentInstaller.install(Opt.hooks);
+    if(!Opt.isAndroid) {
+      AgentInstaller.install(Opt.hooks);
```
fmeum, line 59: This isn't just starting the agent, but also adds certain classes to the bootstrap classpath that instrumented classes may access. Is that something that can be done on Android? 

It may only be necessary when system classes are instrumented though, so it might not even apply to Android.

src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetRunner.java
```diff
@@ -278,7 +280,9 @@ private static int runOne(long dataPtr, int dataLength) {
    * Starts libFuzzer via LLVMFuzzerRunDriver.
    */
   public static int startLibFuzzer(List<String> args) {
-    SignalHandler.initialize();
+    if(!Opt.isAndroid) {
+      SignalHandler.initialize();
```
fmeum, line 284: Let's move this into check into `SignalHandler.initialize()`.

src/main/java/com/code_intelligence/jazzer/runtime/CoverageMap.java
```diff
@@ -96,9 +99,68 @@ public static void enlargeIfNeeded(int nextId) {
...
+    }
+    finally {
+      rl.unlock();
+    }
+
+    return new Integer(result);
+  }
+
+  public static void recordCoverageRemapped(final int id){
+    Integer mappedId = edgeMap.get(id);
```
fmeum, line 135: The agent contains [this abstraction](https://github.com/CodeIntelligenceTesting/jazzer/blob/6b9a31a650441d068bbc92e4f94e7d2b85d79bcf/src/main/java/com/code_intelligence/jazzer/agent/RuntimeInstrumentor.kt#L206-L215) for a similar purpose (multi-threaded fuzzing). 

We are also thinking of supporting offline instrumentation with Jazzer on all platforms, so maybe this could be integrated in a more general way. But I can think about that while we get this working for Android.

src/main/java/com/code_intelligence/jazzer/runtime/CoverageMap.java
```diff
@@ -96,9 +99,68 @@ public static void enlargeIfNeeded(int nextId) {
     }
   }
 
+  private static ConcurrentHashMap<Integer, Integer> edgeMap = new ConcurrentHashMap<Integer, Integer>();
+  private static int nextEdgeId = 0;
+  static ReentrantLock rl = new ReentrantLock();
+  private static Integer initializeEdgeForAndroid(int id) throws InterruptedException {
+    // while loop to run until id gets lock
```
fmeum, line 106: Could you explain what this map is used for? It does seem to map Jazzer's edge IDs obtained from offline instrumentation to consecutive IDs based on the order in which the edges are encountered at runtime. Is that really necessary? It seems that offline instrumentation could just emit globally unique IDs that Jazzer could use, possible enlargening the coverage map if an ID it encounters is larger than the current size.

Even if this remapping is required, I don't see why a lock around the `ConcurrentHashMap` is needed. It looks like `ConcurrentHashMap#computeIfAbsent` combined with `AtomicInteger#getAndIncrement` could be sufficient (and potentially much more performant and simple).

TheCoryBarker: You're right this re-map might not be needed. It is to keep the shared array with LibFuzzer smaller, since there can be some very large .jar files when fuzzing the Android Java framework. I didn't want to have a fuzz target that hits maybe 400 edges but has an array of size 2500,000 because it hit edge ID 250000 first.

Maybe the goal here should be to create smaller fuzz targets though rather than remapping the IDs.

src/main/java/com/code_intelligence/jazzer/Jazzer.java
TheCoryBarker: and also yes Files.createTempFile does work on Android

src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetRunner.java
TheCoryBarker: We are not able to use ByteBuddy on Android, so ByteBuddyAgent.install() will fail. Instead of loading in the class through the Bootstrap jar I have added the needed deps to android_binary that builds "jazzer_standalone_android".

src/main/java/com/code_intelligence/jazzer/runtime/CoverageMap.java
fmeum: A large coverage map by itself shouldn't hurt performance, but instrumenting low-level system library code that runs in critical paths might. Preventing that would be the job of an appropriate Instrumentation filter though, so I think we can drop this remapping logic for now.

sanitizers/BUILD.bazel
```diff
@@ -1,6 +1,8 @@
 java_library(
     name = "sanitizers",
-    visibility = ["//src/main/java/com/code_intelligence/jazzer/runtime:__pkg__"],
+    visibility = [
+        "//visibility:public",
```
fmeum, line 4: Some of the sanitizers need to be on the bootstrap classpath as they have to be visible to e.g. `java.util.regex.Pattern`. The `runtime` package is bundled into a JAR that is appended to the boostrap classpath at runtime. This visibility declaration ensures that we don't accidentally mess up this separation, so I would like to preserve it. 

Is there a way to add classes to the bootstrap classpath on Android? I would expect it to be necessary as otherwise Jazzer may crash when e.g. `Pattern#compile` is called.

launcher/jvm_tooling.cpp
```diff
@@ -14,6 +14,8 @@
 
 #include "jvm_tooling.h"
 
+#include <dlfcn.h>
```
fmeum, line 17: Please add `-ldl` as a `linkopt` to the `jvm_tooling` target.

src/main/java/com/code_intelligence/jazzer/AndroidManifest.xml
```diff
@@ -0,0 +1,5 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="com.code_intelligence.jazzer"
+    android:versionCode="1"
+    android:versionName="1.0" >
+</manifest>
```
fmeum, line 5: Nit: Add the missing newline.

src/main/java/com/code_intelligence/jazzer/Jazzer.java
```diff
@@ -180,7 +198,15 @@ private static String prepareArgv0(Map<String, String> additionalEnvironment) th
           "Setting environment variables in the wrapper is only supported on POSIX systems");
     }
     char shellQuote = isPosix() ? '\'' : '"';
-    String launcherTemplate = isPosix() ? "#!/usr/bin/env sh\n%s $@\n" : "@echo off\r\n%s %%*\r\n";
+    String launcherTemplate;
+    if (isAndroid()) {
+      launcherTemplate = "%s $@\n";
```
fmeum, line 203: Where does the shell live under Android? Could we make changes to the posix template so that it also works on Android?

src/main/java/com/code_intelligence/jazzer/Jazzer.java
```diff
@@ -197,28 +223,51 @@ private static String prepareArgv0(Map<String, String> additionalEnvironment) th
                          .collect(joining(" "));
     String invocation = env.isEmpty() ? command : env + " " + command;
     String launcherContent = String.format(launcherTemplate, invocation);
-    Path launcher = Files.createTempFile("jazzer-", launcherExtension, launcherScriptAttributes);
+
+    Path launcher;
+    if (!isAndroid()) {
```
fmeum, line 228: Flipping the branches gets rid of the negation.

src/main/java/com/code_intelligence/jazzer/Jazzer.java
```diff
@@ -238,6 +287,26 @@ private static String appendWithPathListSeparator(String name, String... options
     return currentValue + File.pathSeparator + additionalOptions;
   }
 
+  private static Path findLibrary(List<String> candidateNames) {
+    if (isAndroid()) {
```
fmeum, line 291: Flipping the condition makes it so that the simple common case is handled first.

src/main/java/com/code_intelligence/jazzer/Jazzer.java
```diff
@@ -238,6 +287,26 @@ private static String appendWithPathListSeparator(String name, String... options
...
+      for (String candidateName : candidateNames) {
+        String candidateFullPath = "/apex/com.android.runtime/lib64/bionic/" + candidateName;
```
fmeum, line 293: Can we expect this path to be relatively stable?

src/main/java/com/code_intelligence/jazzer/Jazzer.java
```diff
@@ -321,6 +411,10 @@ private static boolean isPosix() {
     return FileSystems.getDefault().supportedFileAttributeViews().contains("posix");
   }
 
+  private static boolean isAndroid() {
+    return Boolean.parseBoolean(System.getProperty("jazzer.android", "false"));
```
fmeum, line 415: Just to be safe, could we add `!isAndroid()` to `isPosix()` in case the latter currently returns `true` on Android?

fmeum:
Don't worry about the conflicts, I can fix those when we are done with the review.

launcher/jvm_tooling.cpp
```diff
@@ -166,30 +213,61 @@ JVM::JVM(const std::string &executable_path) {
...
+  std::string class_path = GetClassPath(executable_path, android);
+
+  std::vector<JavaVMOption> options;
+  options.push_back(
+      JavaVMOption{.optionString = const_cast<char *>(class_path.c_str())});
+  SetVMOptions(options, android);
+
+  jint jni_version = JNI_VERSION_1_8;
+  if (android) {
+    jni_version = JNI_VERSION_1_6;
```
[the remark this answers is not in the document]
TheCoryBarker: The why only up to 6 is supported, I dont know, but I know that Android only supports up to version 6 on JNI. 

There are people who know much more about this than me, but from my understanding we can build for Java 9 or 8, but after we desugar and dex we are going to have Java 7 with a subset of 8.

I added the java_language_version=9 because some flags I was using to desugar needed it, but I removed those. I will remove the java_language_version=9 and retest tomorrow to make sure that doesn't effect anything (it shouldn't) and push up that change.

sanitizers/BUILD.bazel
TheCoryBarker: Ack, I will look into this more.

src/main/java/com/code_intelligence/jazzer/Jazzer.java
TheCoryBarker: I only have access to Pixels, but across Pixels this path is consistent, and after talking with some of my teammates, everything under apex should be consistent across brands. 

Although, this is the hwasan fuzzing path, and this library will only be there on a hwasan Android builds, and as far as I know Google might be the only company currently providing hwasan builds.

https://developer.android.com/ndk/guides/hwasan
https://ci.android.com/builds/branches/aosp-master-with-phones-throttled/grid?
https://flash.android.com/

TheCoryBarker: However, even if you dont have a hwasan build, you can still fuzz Java

sanitizers/BUILD.bazel
TheCoryBarker: I have tested all the example sanitizers, and they all either catch the security error or they run forever, but none crash due to not being able to find the Code Intelligence class from the runtime package.

OsCommandInjectionProcessBuilder
OsCommandInjectionRuntimeExec
RegexRoadblocks
StackOverflowRegexInjection 

These examples all run forever. Is this possibly because we are instrumenting offline, so Pattern#compile never gets instrumented so it never catches the security issue, but also doesn't crash or need these classes loaded to the bootstrap classpath?

Most of the sanitization examples work however, and this was also expected that some sanitizers wouldn't work if we did offline sanitization.

This being said, is it still important to find a way to load these classes into the bootstrap classpath? Or would an acceptable alternative be to create another bazel build target like this:

java_library (
  name = "offline_sanitizers",
  <srcs = sanitizers needed for offline running of Jazzer>
  ...
)

fmeum: It's fine not to load them. 

Could you create a target next to this one called `offline_only_sanitizers` that has the required visibility and use it for this purpose? You don't need to filter out the sanitizers that don't work just yet.

src/main/java/com/code_intelligence/jazzer/Jazzer.java
fmeum: Sounds good, thanks.

src/main/java/com/code_intelligence/jazzer/Jazzer.java
TheCoryBarker: /system/bin/sh 

I'll update this to: /system/bin/env sh

TheCoryBarker:
fixed comments

.idea/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Default ignored files
```
fmeum, line 1: Please remove these files, feel free to add `/.idea/` or similar to `.gitignore`.

src/main/java/com/code_intelligence/jazzer/Jazzer.java
```diff
@@ -288,6 +289,10 @@ private static String appendWithPathListSeparator(String name, String... options
   }
 
   private static Path findLibrary(List<String> candidateNames) {
+    if (!isAndroid()) {
+      return findHostClangLibrary(candidateNames);
+    }
+
     if (isAndroid()) {
```
fmeum, line 296: This seems redundant now?

TheCoryBarker: removed extra if

.idea/.gitignore
TheCoryBarker: removed .idea dir

launcher/jvm_tooling.cpp
```diff
@@ -166,30 +220,61 @@ JVM::JVM(const std::string &executable_path) {
...
+  if (android) {
+    if (CreateArtVM == nullptr) {
+      std::cerr << "JNI_CreateJavaVM for Android not found" << std::endl;
+      exit(1);
+    }
+
+    std::cout << "Starting Art VM" << std::endl;
+    ret = CreateArtVM(&jvm_, (JNIEnv_ **)&env_, &jvm_init_args);
+  } else {
+    std::cout << "Starting JVM" << std::endl;
```
fmeum, line 268:
```suggestion
```
Sorry, missed this earlier.

launcher/jvm_tooling.cpp
```diff
@@ -166,30 +220,61 @@ JVM::JVM(const std::string &executable_path) {
...
+    ret = JNI_CreateJavaVM(&jvm_, (void **)&env_, &jvm_init_args);
+  }
+
   if (ret != JNI_OK) {
     throw std::runtime_error(
         absl::StrFormat("JNI_CreateJavaVM returned code %d", ret));
   }
+
+  std::cout << "VM started successfully" << std::endl;
```
fmeum, line 277:
```suggestion
```

src/main/java/com/code_intelligence/jazzer/runtime/CoverageMap.java
```diff
@@ -96,6 +99,38 @@ public static void enlargeIfNeeded(int nextId) {
     }
   }
 
+  private static ConcurrentHashMap<Integer, Integer> edgeMap =
+      new ConcurrentHashMap<Integer, Integer>();
+  private static int nextEdgeId = 0;
+  static ReentrantLock rl = new ReentrantLock();
+  private static Integer initializeEdgeForAndroid(int id) throws InterruptedException {
```
fmeum, line 106: Is this called anywhere or can it be cleaned up?

fmeum:
@TheCoryBarker I will resolve the conflicts and then check the CI status.

TheCoryBarker:
There is an issue with non-Android builds and ANDROID_NDK_HOME needing to be set. For me this is not an issue but for most people this will be. I will look into a fix for this

fmeum:
@TheCoryBarker I merged in the recent changes from main.

TheCoryBarker:
ANDROID_NDK_HOME and ANDROID_HOME needing to be set is fixed now, however, the test to build all targets will fail for Android targets. How can we fix this?

TheCoryBarker:
I found this as another solution which would probably work better in the long run. Can we update the CI command to run:

bazel query '//... except kind(android.*, //...)' | xargs bazel build

fmeum:
We can do that if nothing else works, but I would like to preserve the assumption that `bazel build //...` works locally and also that core Jazzer can be built as an external dependency of other Bazel projects. The latter means that packages providing targets relevant for core Jazzer must not load from unrelated rulesets (e.g. rules_android), which may anyway require moving Android targets to a different package. 

@TheCoryBarker If you want, I can take a look and propose a fix.

TheCoryBarker:
I moved the Android build targets into their own package, but I'm not sure how to exclude them from being built in //..., if you could propose a fix that would be really great

fmeum:
I made it so that Android targets aren't built locally unless requested explicitly, but are built as part of the Linux CI build. I also made it so that the `android_configure` rule disables itself on Windows.

@TheCoryBarker Windows CI runs are failing since `jvm_tooling.cpp` unconditionally includes `dlfcn.h`. Maybe replace the `bool android` parameter with preprocessor conditionals?

TheCoryBarker:
Thank you, I am having a very tough time getting a Windows machine with correct permissions to test my changes on...

TheCoryBarker:
Thanks for the suggestion Fabian, I updated to use preprocessor conditionals

src/main/native/com/code_intelligence/jazzer/driver/BUILD.bazel
```diff
@@ -35,7 +35,10 @@ cc_library(
     name = "coverage_tracker",
     srcs = ["coverage_tracker.cpp"],
     hdrs = ["coverage_tracker.h"],
-    deps = ["//src/main/java/com/code_intelligence/jazzer/runtime:coverage_map.hdrs"],
+    deps = [
+        "//src/main/java/com/code_intelligence/jazzer/runtime:coverage_map.hdrs",
+        "@jazzer_libfuzzer//:libfuzzer_no_main",
```
fmeum, line 40: Is this needed? I think that it ends up loading libFuzzer twice (once here and once in FuzzTargetRunnerNatives).

fmeum:
@TheCoryBarker I fixed some minor issues, CI looks good now. I left a single comment and will merge when it's resolved.

fmeum (approved):

fmeum:
@TheCoryBarker Thanks for resolving that problem. I squashed the commits, added a commit message and hit merge.

Issue #586, opened 2023-01-04
AutofuzzError: popGroup must be called exactly once for every pushGroup


freedom1b2830:
0.14.0
args ``--autofuzz=test.TestServer::start --autofuzz_ignore=java.lang.NullPointerException``
```
INFO: Loaded 1802 hooks from com.code_intelligence.jazzer.runtime.TraceCmpHooks
INFO: Loaded 4 hooks from com.code_intelligence.jazzer.runtime.TraceDivHooks
INFO: Loaded 2 hooks from com.code_intelligence.jazzer.runtime.TraceIndirHooks
INFO: Loaded 4 hooks from com.code_intelligence.jazzer.runtime.NativeLibHooks
INFO: Loaded 5 hooks from com.code_intelligence.jazzer.sanitizers.Deserialization
INFO: Loaded 5 hooks from com.code_intelligence.jazzer.sanitizers.ExpressionLanguageInjection
INFO: Loaded 70 hooks from com.code_intelligence.jazzer.sanitizers.LdapInjection
INFO: Loaded 46 hooks from com.code_intelligence.jazzer.sanitizers.NamingContextLookup
INFO: Loaded 1 hooks from com.code_intelligence.jazzer.sanitizers.OsCommandInjection
INFO: Loaded 76 hooks from com.code_intelligence.jazzer.sanitizers.ReflectiveCall
INFO: Loaded 8 hooks from com.code_intelligence.jazzer.sanitizers.RegexInjection
INFO: Loaded 16 hooks from com.code_intelligence.jazzer.sanitizers.RegexRoadblocks
INFO: Loaded 19 hooks from com.code_intelligence.jazzer.sanitizers.SqlInjection
INFO: Instrumented test.Fuzzer (took 716 ms, size +17%)
INFO: Instrumented test.TestServer (took 184 ms, size +22%)
INFO: libFuzzer ignores flags that start with '--'
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 3323939128
INFO: Loaded 1 modules   (512 inline 8-bit counters): 512 [0x7f8fbc93b9e0, 0x7f8fbc93bbe0), 
INFO: Loaded 1 PC tables (512 PCs): 512 [0x7f8fbc8d0a70,0x7f8fbc8d2a70), 
INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 4096 bytes
INFO: A corpus is not provided, starting from an empty corpus
[...]

PR #588, opened 2023-01-05
autofuzz: Do not report finding for failed autofuzz in consume

fmeum:
When an invocatin of `autofuzz` in `consume` fails, this is not a finding but rather a failure to construct a valid object.


fmeum:
I will try to move the catch to a more central location.

fmeum:
@kyakdan I made the fix more general, please take a look.

kyakdan (approved):

fmeum:
Fixes #586.

PR #593, opened 2023-01-08
agent: Hook Kotlin's compare method for integral types

kyakdan:
Kotlin's integral types have `compareTo` methods that are compiled to calls into one of two `compare` overloads from the `kotlin.jvm.internal.Intrinsics` class:
1. `public static int compare(long thisVal, long anotherVal)`
2. `public static int compare(int thisVal, int anotherVal)`

### DIFF
--- a/examples/BUILD.bazel
+++ b/examples/BUILD.bazel
@@ -373,6 +373,28 @@ java_fuzz_target_test(
     runtime_deps = [":ExampleKotlinFuzzTarget"],
 )
 
+kt_jvm_library(
+    name = "ExampleKotlinValueProfileFuzzTarget",
+    srcs = [
+        "src/main/java/com/example/ExampleKotlinValueProfileFuzzer.kt",
+    ],
+    deps = [
+        "//deploy:jazzer-api",
+    ],
+)
+
+java_fuzz_target_test(
+    name = "ExampleKotlinValueProfileFuzzer",
+    allowed_findings = [
+        "com.code_intelligence.jazzer.api.FuzzerSecurityIssueMedium",
+    ],
+    fuzzer_args = [
+        "-use_value_profile=1",
+    ],
+    target_class = "com.example.ExampleKotlinValueProfileFuzzer",
+    runtime_deps = [":ExampleKotlinValueProfileFuzzTarget"],
+)
+
 java_fuzz_target_test(
     name = "TurboJpegFuzzer",
     srcs = [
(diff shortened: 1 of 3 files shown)


examples/BUILD.bazel
```diff
@@ -373,6 +373,28 @@ java_fuzz_target_test(
     runtime_deps = [":ExampleKotlinFuzzTarget"],
 )
 
+kt_jvm_library(
+    name = "ExampleKotlinValueProfileFuzzeTarget",
```
fmeum, line 377:
```suggestion
    name = "ExampleKotlinValueProfileFuzzTarget",
```

examples/BUILD.bazel
```diff
@@ -373,6 +373,28 @@ java_fuzz_target_test(
...
+java_fuzz_target_test(
+    name = "ExampleKotlinValueProfileFuzzer",
+    allowed_findings = [
+        "com.code_intelligence.jazzer.api.FuzzerSecurityIssueMedium",
+    ],
+    fuzzer_args = [
+        "-use_value_profile=1",
+    ],
+    target_class = "com.example.ExampleKotlinValueProfileFuzzer",
+    runtime_deps = [":ExampleKotlinValueProfileFuzzeTarget"],
```
fmeum, line 395:
```suggestion
    runtime_deps = [":ExampleKotlinValueProfileFuzzTarget"],
```

examples/src/main/java/com/example/ExampleKotlinValueProfileFuzzer.kt
```diff
@@ -0,0 +1,37 @@
...
+package com.example
+
+import com.code_intelligence.jazzer.api.FuzzedDataProvider
+import com.code_intelligence.jazzer.api.FuzzerSecurityIssueMedium
+
+object ExampleKotlinValueProfileFuzzer {
+
+    @JvmStatic
+    fun fuzzerTestOneInput(data: FuzzedDataProvider) {
+        if (data.consumeInt().compareTo(0x1122) != 0) {
```
fmeum, line 24: We should probably make this a 32-bit comparison as 16 bit values plus null bytes are more easily guessed by chance.

fmeum (approved):

examples/BUILD.bazel
kyakdan: Done.

examples/BUILD.bazel
kyakdan: Done.

examples/src/main/java/com/example/ExampleKotlinValueProfileFuzzer.kt
kyakdan: Good point. Adjusted.

PR #596, opened 2023-01-11
Centralize logging

fmeum:


src/main/java/com/code_intelligence/jazzer/utils/Log.java
```diff
@@ -0,0 +1,101 @@
...
+import java.io.PrintStream;
+
+/**
+ * Provides static functions that should be used for any kind of output (structured or unstructured)
+ * emitted by the fuzzer.
+ *
+ * <p>Output is printed to {@link System#err} and {@link System#out} until {@link
+ * Log#fixOutErr(PrintStream, PrintStream)} is called, which locks in the {@link PrintStream}s to be
+ * used from there on.
+ */
```
bertschneider, line 28: Documenting which functions write to `System.out` and which to `System.err` would be helpful.

src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetRunner.java
```diff
@@ -61,7 +61,23 @@ public final class FuzzTargetRunner {
...
       "org.opentest4j.TestAbortedException";
 
-  private static final Unsafe UNSAFE = UnsafeProvider.getUnsafe();
+  private static final Unsafe UNSAFE;
+
+  static {
+    Unsafe unsafe;
+    try {
+      unsafe = UnsafeProvider.getUnsafe();
+    } catch (Throwable t) {
```
bertschneider, line 69: Is catching `Throwable` really necessary or would `Exception` suffice? It's probably not an issue, but catching `Throwable` always seems extensive.

src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetRunner.java
```diff
@@ -111,7 +126,7 @@ public final class FuzzTargetRunner {
     try {
       fuzzTarget = FuzzTargetFinder.findFuzzTarget(fuzzTargetClass);
     } catch (IllegalArgumentException e) {
-      err.printf("ERROR: %s%n", e.getMessage());
+      Log.error(e);
```
bertschneider, line 129: This will not print the error message anymore.

src/main/java/com/code_intelligence/jazzer/driver/Opt.java
```diff
@@ -144,27 +144,27 @@ public final class Opt {
     OptParser.failOnUnknownArgument();
 
     if (help) {
-      err.println(OptParser.getHelpText());
+      Log.println(OptParser.getHelpText());
       exit(0);
     }
     if (version) {
-      err.printf("Jazzer v%s", JAZZER_VERSION);
+      Log.error("Jazzer v" + JAZZER_VERSION);
```
bertschneider, line 151: Should be `Log.println` as it's no error.

src/main/java/com/code_intelligence/jazzer/instrumentor/Hooks.kt
```diff
@@ -81,12 +82,12 @@ data class Hooks(
...
                         Class.forName(hookClassName, true, Companion::class.java.classLoader)
                     loadHooks(hookClass).also {
-                        println("INFO: Loaded ${it.size} hooks from $hookClassName")
+                        Log.info("Loaded ${it.size} hooks from $hookClassName")
                     }.map {
                         it to hookClass
                     }
                 } catch (e: ClassNotFoundException) {
-                    println("WARN: Failed to load hooks from $hookClassName:")
+                    Log.warn("Failed to load hooks from $hookClassName:")
```
bertschneider, line 90: Combine with exception and remove next line.

tests/src/test/java/com/example/SilencedFuzzer.java
```diff
@@ -0,0 +1,40 @@
...
+
+  public static void fuzzerInitialize() {
+    System.setErr(noopStream);
+    System.setOut(noopStream);
+  }
+
+  public static void fuzzerTestOneInput(byte[] input) {
+    System.err.println("very noisy");
+    System.out.println("even more noise");
+    throw new FuzzerSecurityIssueHigh();
```
bertschneider, line 38: What does this test verify? If it's used to manually check the output, we should mark it as such in Bazel.

bertschneider (commented):
Nice change! Some minor comments.

src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetRunner.java
fmeum: Actually, now that I look at this again, `getUnsafe` doesn't throw anything as `getUnsafeInternal` is called eagerly. I simplified the exception handling accordingly.

src/main/java/com/code_intelligence/jazzer/utils/Log.java
fmeum: I didn't do this on purpose: We might want to change it if we introduce more structured logging (e.g. to a file or as JSON) or rely more heavily on JUnit mechanisms to report messages. I think that distinguishing by message type (error, finding, structured output, ...) should be sufficient information for the call site.

Do you see a particular benefit in documenting what writes where?

src/main/java/com/code_intelligence/jazzer/utils/Log.java
```diff
@@ -0,0 +1,101 @@
...
+
+  public static void error(String message) {
+    error(message, null);
+  }
+
+  public static void error(Throwable t) {
+    error(null, t);
+  }
+
+  public static void error(String message, Throwable t) {
```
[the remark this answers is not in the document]
fmeum: Assuming that we never want to show a stack trace in an `INFO` message, I think that the only missing overload is `warn(Throwable)`. While we are using `error(Throwable)` in cases where we can't really add context to an unexpected error, I don't think we should ever emit a stack trace as a warning without context - after all, if we can't add context, how do we know that it is safe to continue? Do you see a case where we would want to do that? Not having the method may nudge us into not doing it :-)

tests/src/test/java/com/example/SilencedFuzzer.java
fmeum: Added a comment explaining how the test is supposed to work and removed the unnecessary output.

src/main/java/com/code_intelligence/jazzer/utils/Log.java
```diff
@@ -0,0 +1,101 @@
...
+    err.print(prefix);
+    if (message != null) {
+      err.print(message + ":\n");
+    }
+    if (t != null) {
+      t.printStackTrace(err);
+    }
+  }
+
+  private static PrintStream getOut() {
```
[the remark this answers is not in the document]
fmeum: Done.

src/main/java/com/code_intelligence/jazzer/instrumentor/Hooks.kt
fmeum: Changed.

src/main/java/com/code_intelligence/jazzer/driver/Opt.java
fmeum: Fixed.

src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetRunner.java
fmeum: It would, but it would also emit the stack trace. Instead, I changed it to again print the message only.

src/main/java/com/code_intelligence/jazzer/utils/Log.java
bertschneider: No, that clarifies it. Let's leave it as is.

tests/src/test/java/com/example/SilencedFuzzer.java
bertschneider: Ah, now I understand the idea of this test. Thx.

bertschneider (approved):

PR #597, opened 2023-01-12
driver: only create a wrapper script in modes that create sub-processes

kyakdan:


src/main/java/com/code_intelligence/jazzer/Jazzer.java
```diff
@@ -85,9 +85,15 @@ private static void start(List<String> args) throws IOException, InterruptedExce
...
-      args =
-          Stream.concat(Stream.of(prepareArgv0(new HashMap<>())), args.stream()).collect(toList());
-      exit(Driver.start(args));
+      // We only create a wrapper script if libFuzzer runs in a mode that creates subprocesses.
+      // In LibFuzzer's fork mode, the subprocesses created continously by the main libFuzzer
+      // process do not create further subprocesses. Creating a wrapper script for each subprocess
+      // is an unnecessary overhead.
+      final boolean spawnsSubprocesses = args.stream().anyMatch(
+          arg -> arg.startsWith("-fork=") || arg.startsWith("-jobs=") || arg.startsWith("-merge="));
+      String arg0 = spawnsSubprocesses ? prepareArgv0(new HashMap<>()) : "jazzer";
```
fmeum, line 94: Could we make this something more like this:

```suggestion
      String arg0 = spawnsSubprocesses ? prepareArgv0(new HashMap<>()) : "unused_report_a_bug_if_you_see_this";
```

I am worried that we missed something here and users are confused when `jazzer` ends up being searched in path.

fmeum (approved):

src/main/java/com/code_intelligence/jazzer/Jazzer.java
kyakdan: Good point. Changed.

PR #601, opened 2023-01-16
junit: Add maven and gradle classes to custom hook excludes

florianGla:


src/main/java/com/code_intelligence/jazzer/junit/AgentConfigurator.java
```diff
@@ -52,6 +52,8 @@ static void forFuzzing(ExtensionContext executionRequest) {
     String filter = String.join(File.pathSeparator, instrumentationFilter.split(","));
     System.setProperty("jazzer.custom_hook_includes", filter);
     System.setProperty("jazzer.instrumentation_includes", filter);
+    System.setProperty("jazzer.instrumentation_excludes",
+        String.join(File.pathSeparator, "org.apache.maven.**, org.gradle.**"));
```
fmeum, line 56: These classes *should* not be matched by the default instrumentation filter we compute in line 51. Do the projects with which you are experiencing issues have `jazzer.instrument` set?

fmeum, line 56: Could you move these exclusions to the list below in `applyCommonConfiguration` and check whether that already fixes the issue?

florianGla: I don't think so.

src/main/java/com/code_intelligence/jazzer/junit/AgentConfigurator.java
```diff
@@ -60,6 +60,6 @@ private static void applyCommonConfiguration() {
         String.join(File.pathSeparator, "com.google.testing.junit.**", "com.intellij.**",
             "io.github.classgraph.**", "junit.framework.**", "net.bytebuddy.**",
             "org.apiguardian.**", "org.assertj.core.**", "org.hamcrest.**", "org.junit.**",
-            "org.opentest4j.**"));
+            "org.opentest4j.**", "org.apache.maven.**, org.gradle.**"));
```
fmeum, line 63:
```suggestion
            "org.opentest4j.**", "org.apache.maven.**", "org.gradle.**"));
```

fmeum (approved):

PR #604, opened 2023-01-17
junit: Use JUnit-provided test instance

fmeum:
This ensures compatibility with non-trivial test class constructors and test instance post processing, which is common with mocking frameworks such as Mockito.

### DIFF
--- a/examples/junit/src/test/java/com/example/LifecycleFuzzTest.java
+++ b/examples/junit/src/test/java/com/example/LifecycleFuzzTest.java
@@ -26,13 +26,20 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.MethodOrderer;
 import org.junit.jupiter.api.TestMethodOrder;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.TestInstancePostProcessor;
 
 @TestMethodOrder(MethodOrderer.MethodName.class)
+@ExtendWith(LifecycleFuzzTest.LifecycleInstancePostProcessor.class)
 class LifecycleFuzzTest {
   private static int beforeAllCount = 0;
-  private static int beforeEachCount = 0;
-  private static int afterEachCount = 0;
+  private static int beforeEachGlobalCount = 0;
+  private static int afterEachGlobalCount = 0;
   private static int afterAllCount = 0;
+
+  private boolean beforeEachCalledOnInstance = false;
+  private boolean testInstancePostProcessorCalledOnInstance = false;
 
   @BeforeAll
   static void beforeAll() {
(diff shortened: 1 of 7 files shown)


examples/junit/src/test/java/com/example/LifecycleFuzzTest.java
```diff
@@ -42,6 +49,7 @@ static void beforeAll() {
   @BeforeEach
   void beforeEach() {
     beforeEachCount++;
+    beforeEachCalled = true;
```
kyakdan, line 52: Why is `beforeEachCalled` needed? We already have `beforeEachCount` to check that `beforeEach` is called.

src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetFinder.java
```diff
@@ -143,28 +147,31 @@ private static Optional<FuzzTarget> findFuzzTargetByAnnotation(Class<?> clazz) {
...
-          clazz.getName()));
-    }
-    Constructor<?> constructor = clazz.getDeclaredConstructors()[0];
-    // JUnit 5 supports injected constructor parameters, but we don't.
-    if (constructor.getParameterCount() > 0) {
-      throw new IllegalArgumentException(String.format(
-          "The constructor of a class containing a method annotated with @FuzzTest must take no parameters, got a non-zero number in %s",
-          clazz.getName()));
+    Callable<Object> newInstance;
+    if (FuzzTargetInstanceHolder.fuzzTargetInstance != null) {
```
kyakdan, line 151: Could you clarify this check? If I see it correctly, `FuzzTargetInstanceHolder.fuzzTargetInstance` is only set by `FuzzTestExecutor` when fuzzing from JUnit. In which cases is the fuzz test instance not `null` when finding the fuzz test in the driver?

examples/junit/src/test/java/com/example/LifecycleFuzzTest.java
fmeum: I renamed the variables to clarify their purpose. The difference is that one is static whereas the other is an instance variable.

src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetFinder.java
fmeum: The `FuzzTargetRunner` is also used by the JUnit executor. That is the case where this is not `null`, but the other branch will go away when we use the JUnit launcher API. I added comments explaining that.

kyakdan: Thanks for the clarification. I've overseen the fact that control flows through the static block and the `newInstance` lambda.

kyakdan (approved):

src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetFinder.java
fmeum: Yes, it is confusing. I hope that we can get rid of most of this logic when we migrate to the JUnit launcher.

PR #612, opened 2023-01-25
Run `@FuzzTest` via JUnit in standalone Jazzer

fmeum:
The new runner uses the JUnit launcher API to run `@FuzzTest`s using JUnit, which provides support for Jupiter extensions even when running these tests from the command line.


src/main/java/com/code_intelligence/jazzer/driver/junit/JUnitRunner.java
```diff
@@ -0,0 +1,146 @@
...
+      }
+
+      @Override
+      public void reportingEntryPublished(TestIdentifier testIdentifier, ReportEntry entry) {
+        entry.getKeyValuePairs().values().forEach(Log::info);
+      }
+    });
+
+    TestExecutionResult result = resultHolder.get();
+    if (result.getStatus() != TestExecutionResult.Status.FAILED) {
```
bertschneider, line 129: Should we handle `ABORTED` separately and return a dedicated status code?

src/main/java/com/code_intelligence/jazzer/driver/junit/JUnitRunner.java
```diff
@@ -0,0 +1,146 @@
...
+      return 0;
+    }
+
+    Throwable throwable = result.getThrowable().get();
```
bertschneider, line 133: Do all failed tests have a `Throwable`? Otherwise `get` could throw an exception.

src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetRunner.java
```diff
@@ -126,6 +119,8 @@ public final class FuzzTargetRunner {
     fuzzerTearDown = fuzzTarget.tearDown.orElse(null);
     reproducerTemplate = new ReproducerTemplate(fuzzTargetClass.getName(), useFuzzedDataProvider);
 
+    JazzerInternal.onFuzzTargetReady(targetClassName);
```
bertschneider: Is this functionality used / useful anymore?

bertschneider (approved):
Great addition. Just some small remarks.

src/main/java/com/code_intelligence/jazzer/driver/junit/JUnitRunner.java
fmeum: They do, added a comment with a reference and restructured the logic to make this more clear.

src/main/java/com/code_intelligence/jazzer/driver/junit/JUnitRunner.java
fmeum: I don't think we should return a non-zero exit code here as a JUnit test run would pass in this case, but I added a warning.

src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetRunner.java
fmeum: It's not used by Jazzer itself, but I know of third-party hooks that do rely on it (e.g. static analysis developed as part of AI-DevAssist).

kyakdan (approved):

src/main/java/com/code_intelligence/jazzer/driver/junit/JUnitRunner.java
fmeum: Had to tweak this logic once more.

src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetRunner.java
```diff
@@ -64,9 +65,6 @@ public final class FuzzTargetRunner {
 
   private static final long BYTE_ARRAY_OFFSET = UNSAFE.arrayBaseOffset(byte[].class);
 
-  // Default value of the libFuzzer -error_exitcode flag.
-  private static final int LIBFUZZER_ERROR_EXIT_CODE = 77;
-
   // Possible return values for the libFuzzer callback runOne.
   private static final int LIBFUZZER_CONTINUE = 0;
   private static final int LIBFUZZER_RETURN_FROM_DRIVER = -2;
```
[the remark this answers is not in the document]
fmeum: I only extracted this one because it's relevant outside `FuzzTargetRunner` and needed elsewhere. If the other constants aren't needed elsewhere, I would prefer to keep them private to this class.

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/RegexRoadblocks.java
```diff
@@ -65,13 +65,6 @@ public final class RegexRoadblocks {
   private static final ThreadLocal<WeakHashMap<Object, Character>> PREDICATE_SOLUTIONS =
       ThreadLocal.withInitial(WeakHashMap::new);
 
-  // Do not act on instrumented regexes used by Jazzer internally, e.g. by ClassGraph.
-  private static boolean HOOK_DISABLED = true;
-
-  static {
-    Jazzer.onFuzzTargetReady(() -> HOOK_DISABLED = UNSAFE == null);
```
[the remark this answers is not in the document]
fmeum: It's part of our public API and very low cost to keep around, so I wouldn't want to delete it just because it's not used by us anymore. What do you think?

src/main/java/com/code_intelligence/jazzer/driver/junit/JUnitRunner.java
bertschneider: Looks good.

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/RegexRoadblocks.java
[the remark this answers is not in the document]
fmeum: Let's defer to a follow-up if needed since everything else is resolved.

PR #618, opened 2023-02-07
agent: Fix Kotlin compiler warnings about `@OptIn`

fmeum:
Some `@OptIn`s aren't needed anymore, the other can be replaced with dedicated annotations that do not require special compiler flags.


bertschneider (approved):
Getting rid of the `OptIn`s :+1:

PR #620, opened 2023-02-09
mutation: Add support for embedded message fields

fmeum:


src/main/java/com/code_intelligence/jazzer/mutation/api/InPlaceMutator.java
```diff
@@ -31,13 +31,13 @@
...
- *   <li>SHOULD override {@link Object#toString()} to return a stable but otherwise unspecified
- *   debug representation of their subtree of mutators.
+ *   <li>SHOULD override {@link Object#toString()} to return {@code
+ * Debuggable.getDebugString(this)}.
  * </ul>
  *
  * @param <T> the reference type this mutator operates on
  */
-public interface InPlaceMutator<T> {
+public interface InPlaceMutator<T> extends Debuggable {
```
bertschneider, line 40: Provide an abstract `InPlaceMutator` implementing `toString()` in the described way?

src/main/java/com/code_intelligence/jazzer/mutation/combinator/MutatorCombinators.java
```diff
@@ -127,4 +176,73 @@ public R detach(R value) {
...
+    private final Supplier<T> makeDefaultInstance;
+    private final Serializer<T> serializer;
+    private final InPlaceMutator<T> mutator;
+
+    private DelegatingSerializingInPlaceMutator(Consumer<SerializingInPlaceMutator<T>> registerSelf,
+        Supplier<T> makeDefaultInstance, Serializer<T> serializer,
+        Supplier<InPlaceMutator<T>> lazyMutator) {
+      requireNonNull(makeDefaultInstance);
+      requireNonNull(serializer);
+      requireNonNull(makeDefaultInstance);
```
bertschneider, line 190: Checks duplicated

bertschneider (approved):

src/main/java/com/code_intelligence/jazzer/mutation/api/InPlaceMutator.java
fmeum: That's a good point. I haven't done it yet for the following reason: `SerializingMutator` implements `toString` in this way and is the only mutator that can appear on the top level (that is, be created by a `MutatorFactory`), everything that doesn't inherits from it will probably never have `toString` called on itself outside of very specific tests in which a failure to implement this method would be noticed immediately.

Ideally we would make `InPlaceMutator` an abstract class, but that clashes with `SerializingMutator`. We could provide an additional `AbstractInPlaceMutator`, but following the reasoning above, I would expect that to cause more confusion than it resolves. What do you think?

bertschneider: Thanks for the clarification. Convincing reason. Also complicating the class hierarchy for such a small gain is debatable to begin with. Let's keep it as is.

bertschneider (approved):

PR #621, opened 2023-02-10
Ignore offline instrumentation

bertschneider:
JaCoCo is able to offline instrument classes. It determines if classes,
which should be instrumented dynamically, are already instrumented by
checking the existence of the coverage data field, and throws an
exception if it's present.

Jazzer's `IProbeArrayStrategy` already does not add the coverage data
field. This PR also removes JaCoCo's check if the field is present, and
so enables fuzzing of offline instrumented classes.

### DIFF
--- a/tests/BUILD.bazel
+++ b/tests/BUILD.bazel
@@ -389,3 +389,38 @@ java_fuzz_target_test(
     allowed_findings = ["com.code_intelligence.jazzer.api.FuzzerSecurityIssueHigh"],
     target_class = "com.example.SilencedFuzzer",
 )
+
+java_binary(
+    name = "jacococli",
+    main_class = "org.jacoco.cli.internal.Main",
+    runtime_deps = ["@jacococli//file:jacococli.jar"],
+)
+
+java_library(
+    name = "OfflineInstrumentedTarget",
+    srcs = ["src/test/java/com/example/OfflineInstrumentedTarget.java"],
+)
+
+genrule(
+    name = "OfflineInstrumentedTargetInstrumented",
+    srcs = [":OfflineInstrumentedTarget"],
+    outs = ["OfflineInstrumentedTargetInstrumented.jar"],
+    cmd = """
+$(location :jacococli) instrument $< --dest jacoco-instrumented
+cp jacoco-instrumented/*.jar $@
+""",
+    tags = ["manual"],
+    tools = [":jacococli"],
+)
+
+java_fuzz_target_test(
+    name = "OfflineInstrumentedFuzzer",
+    timeout = "short",
+    srcs = ["src/test/java/com/example/OfflineInstrumentedFuzzer.java"],
+    allowed_findings = ["java.lang.IllegalStateException"],
+    target_class = "com.example.OfflineInstrumentedFuzzer",
+    deps = [
+        ":OfflineInstrumentedTargetInstrumented",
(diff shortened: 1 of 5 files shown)


tests/BUILD.bazel
```diff
@@ -389,3 +389,49 @@ java_fuzz_target_test(
...
+
+# Copy instrumented jar to expected output destination
+cp $$PWD/$$(basename "$<") $@
+""",
+    target_compatible_with = LINUX_ONLY,
+    toolchains = [
+        "@bazel_tools//tools/jdk:current_java_runtime",
+    ],
+    tools = [
+        "@jacococli//file:jacococli.jar",
```
fmeum, line 421: If you wrap this jar in a `java_binary` target and add that to `tools`, the genrule should become much simpler as you won't need to get and use the  Java toolchain anymore.

tests/BUILD.bazel
```diff
@@ -389,3 +389,49 @@ java_fuzz_target_test(
...
+    visibility = ["//visibility:public"],
+)
+
+java_fuzz_target_test(
+    name = "OfflineInstrumentedFuzzer",
+    timeout = "short",
+    srcs = ["src/test/java/com/example/OfflineInstrumentedFuzzer.java"],
+    allowed_findings = ["java.lang.IllegalStateException"],
+    target_class = "com.example.OfflineInstrumentedFuzzer",
+    target_compatible_with = LINUX_ONLY,
```
fmeum, line 432: Just curious, why does this work on Linux only?

tests/BUILD.bazel
bertschneider: I'm not sure how to do that. `java_binary` requires `srcs` or `main_class` which we don't have here, or do we? Simplifying the `genrule` would be great, though.

tests/BUILD.bazel
bertschneider: To be honest I haven't checked in details. The pipeline failed and I assumed the scipt just couldn't be executed on other systems. But it should work on macOS, shouldn't it? I will check that in detail again.

tests/BUILD.bazel
fmeum: I pushed a commit that implements this, please take a look and feel free to make changes.

fmeum (approved):

tests/BUILD.bazel
bertschneider: Much better, thx! With your change the script also works on all platforms.

tests/BUILD.bazel
```diff
@@ -389,3 +389,39 @@ java_fuzz_target_test(
...
+    tools = [":jacococli"],
+)
+
+java_fuzz_target_test(
+    name = "OfflineInstrumentedFuzzer",
+    timeout = "short",
+    srcs = ["src/test/java/com/example/OfflineInstrumentedFuzzer.java"],
+    allowed_findings = ["java.lang.IllegalStateException"],
+    target_class = "com.example.OfflineInstrumentedFuzzer",
+    target_compatible_with = LINUX_ONLY,
```
fmeum, line 422: Do we still need this then?

```suggestion
```

bertschneider: Ah, my bad. I thought the restriction was removed at both targets.

bertschneider: Removed, tests are still successful.

PR #622, opened 2023-02-10
Add Spring controller fuzz test example

oetr:


.github/workflows/run-all-tests.yml
```diff
@@ -10,6 +10,22 @@ on:
...
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up JDK
+        uses: actions/setup-java@v3
+        with:
+          distribution: zulu
+          java-version: 11
+      - name: Build and run tests
+        # cd to the project directory and run the shell script
```
fmeum, line 23: This comment seems a bit redundant. We could mention that the project is built and tested with Maven instead, which explains why this is a separate job.

examples/junit-spring-web/src/test/java/com/example/JunitSpringWebApplicationTests.java
```diff
@@ -0,0 +1,75 @@
...
+
+@WebMvcTest()
+public class JunitSpringWebApplicationTests {
+  @Autowired private MockMvc mockMvc;
+
+  private boolean beforeCalled = false;
+
+  @BeforeEach
+  public void beforeEach() {
+    this.beforeCalled = true;
```
fmeum, line 38: You can get rid of the explicit `this.` in all methods.

fmeum (approved):

examples/junit-spring-web/src/test/java/com/example/JunitSpringWebApplicationTests.java
oetr: done

.github/workflows/run-all-tests.yml
oetr: Done!

PR #623, opened 2023-02-12
mutation: Add explicit support for oneof message fields

fmeum:
While `oneof`s were handled by the existing code, the mutator wasn't aware of the `oneof`'s state and would thus change it frequently, which ends up invalidating mutations applied to its other fields.

With this commit, `oneof` fields have their current state tracked and mutated in 1 out of 100 cases, which allows individual fields to reach more complicated states.


bertschneider (approved):
This nicely shows how easily the framework can be extended.

PR #624, opened 2023-02-12
all: Use google-java-format for Java formatting

fmeum:


bertschneider (approved):
I didn't look at all the proposed changes, but the ones I saw mostly looked fine / better than before to me. Some unexpected new lines in `String` literals, nothing serious. IMHO we can go ahead with this.

PR #629, opened 2023-02-14
mutation: Add mutators for integral types and message fields

fmeum:


src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/IntegralMutatorFactory.java
```diff
@@ -0,0 +1,279 @@
...
+        largestMutableBitPositive = 0;
+      } else /* minValue < 0 && maxValue >= 0 */ {
+        largestMutableBitNegative = bitWidth(~minValue);
+        largestMutableBitPositive = bitWidth(maxValue);
+      }
+      this.specialValues = collectSpecialValues(minValue, maxValue);
+    }
+
+    private static long[] collectSpecialValues(long minValue, long maxValue) {
+      return LongStream.of(0, 1, minValue, maxValue)
```
bertschneider, line 196: Why is all the `LongStream` code needed? This should return an array with 4 elements, shouldn't it?

src/main/java/com/code_intelligence/jazzer/mutation/api/PseudoRandom.java
```diff
@@ -17,11 +17,57 @@
...
+  /**
+   * @return a {@code boolean} that is {@code true} with probability {@code 1/inverseFrequencyTrue}
+   */
+  boolean trueInOneOutOf(int inverseFrequencyTrue);
+
+  /**
+   * @throws IllegalArgumentException if {@code array.length == 0}
+   * @return a uniformly random index valid for the given array
+   */
+  <T> int index(T[] array);
```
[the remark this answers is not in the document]
fmeum: We should probably have a consistent style, so I would prefer not to add a generic prefix here.

I renamed the function to `indexIn`, what do you think?

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/IntegralMutatorFactory.java
```diff
@@ -0,0 +1,279 @@
...
+      // InRange is not repeatable, so the loop body will apply exactly once.
+      for (Annotation annotation : type.getAnnotations()) {
+        if (annotation instanceof InRange) {
+          InRange inRange = (InRange) annotation;
+          if (inRange.min() != Long.MIN_VALUE) {
+            require(inRange.min() >= defaultMinValueForType,
+                format("@InRange.min=%d is out of range: %s", inRange.min(), type.getType()));
+            minValue = inRange.min();
+          }
+          if (inRange.max() != Long.MIN_VALUE) {
```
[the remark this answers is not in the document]
fmeum: Good catch, that's a bug.

I added a comment to explain why we want the checks.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/IntegralMutatorFactory.java
fmeum: The stream chain has a `distinct`: On an `@InRange(min = 0) int`, we don't want to generate `0` twice as often as `1`.

fmeum:
Rebased onto main.

fmeum:
I pushed two small fixup commits: With the bug fix to the stress test implementation, the TODOs about surprisingly many duplicates can now be resolved and the tests strengthened.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/IntegralMutatorFactory.java
bertschneider: Yes, good point.

bertschneider (approved):

PR #631, opened 2023-02-15
Experimentally integrate the new mutator into Jazzer

fmeum:
The mutator can be enabled via the `--experimental_mutator` flag and, if disabled, should not add any measurable overhead.

The integration is highly unoptimized with many redundant copies. As follow-up work, this should be properly benchmarked and improved where necessary.


fmeum:
@bertschneider @bertschneider This is ready now. I had to polyfill a function for Java 8 compatibility and fixed more issues in `read` functions, but the example test is now passing very quickly (< 3s on my laptop) and also showcases our annotations.

bertschneider (approved):
Really nice integration. As you also mentioned, let's get everything working and only optimize afterwards.

PR #632, opened 2023-02-15
mutate: Add enum mutator

oetr:


fmeum:
@oetr You can set your base branch to one of my branches so that the diff becomes more manageable.

src/main/java/com/code_intelligence/jazzer/mutation/api/SerializingMutator.java
```diff
@@ -29,7 +29,7 @@
 @DoNotMock("Use TestSupport#mockMutator instead")
 public abstract class SerializingMutator<T> implements Serializer<T>, ValueMutator<T> {
   @Override
-  public final String toString() {
+  public String toString() {
```
fmeum, line 32: This function is `final` since subclasses shouldn't override it. Instead, they should (and have to) override `getDebugString`.

src/main/java/com/code_intelligence/jazzer/mutation/combinator/MutatorCombinators.java
```diff
@@ -174,6 +174,57 @@ public R detach(R value) {
     };
   }
 
+  public static <T, R> SerializingMutator<R> mutateThenMapToImmutable(
```
fmeum, line 177: Why is this function needed? The overload that already exists should emit both the `toString()` of the mutator it uses under the hood as well as the target type.

src/main/java/com/code_intelligence/jazzer/mutation/combinator/MutatorCombinators.java
```diff
@@ -174,6 +174,57 @@ public R detach(R value) {
...
+      }
+
+      @Override
+      public String toString() {
+        return name;
+      }
+    };
+  }
+
+  public static SerializingMutator<Integer> mutateIntegralInRange(
```
fmeum, line 192: Since this is only an internal mutator that we use to construct other ones, we could simplify it by restricting to ranges of the form `[0, upperExclusive)`, which fits zero-based indices very well.

If you do that, you also don't need `PseudoRandom#otherInClosedRange`, but just an additional overload for `PseudoRandom#otherIndex`.

src/main/java/com/code_intelligence/jazzer/mutation/engine/SeededPseudoRandom.java
```diff
@@ -111,4 +111,17 @@ public long closedRange(long lowerInclusive, long upperInclusive) {
       return r;
     }
   }
+
+  @Override
+  public int otherInClosedRange(int lowerInclusive, int upperInclusive, int currentValue) {
```
fmeum, line 116: See above, could be simplified to `otherIndex(int length, int currentIndex)`.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/EnumMutatorFactory.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.Optional;
+
+final class EnumMutatorFactory extends MutatorFactory {
+  @Override
+  public Optional<SerializingMutator<?>> tryCreate(AnnotatedType type, MutatorFactory factory) {
+    return findFirstParentIfClass(type, Enum.class).map(parent -> {
+      Enum[] values = ((Class<Enum>) type.getType()).getEnumConstants();
+      // TODO: when 1 or 0 values, return a constant mutator?
+      if (values.length == 0) {
+        return null;
```
fmeum, line 35: The equivalent of `null` for `Optional<T>` is `Optional.empty()`. While `null` is a valid return value, it shouldn't be used together with `Optional`s.

fmeum (commented):
This looks pretty good and functional already, I just left some comments that may help to simplify it further.

fmeum:
All my PRs have been merged, you can just rebase onto main.

src/main/java/com/code_intelligence/jazzer/mutation/combinator/MutatorCombinators.java
oetr: The idea was to ultimately pass the concrete enum type. We might still come back to this if we cannot get it in any other way!

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/EnumMutatorFactory.java
oetr: Do we need Optional.empty()  here? At the moment we require at least 2 elements in the enum.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/EnumMutatorFactory.java
```diff
@@ -0,0 +1,41 @@
...
+      }
+      return mutateThenMapToImmutable(mutateIntegralInRange(0, values.length - 1),
+          (index) -> values[index], Enum::ordinal, "Enum");
```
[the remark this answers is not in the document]
oetr: At the moment it's not working (it prints ´Enum´ and not `EnumTest` in the tests).

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/EnumMutatorFactory.java
```diff
@@ -0,0 +1,38 @@
...
+final class EnumMutatorFactory extends MutatorFactory {
+  @Override
+  public Optional<SerializingMutator<?>> tryCreate(AnnotatedType type, MutatorFactory factory) {
+    return findFirstParentIfClass(type, Enum.class).map(parent -> {
+      Enum<?>[] values = ((Class<Enum<?>>) type.getType()).getEnumConstants();
+      return mutateThenMapToImmutable(
+          mutateIndices(values.length), (index) -> values[index], Enum::ordinal);
+    });
+  }
+}
```
fmeum, line 38: nit: Add the missing trailing newline

src/main/java/com/code_intelligence/jazzer/mutation/engine/SeededPseudoRandom.java
```diff
@@ -73,6 +73,16 @@ public <T> int otherIndex(T[] array, int currentIndex) {
     }
   }
 
+  @Override
+  public int otherIndex(int length, int currentIndex) {
```
fmeum, line 77: I missed this during the merges, but @bertschneider suggested renaming `index` to `indexIn`. For the sake of consistency and clarity, we could also rename `otherIndex` (both overloads) to `otherIndexIn` and rename the first parameter to `range`.

src/main/java/com/code_intelligence/jazzer/mutation/engine/SeededPseudoRandom.java
```diff
@@ -73,6 +73,16 @@ public <T> int otherIndex(T[] array, int currentIndex) {
...
+    int otherIndex = currentIndex + closedRange(1, length - 1);
```
fmeum, line 78: The other overload can just call this method.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/EnumMutatorFactory.java
fmeum: As far as I can tell we would now reach the `require` in `mutateIndices` if an `Enum` has less than two elements. We could have a more user-facing error message by requiring right in the `tryCreate` function with a message such as "%s defines less than two enum constants and can't be mutated, use a constant instead".

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/EnumMutatorFactory.java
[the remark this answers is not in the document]
fmeum: True, it uses the declared return type of the lambda, not the actual class of the returned value. The latter could be achieved by constructing an element with something like `readExclusive(new ByteArrayInputStream(new byte[0]))` and then calling `getClass` on that.

@bertschneider What do you think, should we mention concrete types in the debug string? My initial thought was not to do this so that debug strings are certain not to contain user information (e.g. concrete class names), but of course that additional information could still be useful.

src/test/java/com/code_intelligence/jazzer/mutation/mutator/lang/EnumMutatorTest.java
```diff
@@ -0,0 +1,63 @@
...
+      cl = mutator.mutate(cl, prng);
+    }
+    assertThat(cl).isEqualTo(TestEnum.C);
+
+    try (MockPseudoRandom prng = mockPseudoRandom(1)) {
+      cl = mutator.mutate(cl, prng);
+    }
+    assertThat(cl).isEqualTo(TestEnum.B);
+  }
+}
```
fmeum, line 63: nit: Missing trailing newline

@bertschneider I have a PR in draft that would switch us over to google-java-format, which among other things also adds the newlines.

oetr: - Added `require` in `tryCreate`
- Overloaded `mutateThenMap` with a lambda that will override `toDebugString` when needed
- I expected format.sh to take care of the nit. Added!

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/EnumMutatorFactory.java
oetr: Done!

src/main/java/com/code_intelligence/jazzer/mutation/engine/SeededPseudoRandom.java
oetr: Done!

src/main/java/com/code_intelligence/jazzer/mutation/engine/SeededPseudoRandom.java
oetr: Done!

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/EnumMutatorFactory.java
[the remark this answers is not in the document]
bertschneider: IMHO concrete type names, even user defined ones, are fine, as they are not published anywhere and can help greatly during debugging.

src/main/java/com/code_intelligence/jazzer/mutation/combinator/MutatorCombinators.java
```diff
@@ -179,6 +189,57 @@ public R detach(R value) {
...
+      }
+
+      @Override
+      public String toDebugString(Predicate<Debuggable> isInCycle) {
+        return debug.apply(isInCycle);
+      }
+    };
+  }
+
+  public static SerializingMutator<Integer> mutateIndices(int length) {
```
bertschneider, line 208: `mutateIndices` basically returns a bounded `IntegralMutator`, doesn't it? I wonder if we could reuse that, or the integral mutation part, here as well.

src/test/java/com/code_intelligence/jazzer/mutation/mutator/lang/EnumMutatorTest.java
```diff
@@ -0,0 +1,76 @@
...
+  enum TestEnumOne { A }
+
+  enum TestEnum { A, B, C }
+
+  @Test
+  void testBoxed() {
+    SerializingMutator<TestEnum> mutator =
+        (SerializingMutator<TestEnum>) LangMutators.newFactory().createOrThrow(
+            new TypeHolder<@NotNull TestEnum>() {}.annotatedType());
+    assertThat(mutator.toString()).isEqualTo("TestEnum");
```
bertschneider, line 40: It could be helpful to explicitly state that this is an `Enum` mutator, as the class name could be misleading. How do you like `Enum<TestEnum>`?

src/main/java/com/code_intelligence/jazzer/mutation/combinator/MutatorCombinators.java
oetr: I was thinking the same when I saw the PR that introduced IntegralMutator! However, after looking at `bitFlip` and `randomWalk` I am not sure how useful that is for enums. Should keep this in mind until we have a benchmark ready to test the effectivenes of each mutator?

src/test/java/com/code_intelligence/jazzer/mutation/mutator/lang/EnumMutatorTest.java
oetr: It's much better, I added it, thanks :smile:

src/main/java/com/code_intelligence/jazzer/mutation/combinator/MutatorCombinators.java
bertschneider: Sounds good.

bertschneider (approved):

PR #635, opened 2023-02-17
Preparations for experimental mutator integration

fmeum:


src/main/java/com/code_intelligence/jazzer/mutation/support/InputStreamSupport.java
```diff
@@ -51,8 +51,11 @@ public static byte[] readAllBytes(InputStream stream) throws IOException {
     byte[] result = new byte[arrayLength];
     int offset = 0;
     byte[] buffer;
+    int remaining = arrayLength;
```
bertschneider, line 54: Why is this necessary? I expected sum of `buffer.length` to be `arrayLength`.

bertschneider (approved):
Nice cleanup!

src/main/java/com/code_intelligence/jazzer/mutation/support/InputStreamSupport.java
fmeum: That's not necessarily true for the last buffer, which is exactly what caused the bug. I feel a bit better now about missing this at first :-D

PR #643, opened 2023-02-28
Add SSRF detector

bertschneider:
Server Side Request Forgery (SSRF) attacks occur when requests to destinations based on untrusted data are executed. This could lead to exfiltration of sensitive data or exposure of internal services.

The sanitizer detects such attacks by checking if the fuzzer was able to inject a honeypot host name into the requested destination. If this is the case, it reports a high severity security issue.

Checks are implemented via `java.net.Socket.connect` and `java.nio.channels.SocketChannel.connect` methods.


sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/SsrfInjection.kt
```diff
@@ -0,0 +1,157 @@
...
+    // - java.net.AbstractPlainSocketImpl (no direct usage)
+    // - java.net.PlainSocketImpl (no direct usage)
+    // - java.net.HttpConnectSocketImpl (only used in Socket)
+    // - java.net.SocksSocketImpl (used in Socket, but also invoking super.connect directly, hook required)
+    // - java.net.ServerSocket (security check, no hook required)
+    @MethodHook(
+        type = HookType.BEFORE,
+        targetClassName = "java.net.SocketImpl",
+        targetMethod = "connect",
+        additionalClassesToHook = [
```
0xricksanchez, line 94: I find the comment with the `additionalClassesToHook` difficult to understand at first glance. Why do we need to hook the 4 additional classes, from which only 2 appear in the above comment? What about those that are explicitly mentioned in the comment?

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/SsrfInjection.kt
```diff
@@ -0,0 +1,157 @@
...
+        checkSsrf(arguments, hookId)
+    }
+
+    @JvmStatic
+    fun checkSsrf(arguments: Array<Any>, hookId: Int) {
+        if (arguments.isEmpty()) {
+            return
+        }
+
+        val host = when (val arg = arguments[0]) {
```
0xricksanchez, line 112: Couldn't we collapse this block into:

```java
val host = when (val arg = arguments[0]) {
    is String -> try {
        URI.create(arg).host
    } catch (e: Exception) {
        null
    }
    is InetAddress, is InetSocketAddress -> arg.hostName
    else -> null
} ?:return
```

The try-catch block can be expensive if an invalid URL is provided. Instead of catching the exception, we can use the `URI.create` which AFAIK should throw an exception if the provided URL is invalid.

Also, combining `InetAddress` and `InetSocketAddress` seems reasonable

0xricksanchez (commented):
Looks reasonable to me. The example fuzzer is quite slow, and I was able to get one slow output (as mentioned in the sanitizer comments), which I'm not sure is a good finding here..

```
Slowest unit: 75 s:
artifact_prefix='/private/var/tmp/_bazel_christopher/8c60c24fbbc15a396e9df7c56a2aae10/execroot/jazzer/bazel-out/darwin_arm64-opt/testlogs/examples/BatikTranscoderFuzzer/test.outputs/'; Test unit written to /private/var/tmp/_bazel_christopher/8c60c24fbbc15a396e9df7c56a2aae10/execroot/jazzer/bazel-out/darwin_arm64-opt/testlogs/examples/BatikTranscoderFuzzer/test.outputs/slow-unit-17f8ba6eb1d560a71023696c46f7d388e9d7405e
Base64: ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODg4ODHzMQ==
```

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/SsrfInjection.kt
bertschneider: I tried to explain the reason to hook those classes in more detail. The classes that were not mentioned are not required, though. Thx for noticing that!

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/SsrfInjection.kt
bertschneider: Can you elaborate how switching from `URL` to `URI` would improve the performance? The try-catch-block would still be needed, wouldn't it?

Indeed, combining `InetAddress` and `InetSocketAddress` would be nice, but the compiler can not derive a common type, beside `Object`, and so `hostName` wouldn't be accessible anymore.

0xricksanchez: Ah that makes sense. 

Regarding the first point: I was sure `URI.create()` allows for unchecked exceptions which would be cheaper (as we will run into this exception a lot) and would allow us to get rid of the try-catch block. Upon further inspection of the OpenJDK code I was wrong. The only difference here is that using URI could cover different aspects, as every URL is an URI, but not vice versa. In the end as we're specifically testing for URLs here using `URL()` should be fine.

`URI` uses [parse](https://github.com/openjdk/jdk/blob/52d30087734ad95761078793da6e207797558e2b/src/java.base/share/classes/java/net/URI.java#L3152)
`URL` ends up [here](https://github.com/openjdk/jdk/blob/52d30087734ad95761078793da6e207797558e2b/src/java.base/share/classes/java/net/URL.java#L490) 

If we really  wanted, we could check which one is faster to fail, as we'd be triggering the bad path over and over again..

0xricksanchez (approved):
Please don't hesitate to resolve as is

fmeum:
@bertschneider I updated the PR, could you take a look? 

I am thinking that we should maybe reduce the severity to Medium. What do you think?

src/main/java/com/code_intelligence/jazzer/api/BugDetectors.java
```diff
@@ -0,0 +1,111 @@
...
+
+  private static AtomicReference<BiPredicate<String, Integer>> getConnectionPermittedReference() {
+    try {
+      Class<?> ssrfSanitizer =
+          Class.forName("com.code_intelligence.jazzer.sanitizers.ServerSideRequestForgery");
+      return (AtomicReference<BiPredicate<String, Integer>>) ssrfSanitizer
+          .getField("connectionPermitted")
+          .get(null);
+    } catch (ClassNotFoundException | NoSuchFieldException | IllegalAccessException e) {
+      System.err.println("WARNING: ");
```
bertschneider, line 104: Use `Log.warn`?

bertschneider:
I really like this approach, especially the configuration via the scopes.
We should also add some kind of documentation, but can combine that with overall sanitizer docs in different PR.

src/main/java/com/code_intelligence/jazzer/api/BugDetectors.java
fmeum: We can't in sanitizers as `Log` is not (and probably shouldn't be) in the API package.

fmeum:
@bertschneider Do you have an idea why the HTTPClient test isn't skipped in CI? Am I using the tag incorrectly?

src/main/java/com/code_intelligence/jazzer/api/BugDetectors.java
bertschneider: Ah, yes, right!

PR #653, opened 2023-03-07
mutator: Add map mutator

bertschneider:
Support to mutate maps.

### DIFF
--- a/src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
+++ b/src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
@@ -1,0 +1,274 @@
+/*
+ * Copyright 2023 Code Intelligence GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.code_intelligence.jazzer.mutation.mutator.collection;
+
+import static com.code_intelligence.jazzer.mutation.support.Preconditions.check;
+import static com.code_intelligence.jazzer.mutation.support.Preconditions.require;
+import static com.code_intelligence.jazzer.mutation.support.TypeSupport.parameterTypesIfParameterized;
+import static java.lang.Math.min;
+import static java.lang.String.format;
+import static java.util.stream.Collectors.toMap;
+
+import com.code_intelligence.jazzer.mutation.annotation.WithSize;
+import com.code_intelligence.jazzer.mutation.api.*;
+import com.code_intelligence.jazzer.mutation.support.StreamSupport;
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.AnnotatedType;
+import java.util.AbstractMap;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+final class MapMutatorFactory extends MutatorFactory {
+  @Override
+  public Optional<SerializingMutator<?>> tryCreate(AnnotatedType type, MutatorFactory factory) {
+    return parameterTypesIfParameterized(type, Map.class)
+        .map(parameterTypes
+            -> parameterTypes.stream()
+                   .map(factory::tryCreate)
+                   .flatMap(StreamSupport::getOrEmpty)
+                   .collect(Collectors.toList()))
+        .map(elementMutators -> {
+          check(elementMutators.size() == 2);
+          int min = MapMutator.DEFAULT_MIN_SIZE;
+          int max = MapMutator.DEFAULT_MAX_SIZE;
+          for (Annotation annotation : type.getDeclaredAnnotations()) {
+            if (annotation instanceof WithSize) {
+              WithSize withSize = (WithSize) annotation;
+              min = withSize.min();
+              max = withSize.max();
+            }
+          }
+          return new MapMutator<>(elementMutators.get(0), elementMutators.get(1), min, max);
+        });
+  }
+
+  private static final class MapMutator<K, V> extends SerializingInPlaceMutator<Map<K, V>> {
+    private static final int DEFAULT_ATTEMPTS_COUNT = 100;
+    private static final int DEFAULT_MIN_SIZE = 0;
+    private static final int DEFAULT_MAX_SIZE = 1000;
+
+    private final SerializingMutator<K> keyMutator;
+    private final SerializingMutator<V> valueMutator;
+    private final int minSize;
+    private final int maxSize;
+
+    MapMutator(SerializingMutator<K> keyMutator, SerializingMutator<V> valueMutator, int minSize,
+        int maxSize) {
+      this.keyMutator = keyMutator;
+      this.valueMutator = valueMutator;
+      this.minSize = Math.max(minSize, DEFAULT_MIN_SIZE);
+      this.maxSize = Math.min(maxSize, DEFAULT_MAX_SIZE);
+
+      require(maxSize >= 1, "WithSize#max needs to be greater than 0");
+      require(minSize >= 0, "WithSize#min size needs to be greater or equal 0");
+      require(minSize <= maxSize,
+          format("WithSize#min %d needs to be smaller or equal than WithSize#max %d", minSize,
+              maxSize));
+    }
+
+    @Override
+    public Map<K, V> read(DataInputStream in) throws IOException {
+      int size = Math.min(Math.max(in.readInt(), minSize), maxSize);
+      Map<K, V> map = new HashMap<>(size);
+      for (int i = 0; i < size; i++) {
+        map.put(keyMutator.read(in), valueMutator.read(in));
+      }
+      // Wrap in an immutable view for additional protection against accidental mutation in fuzz
+      // tests.
+      return toImmutableMapView(map);
+    }
+
+    @Override
+    public void write(Map<K, V> map, DataOutputStream out) throws IOException {
+      out.writeInt(map.size());
+      for (Map.Entry<K, V> entry : map.entrySet()) {
+        keyMutator.write(entry.getKey(), out);
+        valueMutator.write(entry.getValue(), out);
+      }
+    }
+
+    @Override
+    protected Map<K, V> makeDefaultInstance() {
+      // Wrap in an immutable view for additional protection against accidental mutation in fuzz
+      // tests.
+      return toImmutableMapView(new HashMap<>(maxInitialSize()));
+    }
+
+    @Override
+    public void initInPlace(Map<K, V> reference, PseudoRandom prng) {
+      int targetSize = prng.closedRange(minInitialSize(), maxInitialSize());
+      Map<K, V> map = underlyingMutableList(reference);
+      map.clear();
+      for (int i = 0; i < targetSize; i++) {
+        initElement(map, prng);
+      }
+    }
+
+    @Override
+    public void mutateInPlace(Map<K, V> reference, PseudoRandom prng) {
+      Map<K, V> map = underlyingMutableList(reference);
+      if (map.isEmpty()) {
+        initElement(map, prng);
+      } else if (!prng.trueInOneOutOf(4)) {
+        // Choose a random entry, and then key or value, to mutate.
+        int i = prng.indexIn(map.size());
+        map.entrySet().stream().skip(i).findFirst().ifPresent(
+            entry -> mutateElement(entry, map, prng));
+      } else {
+        // This will increase the map size up to max and then oscillate around it.
+        // Smaller sizes are already used, so increasing the size should be a valid choice.
+        // FIXME: Think about a better strategy to change the size.
+        int currentSize = map.size();
+        if (currentSize < maxSize) {
+          // Create a new entry, as a deep copy could be expensive.
+          initElement(map, prng);
+        } else if (currentSize > minSize) {
+          // Remove a random entry.
+          int i = prng.indexIn(currentSize);
+          map.keySet().stream().skip(i).findFirst().ifPresent(map::remove);
+        } else {
+          // One element map, mutate that one.
+          mutateElement(map.entrySet().iterator().next(), map, prng);
+        }
+      }
+    }
+
+    private void initElement(Map<K, V> map, PseudoRandom prng) {
+      // Lookup an unused key first, as the map may already contain and
+      // deduplicate it. Keys are solely based on the prng so that an
+      // unused one should eventually be found. In case all or most values
+      // of the given type are already used, give up after some attempts.
+      int attempts = 0;
+      K key;
+      do {
+        if (attempts++ > DEFAULT_ATTEMPTS_COUNT) {
+          // Give up, as no unused key could be found.
+          return;
+        }
+        key = keyMutator.init(prng);
+      } while (map.containsKey(key));
+      map.put(key, valueMutator.init(prng));
+    }
+
+    private void mutateElement(Map.Entry<K, V> entry, Map<K, V> map, PseudoRandom prng) {
+      K key = entry.getKey();
+      V value = entry.getValue();
+      boolean mutateKeyOrValue = prng.choice();
+
+      map.remove(key);
+
+      // Try to mutate the current key into an unused one.
+      // If that doesn't succeed after some attempts, mutate the value.
+      if (mutateKeyOrValue) {
+        int attempts = 0;
+        K mutated = key;
+        while (true) {
+          mutated = keyMutator.mutate(mutated, prng);
+          if (!mutated.equals(key) && !map.containsKey(mutated)) {
+            key = mutated;
+            break;
+          } else if (attempts++ > DEFAULT_ATTEMPTS_COUNT) {
+            // Give up, as no unused key could be found, and mutate the value instead.
+            mutateKeyOrValue = false;
+            break;
+          }
+        }
+      }
+
+      // The value mutator is responsible to produce different values, no need for a retry.
+      if (!mutateKeyOrValue) {
+        value = valueMutator.mutate(value, prng);
+      }
+      map.put(key, value);
+    }
+
+    @Override
+    public Map<K, V> detach(Map<K, V> value) {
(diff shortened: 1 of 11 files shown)


src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
```diff
@@ -0,0 +1,203 @@
...
+                   .collect(Collectors.toList()))
+        .filter(elementMutators -> elementMutators.size() == 2)
+        .map(elementMutators
+            -> new MapMutator<>(elementMutators.get(0), elementMutators.get(1),
+                MapMutator.DEFAULT_MIN_SIZE, MapMutator.DEFAULT_MAX_SIZE));
+  }
+
+  private static final class MapMutator<K, V> extends SerializingInPlaceMutator<Map<K, V>> {
+    private static final int DEFAULT_MIN_SIZE = 0;
+    private static final int DEFAULT_MAX_SIZE = 1000;
```
oetr, line 51: Can users increase this size without building their own Jazzer, if for some reason they need `> DEFAULT_MAX_SIZE` keys in their map?

bertschneider: Good point. I added support for the `@WithSize` annotation to set min/max freely.

oetr: An observation: By itself, the fuzzer cannot generate maps with more keys than `maxSize`.
However, at the moment `maxSize` is not checked when reading a map with more keys than `maxSize`.
That means that should our corpus contain inputs with larger maps from previous runs, even if we set the `maxSize` to some value, those inputs will currently be loaded by the mutator framework.
Depending on what the user wants, this might get confusing.

In libFuzzer, inputs larger than `-max_len` are usually ignored.

Maybe this is something we can address in the future.

fmeum: Good catch! We should address this now as changing the `read`/`write` methods is a breaking change. We could just use the min/max with the configured min/max sizes there. I think that e.g. the integer mutator is already doing that.

oetr: In that case `ListMutatorFactory` should be updated as well.

Is the strategy then to read all elements, but to keep only the `maxSize`; and fill up to 'minSize` if there are not enough elements? Or should such inputs be discarded altogether? In both cases only the `read` method should be changed.

The issue with the always read strategy is that reproducing a finding found with specific size settings is now tied to the input as well as to the configured sizes.
If the user changes the size configuration, a finding might stop being a finding until they change the sizes back.

During fuzzing, libFuzzer ignores inputs greater than `max_len` and won't generate inputs greater than that; but when executing the fuzzer with specific inputs, libFuzzer ignores the `max_len` setting and uses all available data.

I think, we should do the same---inputs not compliant with configured sizes should be ignored during fuzzing, but loaded during the reproducing phase. What do you think?

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
```diff
@@ -0,0 +1,215 @@
...
+        int maxSize) {
+      this.keyMutator = keyMutator;
+      this.valueMutator = valueMutator;
+      this.minSize = minSize;
+      this.maxSize = maxSize;
+    }
+
+    @Override
+    public Map<K, V> read(DataInputStream in) throws IOException {
+      int size = Math.max(in.readInt(), 0);
```
oetr, line 80: What if the user changes the code, and a large number that was interpreted as an integer in previous runs, suddenly becomes the size of a map/list? We aren't saving mutator IDs in the corpus, right?

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
```diff
@@ -0,0 +1,215 @@
...
+
+final class MapMutatorFactory extends MutatorFactory {
+  @Override
+  public Optional<SerializingMutator<?>> tryCreate(AnnotatedType type, MutatorFactory factory) {
+    // noinspection OptionalIsPresent
+    return parameterTypesIfParameterized(type, Map.class)
+        .map(parameterTypes
+            -> parameterTypes.stream()
+                   .map(factory::tryCreate)
+                   .flatMap(o -> o.isPresent() ? Stream.of(o.get()) : Stream.empty())
```
fmeum, line 43: This could be extracted into a polyfill for `Optional#stream`, e.g. `public static <T> Stream<T>(Optional<T> optional)` in `StreamSupport`.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
```diff
@@ -0,0 +1,215 @@
...
+          map.put(key, value);
+        });
+      } else {
+        // This will increase the map size up to max and then oscillate around it.
+        // Smaller sizes are already used, so increasing the size should be a valid choice.
+        if (map.size() < maxSize) {
+          // Create a new entry, as a deep copy could be expensive.
+          map.put(keyMutator.init(prng), valueMutator.init(prng));
+        } else if (map.size() > minSize) {
+          // Maps are not ordered, removing the first one is as good as any.
```
fmeum, line 142: Not sure that's true. I don't know how Java randomizes the map order exactly, but it's possible that it stays the same within one JVM process. Maybe we should again traverse a random number of elements before removing one?

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/StringMutatorFactory.java
```diff
@@ -142,14 +144,19 @@ public Optional<SerializingMutator<?>> tryCreate(AnnotatedType type, MutatorFact
...
+                if (fixUpAscii) {
+                  fixUpAscii(bytes);
+                } else {
+                  fixUpUtf8(bytes);
+                }
+                return new String(bytes, StandardCharsets.UTF_8);
+              },
+              string
+              -> string.getBytes(StandardCharsets.UTF_8),
+              (Predicate<Debuggable> inCycle) -> "String");
```
fmeum, line 159: I guess "better" is subjective here. I kind of liked the more verbose but also more consistent output better. I'm not against this change, but could you explain why you prefer it this way?

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
bertschneider: Done.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
bertschneider: Done.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/StringMutatorFactory.java
bertschneider: IMHO the underlying `byte` array is an implementation detail of `String` / the `String` mutator and somehow confused me more than it provided help during debugging. That being said, I'm not really attached to this change. If it may help debugging `String` mutator problems I can gladly revert it.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
```diff
@@ -0,0 +1,238 @@
...
+                   .map(factory::tryCreate)
+                   .flatMap(StreamSupport::getOrEmpty)
+                   .collect(Collectors.toList()))
+        .filter(elementMutators -> elementMutators.size() == 2)
+        .map(elementMutators -> {
+          int min = MapMutator.DEFAULT_MIN_SIZE;
+          int max = MapMutator.DEFAULT_MAX_SIZE;
+          Optional<WithSize> withSize = Arrays.stream(type.getAnnotations())
+                                            .filter(annotation -> annotation instanceof WithSize)
+                                            .findFirst()
```
fmeum, line 51: I think that a simple `for` loop over the annotations would be simpler here. `@WithSize` isn't repeatable and probably never will be, but having "last one wins semantics" would be correct even in that case whereas `findFirst` wouldn't.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
```diff
@@ -0,0 +1,238 @@
...
+    private final int maxSize;
+
+    MapMutator(SerializingMutator<K> keyMutator, SerializingMutator<V> valueMutator, int minSize,
+        int maxSize) {
+      this.keyMutator = keyMutator;
+      this.valueMutator = valueMutator;
+      this.minSize = Math.max(minSize, DEFAULT_MIN_SIZE);
+      this.maxSize = Math.min(maxSize, DEFAULT_MAX_SIZE);
+
+      require(maxSize >= 1, "Max. size needs to be greater than 0");
```
fmeum, line 77: Since this is potentially user-facing, maybe change to something like:
```suggestion
      require(maxSize >= 1, "WithSize#max needs to be greater than 0");
```

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
```diff
@@ -0,0 +1,238 @@
...
+      Map<K, V> map = underlyingMutableList(reference);
+      if (map.isEmpty()) {
+        initElement(map, prng);
+      } else if (!prng.trueInOneOutOf(4)) {
+        // Choose a random entry, and then key or value, to mutate.
+        int i = prng.indexIn(map.size());
+        map.entrySet().stream().skip(i).findFirst().ifPresent(
+            entry -> mutateElement(entry, map, prng));
+      } else {
+        // This will increase the map size up to max and then oscillate around it.
```
fmeum, line 132: This oscillation behavior is a bit weird. I know we have it for the list mutator, but we probably want to get rid of it sooner rather than later. Could you turn this into a FIXME and create a follow-up ticket?

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
```diff
@@ -0,0 +1,238 @@
...
+      }
+    }
+
+    private void initElement(Map<K, V> map, PseudoRandom prng) {
+      // Lookup an unused key first, as the map may already contain it.
+      // Keys are solely based on the prng and should eventually be unique.
+      K key;
+      do {
+        key = keyMutator.init(prng);
+      } while (map.containsKey(key));
```
fmeum, line 155: This is potentially an infinite loop if e.g. the key type is `Boolean`. We should limit the maximum number of attempts. Did you check what fuzztest does in this case?

src/main/java/com/code_intelligence/jazzer/mutation/support/TypeSupport.java
```diff
@@ -360,16 +373,10 @@ public static Optional<AnnotatedType> parameterTypeIfParameterized(
...
-    }
-    AnnotatedType elementType = typeArguments[0];
-    if (!(elementType.getType() instanceof ParameterizedType)
-        && !(elementType.getType() instanceof Class)) {
+    if (typeArguments.length == 0) {
       return Optional.empty();
     }
-
-    return Optional.of(elementType);
+    return Optional.of(Arrays.asList(typeArguments));
```
fmeum, line 379: Let's make the list unmodifiable.

src/test/java/com/code_intelligence/jazzer/mutation/mutator/BUILD.bazel
```diff
@@ -2,7 +2,7 @@ load("@contrib_rules_jvm//java:defs.bzl", "java_junit5_test")
 
 java_junit5_test(
     name = "StressTest",
-    size = "medium",
+    size = "large",
```
fmeum, line 5: I am looking into ways to run the test cases in parallel.

src/test/java/com/code_intelligence/jazzer/mutation/mutator/StressTest.java
```diff
@@ -90,6 +88,16 @@ public static Stream<Arguments> stressTestCases() {
...
+        arguments(
+            new TypeHolder<@NotNull Map<@NotNull String, @NotNull String>>() {}.annotatedType(),
+            "Map<String,String>", manyDistinctElements(), manyDistinctElements()),
+        arguments(new TypeHolder<Map<@NotNull String, @NotNull String>>() {}.annotatedType(),
+            "Nullable<Map<String,String>>", manyDistinctElements(), manyDistinctElements()),
+        arguments(
+            new TypeHolder<@WithSize(
+                min = 1, max = 3) @NotNull Map<@NotNull Integer, @NotNull Integer>>() {
+            }.annotatedType(),
+            "Map<Integer,Integer>", hasMapSizeInClosedRange(1, 3), hasMapSizeInClosedRange(1, 3)),
```
fmeum, line 100: Do the different values after mutations satisfy `manyDistinctElements`?

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
bertschneider: Added size checks.

src/test/java/com/code_intelligence/jazzer/mutation/mutator/StressTest.java
bertschneider: Yes, added it as well.

src/main/java/com/code_intelligence/jazzer/mutation/support/TypeSupport.java
bertschneider: Done

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
bertschneider: Good point, I didn't think about such narrow types. fuzztest tries 1k times before giving up, if I understand [the code](https://github.com/google/fuzztest/blob/a663ded6c36f050fbdc634a8fc81d553068d71d7/fuzztest/internal/domain.h#L1309) correctly. That seems excessive, doesn't it. Added a retry mechanism.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
bertschneider: Done

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
bertschneider: Done

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
bertschneider: A for-loop seemed so out of place compared to the functional style beforehand. But I agree, it's probably easier to follow. Changed.W#

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
```diff
@@ -0,0 +1,238 @@
...
+        .map(parameterTypes
+            -> parameterTypes.stream()
+                   .map(factory::tryCreate)
+                   .flatMap(StreamSupport::getOrEmpty)
+                   .collect(Collectors.toList()))
+        .filter(elementMutators -> elementMutators.size() == 2)
+        .map(elementMutators -> {
+          int min = MapMutator.DEFAULT_MIN_SIZE;
+          int max = MapMutator.DEFAULT_MAX_SIZE;
+          Optional<WithSize> withSize = Arrays.stream(type.getAnnotations())
```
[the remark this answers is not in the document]
bertschneider: Done

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
```diff
@@ -0,0 +1,238 @@
...
+final class MapMutatorFactory extends MutatorFactory {
+  @Override
+  public Optional<SerializingMutator<?>> tryCreate(AnnotatedType type, MutatorFactory factory) {
+    return parameterTypesIfParameterized(type, Map.class)
+        .map(parameterTypes
+            -> parameterTypes.stream()
+                   .map(factory::tryCreate)
+                   .flatMap(StreamSupport::getOrEmpty)
+                   .collect(Collectors.toList()))
+        .filter(elementMutators -> elementMutators.size() == 2)
```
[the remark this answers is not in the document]
bertschneider: Done

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
```diff
@@ -0,0 +1,238 @@
...
+package com.code_intelligence.jazzer.mutation.mutator.collection;
+
+import static com.code_intelligence.jazzer.mutation.support.Preconditions.require;
+import static com.code_intelligence.jazzer.mutation.support.TypeSupport.parameterTypesIfParameterized;
+import static java.lang.Math.min;
+import static java.lang.String.format;
+import static java.util.stream.Collectors.toMap;
+
+import com.code_intelligence.jazzer.mutation.annotation.WithSize;
+import com.code_intelligence.jazzer.mutation.api.*;
```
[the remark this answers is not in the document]
bertschneider: Done

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
```diff
@@ -0,0 +1,254 @@
...
+      } while (map.containsKey(key));
+      map.put(key, valueMutator.init(prng));
+    }
+
+    private void mutateElement(Map.Entry<K, V> entry, Map<K, V> map, PseudoRandom prng) {
+      K key = entry.getKey();
+      V value = entry.getValue();
+      map.remove(key);
+      if (prng.choice()) {
+        key = keyMutator.mutate(key, prng);
```
fmeum, line 180: This also needs a retry as you may mutate the key to be something that is already in the map.

fmeum (approved):
Looks good, only left a single comment.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
bertschneider: Good catch! Changed as well. If no unused key can be found during mutation, the value is mutated instead.

PR #654, opened 2023-03-08
junit: Add transitive dependency on launcher

bertschneider:
The transitive dependency on `org.junit.platform:junit-platform-launcher` removes the need to specify it manually in every project.

### DIFF
--- a/examples/junit-spring-web/build-and-run-tests.sh
+++ b/examples/junit-spring-web/build-and-run-tests.sh
@@ -31,7 +31,6 @@ sed -i "/<artifactId>jazzer-junit<\/artifactId>/ {n;s/<version>.*<\/version>/<ve
 ./mvnw install:install-file -Dfile=../../bazel-bin/deploy/jazzer-project.jar       -DpomFile=../../bazel-bin/deploy/jazzer-pom.xml
 ./mvnw install:install-file -Dfile=../../bazel-bin/deploy/jazzer-api-project.jar   -DpomFile=../../bazel-bin/deploy/jazzer-api-pom.xml
 
-
 ## Regression and unit tests
 echo "[SPRINGBOOT-JUNIT]: These unit and regression fuzz tests should pass"
 ./mvnw test -Dtest="JunitSpringWebApplicationTests#unitTestShouldPass+fuzzTestShouldPass"
@@ -52,7 +51,6 @@ else
   exit 1
 fi
 
-
 ## Fuzz tests
 echo "[SPRINGBOOT-JUNIT]: This fuzz test should pass"
 JAZZER_FUZZ=1 ./mvnw test -Dtest="JunitSpringWebApplicationTests#fuzzTestShouldPass"
@@ -70,5 +68,38 @@ else
   echo "[SPRINGBOOT-JUNIT]: Expected exit code 1, but got $exit_code"
   exit 1
 fi
+
+## CLI tests
+## Assert transitive JUnit dependencies are specified
+assertDependency() {
+  if ./mvnw dependency:tree | grep -q "$1"
+  then
+    echo "[SPRINGBOOT-JUNIT]: Found $1 dependency in project"
+  else
+    echo "[SPRINGBOOT-JUNIT]: Did not find $1 dependency in project"
+    exit 1
+  fi
+}
+assertDependency "org.junit.jupiter:junit-jupiter-api"
+assertDependency "org.junit.jupiter:junit-jupiter-params"
+assertDependency "org.junit.platform:junit-platform-launcher"
+
+# Only build project and test jars, no need for a fat-jar or test execution
+./mvnw jar:jar
+./mvnw jar:test-jar
+
+# Extract dependency locations
+out=$(./mvnw dependency:build-classpath -DforceStdout)
+deps=$(echo "$out" | sed '/^\[/d')
+
+# Directly execute Jazzer without Maven
+echo "[SPRINGBOOT-JUNIT]: Direct Jazzer execution of fuzz test should pass"
+java -cp "target/*:${deps}" \
+  com.code_intelligence.jazzer.Jazzer \
+  --target_class=com.example.JunitSpringWebApplicationTests \
+  --target_method=fuzzTestShouldPass \
+  --instrumentation_includes=com.example.* \
+  --custom_hook_includes=com.example.*
+
 
 echo "[SPRINGBOOT-JUNIT]: All tests passed"
--- a/src/main/java/com/code_intelligence/jazzer/driver/Driver.java
+++ b/src/main/java/com/code_intelligence/jazzer/driver/Driver.java
@@ -117,12 +117,10 @@ public class Driver {
       exit(0);
     }
 
-    if (!isAndroid) {
-      AgentInstaller.install(Opt.hooks);
-    }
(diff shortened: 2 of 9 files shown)


src/main/java/com/code_intelligence/jazzer/junit/BUILD.bazel
```diff
@@ -40,6 +40,7 @@ java_library(
         "//src/main/java/com/code_intelligence/jazzer/driver:opt",
         "@maven//:org_junit_jupiter_junit_jupiter_api",
         "@maven//:org_junit_jupiter_junit_jupiter_params",
+        "@maven//:org_junit_platform_junit_platform_launcher",
```
fmeum, line 43: We should move this to `runtime_deps` as its not used during compilation and add a comment along the lines of "The JUnit launcher that is part of the Jazzer driver needs this on the classpath to run an @FuzzTest with JUnit".

examples/junit-spring-web/build-and-run-tests.sh
```diff
@@ -31,6 +31,19 @@ sed -i "/<artifactId>jazzer-junit<\/artifactId>/ {n;s/<version>.*<\/version>/<ve
...
+  then
+    echo "[SPRINGBOOT-JUNIT]: Found $1 dependency in project"
+  else
+    echo "[SPRINGBOOT-JUNIT]: Did not find $1 dependency in project"
+    exit 1
+  fi
+}
+assertDependency "org.junit.jupiter:junit-jupiter-api"
+assertDependency "org.junit.jupiter:junit-jupiter-params"
+assertDependency "org.junit.platform:junit-platform-launcher"
```
fmeum, line 46: This is good test coverage in the sense that it will prevent a regression of this fix, but I would be more comfortable with a test that actually verifies we can run the passing fuzz test with Jazzer without a crash.

This would require running Jazzer with the classpath obtained as done by cifuzz here: https://github.com/CodeIntelligenceTesting/cifuzz/blob/da75a4201cbd5587ca536162d48a60f37618b614/internal/build/maven/maven.go#L137-L169.

fmeum:
After this has been merged, should we do a bugfix or just a regular release? I am leaning towards the latter.

src/main/java/com/code_intelligence/jazzer/driver/Driver.java
```diff
@@ -140,6 +138,9 @@ public static int start(List<String> args, boolean spawnsSubprocesses) throws IO
       }
     }
 
+    // Installing the agent after "findFuzzTarget" leads to an asan error in it on
+    // "Class.forName(targetClassName)", but only during native fuzzing.
+    AgentInstaller.install(Opt.hooks);
```
bertschneider, line 143: @fmeum Any idea why the agent needs to be installed before `findFuzzTarget` is called?

examples/junit-spring-web/build-and-run-tests.sh
bertschneider: Good point. I also added a test which directly invokes Jazzer and checks that no error is thrown. Did you have something like this in mind?

src/main/java/com/code_intelligence/jazzer/junit/FuzzTestExecutor.java
```diff
@@ -42,16 +37,23 @@
 import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.jupiter.api.extension.ReflectiveInvocationContext;
 
-abstract class FuzzTestExecutor {
+class FuzzTestExecutor {
```
bertschneider, line 40: @fmeum As discussed, I removed the different `TestExecutors`, and `libFuzzerArgs` are now appended accordingly, respectively used as fake `argv0`. I'm somewhat unsure why the `execute` methods of the two implementations differed, though.

src/main/java/com/code_intelligence/jazzer/driver/Driver.java
fmeum: We have a hook on `Class.forName`, but that's only installed by the agent... I have to take a deeper look, I don't know.

src/main/java/com/code_intelligence/jazzer/junit/FuzzTestExecutor.java
bertschneider: The `--keep_going` flag is only evaluated after checking the `findingHandler`. As that  always returns `false`, it looks like execution should stop at the first finding. So you're right, the previous behavior differed between fuzz tests started by JUnit and our CLI. Was/is that intended?

bertschneider: Ok, so it was intended. As there is probably no idiomatic way to support something like `keep_going` in JUnit (why would there) the different behavior seems reasonable. 

I only passed down a flag to distinguish the two modes, as the two implementations were almost identical. That being said, it's not super clean. Having multiple implementations or other means to configure `execute` seemed similar overkill.

src/main/java/com/code_intelligence/jazzer/driver/Driver.java
fmeum: @bertschneider Could you create a ticket for this and assign it to me so that we don't lose track of it?

src/main/java/com/code_intelligence/jazzer/junit/FuzzTestExecutor.java
```diff
@@ -65,27 +72,16 @@ public static FuzzTestExecutor prepare(ExtensionContext context, String maxDurat
...
-
-  private static FuzzTestExecutor prepareForTestRunner(ExtensionContext context, String maxDuration)
-      throws IOException {
     Path baseDir =
         Paths.get(context.getConfigurationParameter("jazzer.internal.basedir").orElse(""))
             .toAbsolutePath();
 
-    final Class<?> fuzzTestClass = context.getRequiredTestClass();
+    List<String> originalLibFuzzerArgs = getLibFuzzerArgs(context);
+    String fakeArgv0 =
```
fmeum, line 80: Let's call this `argv0` as it may not bee fake anymore.

src/main/java/com/code_intelligence/jazzer/junit/BUILD.bazel
```diff
@@ -29,6 +29,11 @@ java_library(
     visibility = [
         "//examples/junit/src/test/java/com/example:__pkg__",
     ],
+    runtime_deps = [
+        # The JUnit launcher that is part of the Jazzer driver needs this on the classpath to run an @fuzztest with JUnit.
```
fmeum, line 33:
```suggestion
        # The JUnit launcher that is part of the Jazzer driver needs this on the classpath to run an @FuzzTest with JUnit.
```

fmeum (approved):

src/main/java/com/code_intelligence/jazzer/junit/FuzzTestExecutor.java
bertschneider: done

src/main/java/com/code_intelligence/jazzer/junit/BUILD.bazel
bertschneider: done

PR #655, opened 2023-03-08
Keep jars separated by jar when instrumneting offline

TheCoryBarker:
This is to keep jars separated when instrumenting offline, rather than merge them.

For example, with this input:

## INPUT ##
jarOne.jar
|__ classOne.java
|__ classTwo.java

jarTwo.jar
|__ classThree.java
|__ classFour.java

## CURRENT OUTPUT WOULD BE: ##
output.jar
|__ classOne.java
|__ classTwo.java
|__ classThree.java
|__ classFour.java

## WITH THIS UPDATE, OUTPUT IS: ##
jarOne.instrumented.jar
|__ classOne.java
|__ classTwo.java

jarTwo.instrumented.jar
|__ classThree.java
|__ classFour.java


src/main/java/com/code_intelligence/jazzer/driver/OfflineInstrumentor.java
```diff
@@ -55,11 +55,14 @@ public static boolean instrumentJars(List<String> jarLists) {
 
     List<String> errorMessages = new ArrayList<>();
     for (String jarPath : jarLists) {
+      String jarBaseName =
```
bertschneider, line 58: Come to think of it, `cp` could [contain directories](https://docs.oracle.com/javase/8/docs/technotes/tools/windows/classpath.html#A1100762), in which case all jars in those are included in the classpath. As that's probably not the intended use-case here, we could ignore directories altogether. That could lead to some confusion those. 

Another option could be to change the flag `instrument_only` to contain the list of jar files and not "misuse" `cp`. WDYT?

src/main/java/com/code_intelligence/jazzer/driver/OfflineInstrumentor.java
```diff
@@ -55,11 +55,14 @@ public static boolean instrumentJars(List<String> jarLists) {
...
+          jarPath.substring(jarPath.lastIndexOf('/') + 1, jarPath.lastIndexOf('.'));
+
       try {
         Log.info("Instrumenting jar file: " + jarPath);
         errorMessages = createInstrumentedClasses(jarPath);
         createInstrumentedJar(
-            jarPath, Opt.dumpClassesDir, Opt.dumpClassesDir + ".instrumented.jar");
+            jarPath, Opt.dumpClassesDir + "/" + jarBaseName, jarBaseName + ".instrumented.jar");
```
bertschneider, line 65: Use `File.separator` for system independent separator character, or build up the path as `File` or `Path`.

src/main/java/com/code_intelligence/jazzer/agent/RuntimeInstrumentor.kt
```diff
@@ -57,6 +57,17 @@ class RuntimeInstrumentor(
         protectionDomain: ProtectionDomain?,
         classfileBuffer: ByteArray,
     ): ByteArray? {
+        var pathPrefix = ""
+        if (Opt.instrumentOnly && protectionDomain != null) {
```
bertschneider, line 61: `protectionDomain` was already null-checked so that `?` isn't required anymore. Also the methods calls on `protectionDomain` could be changed to  a property access for more idiomatic Kotlin code.

src/main/java/com/code_intelligence/jazzer/agent/RuntimeInstrumentor.kt
```diff
@@ -57,6 +57,17 @@ class RuntimeInstrumentor(
...
+            var jarUrl = protectionDomain?.getCodeSource().getLocation().getFile().toString()
+            if (!jarUrl.isNullOrEmpty()) {
```
bertschneider, line 63: `jarUrl.isNotEmpty()` could be simpler.

src/main/java/com/code_intelligence/jazzer/agent/RuntimeInstrumentor.kt
```diff
@@ -57,6 +57,17 @@ class RuntimeInstrumentor(
...
+                var fileName = jarUrl.substring(jarUrl.lastIndexOf('/') + 1, jarUrl.lastIndexOf('.'))
```
bertschneider, line 64: I'm not sure if this will work in all situations, e.g. if classes are located directly in a directory and not a jar file. Could you please add checks if the returned indices are valid?

src/main/java/com/code_intelligence/jazzer/agent/RuntimeInstrumentor.kt
```diff
@@ -57,6 +57,17 @@ class RuntimeInstrumentor(
...
+                if (!fileName.isNullOrEmpty()) {
+                    pathPrefix = fileName + "/"
```
bertschneider, line 66: Use `Path.separator`

src/main/java/com/code_intelligence/jazzer/agent/RuntimeInstrumentor.kt
bertschneider, line 61: I wonder if we should always do this. WDYT?

bertschneider (changes requested):

src/main/java/com/code_intelligence/jazzer/agent/RuntimeInstrumentor.kt
TheCoryBarker: You'll know more about the expected average use case here, but to me this makes sense to always keep these separated when running instrument_only. If I need my jars merged into one jar, I think I should do that before running --instrument_only command on my list of jars.

src/main/java/com/code_intelligence/jazzer/agent/RuntimeInstrumentor.kt
TheCoryBarker: The code will only work on instrumenting a list of jars that are passed in, directories aren't supported in this check-in. The jar doesn't need to end with .jar though so I did update this to account for that.

Since we switched to using --instrument_only=<list_of_jars> instead of --cp which can contain directories I think this makes sense

src/main/java/com/code_intelligence/jazzer/driver/OfflineInstrumentor.java
TheCoryBarker: Good idea, I updated this to use --instrument_only=<list_of_jars>

src/main/java/com/code_intelligence/jazzer/driver/Opt.java
```diff
@@ -183,7 +183,7 @@ public final class Opt {
       Log.error("--nodedup is not supported with --ignore or --keep_going");
       exit(1);
     }
-    if (instrumentOnly && dumpClassesDir.isEmpty()) {
+    if (instrumentOnly != null && !instrumentOnly.isEmpty() && dumpClassesDir.isEmpty()) {
```
bertschneider, line 186: As with the checks of the other parameters above, we can rely on `instrumentOnly` to never be `null`.

src/main/java/com/code_intelligence/jazzer/driver/Opt.java
```diff
@@ -147,8 +147,8 @@ public final class Opt {
       boolSetting("internal.conditional_hooks", false, null);
 
   // Some scenarios require instrumenting the jar before fuzzing begins
-  public static final boolean instrumentOnly =
-      boolSetting("instrument_only", false, "Instrument jar file. No fuzzing is performed.");
+  public static final List<String> instrumentOnly =
+      stringListSetting("instrument_only", ',', "Instrument jar file. No fuzzing is performed.");
```
bertschneider, line 151: Mention the separator char in the description? Something like `comma separated list of ...`?

src/main/java/com/code_intelligence/jazzer/driver/OfflineInstrumentor.java
```diff
@@ -55,13 +57,31 @@ public static boolean instrumentJars(List<String> jarLists) {
...
+      Log.info("Instrumenting jar file: " + jarPath);
+
       try {
-        Log.info("Instrumenting jar file: " + jarPath);
         errorMessages = createInstrumentedClasses(jarPath);
-        createInstrumentedJar(
-            jarPath, Opt.dumpClassesDir, Opt.dumpClassesDir + ".instrumented.jar");
+      } catch (IOException e) {
+        errorMessages.add("Failed to instrument jar: " + jarPath
+            + ". Please ensure the file at this location is a jar file.\n\tError Message: " + e);
```
bertschneider, line 76: There is also a system independent EOL function: `System.lineSeparator()`.

src/main/java/com/code_intelligence/jazzer/agent/RuntimeInstrumentor.kt
```diff
@@ -57,6 +58,24 @@ class RuntimeInstrumentor(
         protectionDomain: ProtectionDomain?,
         classfileBuffer: ByteArray,
     ): ByteArray? {
+        var pathPrefix = ""
+        if (!Opt.instrumentOnly.isEmpty() && protectionDomain != null) {
+            var outputPathPrefix = protectionDomain.getCodeSource().getLocation().getFile().toString()
+            if (outputPathPrefix.isNotEmpty()) {
+                if (outputPathPrefix.contains(File.separator)) {
+                    outputPathPrefix = outputPathPrefix.substring(outputPathPrefix.lastIndexOf(File.separator) + 1, outputPathPrefix.length)
```
bertschneider, line 66: No need to specify the second parameter of `substring`, as that's implicitly the end of the string. Whichever version you prefer, though.

src/main/java/com/code_intelligence/jazzer/agent/RuntimeInstrumentor.kt
bertschneider: Well, let's keep it as is for now then.

bertschneider:
Just minor comments left. Thanks for your work!

bertschneider (approved):
Looks good. Thx for all your work! We can merge this, once rebased on the latest `main` version.

bertschneider:
I took the liberty to rebase + merge. Thanks again.

TheCoryBarker:
Sorry, I was slow getting to the rebase, thank you Norbert

PR #661, opened 2023-03-13
Add support for length limits to `ByteArrayMutatorFactory`

br-lewis:
This adds an annotation, `WithLength`, that can be applied to byte arrays and adds support in the `ByteArrayMutatorFactory` for it.

I'm not sure about the approach to `mutate` which will extend with 0s or truncate the array returned by libfuzzer to fit within the min and maximum size limit. I'm also not sure how to add tests for this considering it uses libfuzzer for mutation rather than our PRNG.


fmeum:
@br-lewis The byte array mutator has a mock version that doesn't use libFuzzer. You can enable it with https://github.com/CodeIntelligenceTesting/jazzer/blob/89ced1faf695ecc9b0e7084020724df654b2ff21/src/test/java/com/code_intelligence/jazzer/mutation/mutator/BUILD.bazel#L7 
The actual implementation is here and can be relied upon in tests: https://github.com/CodeIntelligenceTesting/jazzer/blob/89ced1faf695ecc9b0e7084020724df654b2ff21/src/main/java/com/code_intelligence/jazzer/mutation/mutator/libfuzzer/LibFuzzerMutator.java#L72

br-lewis:
Thanks! I added some tests for the byte array mutator. I couldn't test that minimum length works as expected in all cases but I checked that it at least works for init.

src/test/java/com/code_intelligence/jazzer/mutation/mutator/lang/ByteArrayMutatorTest.java
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  @Test
+  void testMinLengthInitClamp() {
+    SerializingMutator<byte[]> mutator =
+        (SerializingMutator<byte[]>) LangMutators.newFactory().createOrThrow(
+            new TypeHolder<byte @NotNull @WithLength(min = 5)[]>() {}.annotatedType());
+    assertThat(mutator.toString()).isEqualTo("byte[]");
+
+    // The mock libfuzzer mutator doesn't currently support outputting less data than it's given, so
```
fmeum, line 96: Feel free to modify the mock mutator so that you can test this case more easily. I just made up its implementation on the spot, no reason to feel attached to it. :-)

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/ByteArrayMutatorFactory.java
```diff
@@ -80,16 +95,31 @@ public byte[] detach(byte[] value) {
...
+      int len = prng.closedRange(minLength, maxLength);
+      byte[] bytes = new byte[len];
       prng.bytes(bytes);
       return bytes;
     }
 
     @Override
     public byte[] mutate(byte[] value, PseudoRandom prng) {
+      if (value.length == 0) {
+        return init(prng);
```
fmeum, line 107: Could you explain why this special case is needed?

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/ByteArrayMutatorFactory.java
```diff
@@ -80,16 +95,31 @@ public byte[] detach(byte[] value) {
     public byte[] init(PseudoRandom prng) {
       // TODO: Improve the way the upper bound is determined, e.g. grow it over time and/or add
       //  support for the WithSize annotation.
```
fmeum, line 97: This comment is half outdated with this change.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/ByteArrayMutatorFactory.java
br-lewis: Funnily enough, I realized earlier today that it probably isn't needed because libfuzzer's mutate can probably handle an empty value array but didn't think to remove it. I can to make the implementation simpler if you'd prefer.

src/test/java/com/code_intelligence/jazzer/mutation/mutator/lang/ByteArrayMutatorTest.java
br-lewis: I spent a bit of time thinking about how I could change it but wasn't sure what a better method could be since there's no prng to hook into. Maybe use some of the elements in the array as instructions to modify the rest of it? Like have the first element be the new array's length?

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/ByteArrayMutatorFactory.java
fmeum: In that case I would prefer to remove it, otherwise we will keep wondering about why it is there.

src/test/java/com/code_intelligence/jazzer/mutation/mutator/lang/ByteArrayMutatorTest.java
fmeum: Yes, that would be an option. If it is really only for this one test, you could also have the mock mutator get the length from e.g. `System.getProperty`. For test mocks I wouldn't consider that too hacky.

br-lewis: I went with `System.getProperty` as that seemed the easiest to understand though I think it does mean you can't run these tests in parallel but I don't think that's a very common usecase anyway.

src/main/java/com/code_intelligence/jazzer/mutation/annotation/WithLength.java
```diff
@@ -0,0 +1,33 @@
...
+package com.code_intelligence.jazzer.mutation.annotation;
+
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.ElementType.TYPE_USE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+@Target({PARAMETER, TYPE_USE})
```
fmeum, line 26:
```suggestion
@Target(TYPE_USE)
```
should be sufficient (I may have added `PARAMETER` in places where it isn't necessary).

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/ByteArrayMutatorFactory.java
```diff
@@ -38,18 +39,32 @@
...
+    private ByteArrayMutator(int min, int max) {
+      this.minLength = min;
+      this.maxLength = max;
+    }
 
     @Override
     public byte[] read(DataInputStream in) throws IOException {
-      int length = Math.max(in.readInt(), 0);
+      // clamp the length to within min and max length
+      int length = Math.min(Math.max(in.readInt(), minLength), maxLength);
```
fmeum, line 67: Could make sense to extract a `clamp` function and it to the `support` package.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/libfuzzer/LibFuzzerMutator.java
```diff
@@ -70,7 +76,12 @@ private static int defaultMutate(byte[] buffer, int size) {
   }
 
   private static int defaultMutateMock(byte[] buffer, int size) {
+    String newSizeProp = System.getProperty(MOCK_SIZE_KEY);
     int newSize = (buffer.length + size) / 2;
+    if (newSizeProp != null) {
+      newSize = Integer.parseInt(newSizeProp);
```
fmeum, line 82:
```suggestion
      newSize = Integer.parseUnsignedInt(newSizeProp);
```

src/test/java/com/code_intelligence/jazzer/mutation/mutator/StressTest.java
```diff
@@ -70,7 +70,7 @@
 public class StressTest {
   private static final int NUM_INITS = 1000;
   private static final int NUM_MUTATE_PER_INIT = 100;
-  private static final double MANY_DISTINCT_ELEMENTS_RATIO = 0.5;
+  private static final double MANY_DISTINCT_ELEMENTS_RATIO = 0.45;
```
fmeum, line 73: You can keep this as is and selectively use `distinctElementsRatio` with a lower bound on the individual tests that need it.

src/test/java/com/code_intelligence/jazzer/mutation/mutator/lang/ByteArrayMutatorTest.java
```diff
@@ -0,0 +1,142 @@
...
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Test;
+
+public class ByteArrayMutatorTest {
+  /**
+   * Some tests may set {@link LibFuzzerMutator#MOCK_SIZE_KEY} which can interfere with other tests
+   * unless cleared.
+   */
+  @AfterEach
+  void CleanMockSize() {
```
fmeum, line 36:
```suggestion
  void cleanMockSize() {
```

src/test/java/com/code_intelligence/jazzer/mutation/mutator/lang/ByteArrayMutatorTest.java
```diff
@@ -0,0 +1,142 @@
...
+    System.clearProperty(LibFuzzerMutator.MOCK_SIZE_KEY);
+  }
+
+  @Test
+  void testBasicFunction() {
+    SerializingMutator<byte @NotNull[]> mutator =
```
fmeum, line 42: Not needed:
```suggestion
    SerializingMutator<byte[]> mutator =
```

src/test/java/com/code_intelligence/jazzer/mutation/mutator/lang/ByteArrayMutatorTest.java
```diff
@@ -0,0 +1,142 @@
...
+    SerializingMutator<byte[]> mutator =
+        (SerializingMutator<byte[]>) LangMutators.newFactory().createOrThrow(
+            new TypeHolder<byte @NotNull @WithLength(min = 5)[]>() {}.annotatedType());
+    assertThat(mutator.toString()).isEqualTo("byte[]");
+
+    byte[] arr;
+    try (MockPseudoRandom prng = mockPseudoRandom(3)) {
+      arr = mutator.init(prng);
+    } catch (AssertionError e) {
+      // init will call closedrange(min, max) and the mock prng will assert that the given value
```
fmeum, line 114: I think that there is an `assertThrows` that would make this easier to follow.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/ByteArrayMutatorFactory.java
```diff
@@ -89,7 +104,17 @@ public byte[] init(PseudoRandom prng) {
     public byte[] mutate(byte[] value, PseudoRandom prng) {
       // TODO: The way maxSizeIncrease is determined is just a heuristic and hasn't been
       //  benchmarked.
-      return LibFuzzerMutator.mutateDefault(value, Math.max(8, value.length / 16));
+      byte[] mutated = LibFuzzerMutator.mutateDefault(value, Math.max(8, value.length / 16));
```
fmeum, line 107: What I missed before: The second argument is the allowed length increase. We should probably cap it at `maxLength - value.length` to avoid unnecessary truncations.

src/test/java/com/code_intelligence/jazzer/mutation/mutator/lang/ByteArrayMutatorTest.java
```diff
@@ -0,0 +1,143 @@
...
+  void testMinLengthInitClamp() {
+    SerializingMutator<byte[]> mutator =
+        (SerializingMutator<byte[]>) LangMutators.newFactory().createOrThrow(
+            new TypeHolder<byte @NotNull @WithLength(min = 5)[]>() {}.annotatedType());
+    assertThat(mutator.toString()).isEqualTo("byte[]");
+
+    // init will call closedrange(min, max) and the mock prng will assert that the given value
+    // above is between those values which we want to fail here to show that we're properly
+    // clamping the range
+    Assertions.assertThrows(AssertionError.class, () -> {
```
fmeum, line 114: Could you move the `assertThrows` inside the try block? Otherwise it could mask unexpected `AssertionError`s coming from the try block calling `MockPseudoRandom#close`. Similar consideration applies to the try block in the test above.

fmeum (approved):
Looks good to me, only a few comments that I missed on the earlier review and this is ready to go. Sorry!

fmeum:
Looks good to me - I especially like the comprehensive test coverage.

PR #664, opened 2023-03-14
Extend list of ignored packages

bertschneider:
As instrumenting Mockito can cause instrumentation cycles it's now ignored. Instrumenting Byte Buddy itself could cause similar issues, hence, it's now ignored as well.

Lastly, the whole JUnit package is ignored due to other sub-packages that could distract the fuzzer and slow down startup.


src/main/java/com/code_intelligence/jazzer/junit/AgentConfigurator.java
```diff
@@ -60,6 +60,6 @@ private static void applyCommonConfiguration() {
         String.join(File.pathSeparator, "com.google.testing.junit.**", "com.intellij.**",
             "io.github.classgraph.**", "junit.framework.**", "net.bytebuddy.**",
             "org.apiguardian.**", "org.assertj.core.**", "org.hamcrest.**", "org.junit.**",
-            "org.opentest4j.**", "org.apache.maven.**", "org.gradle.**"));
+            "org.opentest4j.**", "org.mockito.**", "org.apache.maven.**", "org.gradle.**"));
```
bertschneider, line 63: @fmeum Why not exclude these all the time?

bertschneider: Ok, then let's leave it as is for this PR.

.github/workflows/run-all-tests.yml
```diff
@@ -79,10 +79,10 @@ jobs:
         shell: bash
 
       - name: Build
-        run: bazelisk build ${{env.BUILD_BUDDY_CONFIG}} --java_runtime_version=local_jdk_${{ matrix.jdk }} --disk_cache=${{ matrix.cache }} ${{ matrix.bazel_args }} //...
+        run: bazelisk build ${{env.BUILD_BUDDY_CONFIG}} --java_runtime_version=local_jdk_${{ matrix.jdk }} --disk_cache=${{ matrix.cache }} ${{ matrix.bazel_args }} --build_tag_filters="-no-${{ matrix.arch }}-jdk${{ matrix.jdk }},-no-jdk${{ matrix.jdk }}" //...
```
fmeum, line 82: Just want to verify that this is required: Do the tests that use this tag actually fail to build or just fail to run?

fmeum (approved):

.github/workflows/run-all-tests.yml
bertschneider: True, it's not required for this test. In #643 I added (or better tried to) a test which only builds with later Java versions. So I thought, while I'm at it, I add `--build_tag_filters` as well. Adding a differently named set of build tags seems too much. WDYT?

fmeum: Agreed, if this is needed elsewhere than it's fine to add here.

PR #665, opened 2023-03-14
Add support for map fields

fmeum:


bertschneider (approved):
This fits nicely into the mutator framework.

PR #671, opened 2023-03-17
bazel: Use a hermetic Java toolchain compiling for Java 8

fmeum:
This simplifies the setup for contributors as well as the release process. Since we can't use `--release 8`, we still need to check compatibility with Java 8 runtime libraries in our CI, which still uses the local JDK 8 to run tests against.

Also removes stamping from Maven builds as we already have the version number baked in.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,7 +33,6 @@ jobs:
         shell: bash
         run: |
           echo "build --announce_rc" >> .bazelrc
-          echo "build --config=maven" >> .bazelrc
           echo "build:linux --config=toolchain" >> .bazelrc
           echo "build:linux --extra_toolchains=@llvm_toolchain//:cc-toolchain-x86_64-linux" >> .bazelrc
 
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -33,7 +33,6 @@ jobs:
         shell: bash
         run: |
           echo "build --announce_rc" >> .bazelrc
-          echo "build --config=maven" >> .bazelrc
```
bertschneider, line 36: Isn't this still needed to apply the `build:maven --stamp` setting?

fmeum: Good point, I deleted that config in an earlier version and didn't add this back.

Thinking about this setting again, do we even want stamping? The version number should be a sufficient indicator for releases and getting rid of it would improve reproducibility of the released artifacts.

bertschneider: For my taste the version number would be sufficient.

bertschneider (approved):

PR #674, opened 2023-03-21
Add `WithUtf8Length` annotation and support for it

br-lewis:
This adds a new annotation `WithUtf8Length` which can apply to strings to limit the maximum and minimum number of UTF-8 bytes that go into them by creating a `WithLength` annotation of the same `min` and `max` values and applying that to the underlying byte array mutator.

### DIFF
--- a/src/main/java/com/code_intelligence/jazzer/mutation/annotation/WithUtf8Length.java
+++ b/src/main/java/com/code_intelligence/jazzer/mutation/annotation/WithUtf8Length.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2023 Code Intelligence GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.code_intelligence.jazzer.mutation.annotation;
+
+import static java.lang.annotation.ElementType.TYPE_USE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/**
+ * An annotation that applies to {@link String} to set limits around the number of <strong>UTF-8
+ * bytes</strong> in the string. <p> Due to the fact that our String mutator is backed by the byte
+ * array mutator, it's difficult to know how many characters we'll get from the byte array we get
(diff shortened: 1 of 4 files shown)


src/main/java/com/code_intelligence/jazzer/mutation/support/TypeSupport.java
```diff
@@ -186,6 +187,56 @@ public static AnnotatedType notNull(AnnotatedType type) {
...
+  /**
+   * Constructs an anonymous WithLength class that can be applied as an annotation to {@code type}
+   * with the given
+   * {@code min} and {@code max} values.
+   * @param type
+   * @param min
+   * @param max
+   * @return {@code type} with a `WithLength` annotation applied to it
+   */
+  public static AnnotatedType withLength(AnnotatedType type, int min, int max) {
```
fmeum, line 199: Yeah, that's the unfortunate part of using different annotations. I don't necessarily see a clearly better way of doing this. We could consider allowing `WithUtf8Length` on `byte[]` to simplify this.

src/main/java/com/code_intelligence/jazzer/mutation/annotation/WithUtf8Length.java
```diff
@@ -0,0 +1,41 @@
...
+package com.code_intelligence.jazzer.mutation.annotation;
+
+import static java.lang.annotation.ElementType.TYPE_USE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/**
+ * An annotation that applies to {@link String} to set limits around the number of <strong>UTF-8
```
fmeum, line 26: How about "to limit the length of the UTF-8 encoding of the string"?

src/main/java/com/code_intelligence/jazzer/mutation/annotation/WithUtf8Length.java
```diff
@@ -0,0 +1,41 @@
...
+ * bytes</strong> in the string. <p> Due to the fact that our String mutator is backed by the byte
+ * array mutator, it's difficult to know how many characters we'll get from the byte array we get
+ * from libfuzzer. Rather than reuse {@link WithLength} for strings which may give the impression
+ * that {@link String#length()} will return a value between {@code min} and {@code max}, we use this
+ * annotation to help make clear that the string consists of between
+ * {@code min} and {@code max} UTF-8 bytes, not necessarily characters.
```
fmeum, line 32:
```suggestion
 * {@code min} and {@code max} UTF-8 bytes, not necessarily (UTF-16) characters.
```
"character" is quite overloaded in the context of strings, but this is what it means in Java.

fmeum (approved):

src/main/java/com/code_intelligence/jazzer/mutation/support/TypeSupport.java
br-lewis: I don't find this too bad right now. The byte array mutator doesn't care about the distinction between utf8 and not so I think it's conceptually simpler to do this conversion in the string mutator where that distinction is meaningful. We might want to try to simplify how we do annotation manipulation if we find ourselves doing lots of it but that's something we can easily fix later since this isn't part of the public API.

fmeum (approved):

PR #675, opened 2023-03-21
Improve list mutator

0xricksanchez:
Improves the list mutator in a couple of ways:

* Gets rid of oscillation issues once `maxSize` is reached
* Every mutation is decided by a fair random pick
* Added chunk operations that complement the mutation of single elements

### DIFF
--- a/src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/ListMutatorFactory.java
+++ b/src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/ListMutatorFactory.java
@@ -16,34 +16,45 @@
 
 package com.code_intelligence.jazzer.mutation.mutator.collection;
 
+import static com.code_intelligence.jazzer.mutation.support.Preconditions.require;
 import static com.code_intelligence.jazzer.mutation.support.TypeSupport.parameterTypeIfParameterized;
 import static java.lang.Math.min;
+import static java.lang.String.format;
 
+import com.code_intelligence.jazzer.mutation.annotation.WithSize;
 import com.code_intelligence.jazzer.mutation.api.Debuggable;
 import com.code_intelligence.jazzer.mutation.api.MutatorFactory;
 import com.code_intelligence.jazzer.mutation.api.PseudoRandom;
 import com.code_intelligence.jazzer.mutation.api.SerializingInPlaceMutator;
 import com.code_intelligence.jazzer.mutation.api.SerializingMutator;
+import com.code_intelligence.jazzer.mutation.support.MutationAction;
+import com.code_intelligence.jazzer.mutation.support.RandomSupport;
 import java.io.DataInputStream;
 import java.io.DataOutputStream;
 import java.io.IOException;
 import java.lang.reflect.AnnotatedType;
 import java.util.AbstractList;
 import java.util.ArrayList;
+import java.util.EnumSet;
+import java.util.Iterator;
 import java.util.List;
+import java.util.ListIterator;
 import java.util.Optional;
 import java.util.RandomAccess;
+import java.util.Set;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 final class ListMutatorFactory extends MutatorFactory {
   @Override
   public Optional<SerializingMutator<?>> tryCreate(AnnotatedType type, MutatorFactory factory) {
+    Optional<WithSize> withSize = Optional.ofNullable(type.getAnnotation(WithSize.class));
+    int minSize = withSize.map(WithSize::min).orElse(ListMutator.DEFAULT_MIN_SIZE);
+    int maxSize = withSize.map(WithSize::max).orElse(ListMutator.DEFAULT_MAX_SIZE);
     return parameterTypeIfParameterized(type, List.class)
         .flatMap(factory::tryCreate)
-        .map(elementMutator
-            -> new ListMutator<>(
-                elementMutator, ListMutator.DEFAULT_MIN_SIZE, ListMutator.DEFAULT_MAX_SIZE));
+        .map(elementMutator -> new ListMutator<>(elementMutator, minSize, maxSize));
   }
 
   private static final class ListMutator<T> extends SerializingInPlaceMutator<List<T>> {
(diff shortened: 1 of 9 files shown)


0xricksanchez:
Currently I do not know why the `//tests:ExperimentalMutatorComplexProtoFuzzer` does not pass in the CI. It works just fine on my Mac as well as from within a fresh Ubuntu 22.04 VM

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/ListMutatorFactory.java
```diff
@@ -97,18 +112,98 @@ public void initInPlace(List<T> reference, PseudoRandom prng) {
...
+    private void insertRandomChunk(List<T> list, PseudoRandom prng, int maxSize) {
+      // Enforces slower growth for smaller lists and lists that are close to `maxSize`.
+      int chunkSize = prng.closedRange(1, Math.min(maxSize - list.size(), list.size()));
+      int chunkOffset = prng.indexIn(list);
+      for (int i = 0; i < chunkSize; i++) {
+        T newElement = elementMutator.init(prng);
+        try {
+          list.add(chunkOffset + i, newElement);
+        } catch (Exception e) {
+          System.out.println(list.getClass().getName());
```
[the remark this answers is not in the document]
0xricksanchez: that thing is an artifact from debugging yesterday. removed the `try-catch`

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/ListMutatorFactory.java
```diff
@@ -172,5 +264,15 @@ public T get(int i) {
     public int size() {
       return mutableList.size();
     }
+
+    @Override
```
bertschneider, line 268: No override necessary.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/ListMutatorFactory.java
```diff
@@ -97,18 +111,95 @@ public void initInPlace(List<T> reference, PseudoRandom prng) {
...
-      } else if (!prng.trueInOneOutOf(4)) {
-        int i = prng.indexIn(list);
-        list.set(i, elementMutator.mutate(list.get(i), prng));
-      } else if (list.size() < maxSize) {
-        list.add(list.get(list.size() - 1)); // TODO: Create deep copy or init new entry.
-      } else if (list.size() > minSize) {
-        list.remove(list.size() - 1);
+        return;
+      }
+      List<Action> s = new ArrayList<>();
```
bertschneider, line 170: Building up the list of possible `Action`s could be extracted out into a static method in the enum class, expecting a `Collection`. With that, it could be reused for other collection mutators as well.

src/test/java/com/code_intelligence/jazzer/mutation/mutator/StressTest.java
```diff
@@ -117,33 +117,33 @@ null, emptyList(), singletonList(null), singletonList(false), singletonList(true
             "Map<Integer,Integer>", all(mapSizeInClosedRange(1, 3), manyDistinctElements()),
             all(mapSizeInClosedRange(1, 3), manyDistinctElements())),
         arguments(asAnnotatedType(byte.class), "Byte",
-            // init is heavily biased towards special values and only returns a uniformly random
-            // value in 1 out of 5 calls.
+            // init is heavily biased towards special values and only returns a uniformly
+            // random value in 1 out of 5 calls.
```
bertschneider, line 121: Only comment formatting in this file? Revert?

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/ListMutatorFactory.java
0xricksanchez: Extracted into a support class so I can use it for the map improvements in the other PR

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/ListMutatorFactory.java
```diff
@@ -97,18 +112,74 @@ public void initInPlace(List<T> reference, PseudoRandom prng) {
...
+    }
+
+    private void insertRandomChunk(List<T> list, PseudoRandom prng, int maxSize) {
+      // Enforces slower growth for smaller lists and lists that are close to
+      // `maxSize`.
+      int chunkSize = prng.closedRange(1, Math.min(maxSize - list.size(), list.size()));
+      int chunkOffset = prng.indexIn(list);
+      for (int i = 0; i < chunkSize; i++) {
+        T newElement = elementMutator.init(prng);
+        list.add(chunkOffset + i, newElement);
```
fmeum, line 135: This presumably requires quite a bit of copying - I expect this to be on the order of <chunk size> * <list size> as it needs to copy over the tail of the list in each iteration (see https://github.com/openjdk/jdk/blob/9df20600592427550998c6685f103737e3115a51/src/java.base/share/classes/java/util/ArrayList.java#L487-L489).

Instead, you could do a batch `addAll` at the correct position. This requires collecting the new elements in temporary new list (which you could initialize with the correct capacity `chunkSize` directly).

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/ListMutatorFactory.java
```diff
@@ -97,18 +112,74 @@ public void initInPlace(List<T> reference, PseudoRandom prng) {
       }
     }
 
+    private void eraseRandomChunk(List<T> list, PseudoRandom prng, int minSize) {
+      int chunkSize =
+          prng.closedRange(1, Math.max(1, Math.min(list.size() - minSize, list.size() / 2)));
```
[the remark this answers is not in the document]
0xricksanchez: `RandomSupport#clamp` uses `Math.min(Math.max(x,y), z)` here we have `Math.max(Math.min(x,y), z)`. I could add a new variant `clampMax` to reflect this and rename the current one `clampMin`?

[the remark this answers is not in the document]
fmeum: Both `max(min(value, upper), lower)` and `min(max(value, lower), upper)` give the same result, so if the aim is to clamp `list.size() / 2` to the interval `[1, list.size() - minSize]`, `RandomSupport#clamp` should work just fine.

Erasing up to half the list in a single mutation does feel a bit much. Did you check what fuzztest and/or libprotobuf-mutator do in this case?

src/main/java/com/code_intelligence/jazzer/mutation/mutator/proto/BuilderAdapters.java
```diff
@@ -43,11 +43,19 @@ public int size() {
...
-      public boolean add(U u) {
-        builder.addRepeatedField(field, u);
+      public boolean add(U element) {
+        builder.addRepeatedField(field, element);
         return true;
       }
 
+      @Override
+      public void add(int index, U element) {
+        // TODO: Simply appending and ignoring the index is not perfect
```
fmeum, line 53: Could you file a follow-up ticket for this?

fmeum (approved):

src/main/java/com/code_intelligence/jazzer/mutation/mutator/proto/BuilderAdapters.java
0xricksanchez: done

PR #681, opened 2023-03-22
tests: Parallelize StressTest

fmeum:


0xricksanchez:
On my Mac, I'm getting a rough 40% runtime improvement on average over the course of 10 measured runs (~82 seconds down to ~49 seconds).

fmeum:
Thanks for the data, looks like this PR would indeed improve the runtime quite significantly.

I submitted https://github.com/bazel-contrib/rules_jvm/pull/169 to fix the test runner issues.

bertschneider (approved):

PR #687, opened 2023-03-27
junit: Use test instance in autofuzz

bertschneider:
Autofuzz did not use the current test instance to invoke the test
function, rather it created it's own ones. This conflicts with other
JUnit plugins which manipulate the test instance, e.g. inject 
objects or execute other setup code.


src/main/java/com/code_intelligence/jazzer/autofuzz/FuzzTarget.java
```diff
@@ -61,7 +63,7 @@ public static void fuzzerInitialize(String[] args) {
           "Expected the argument to --autofuzz to be a method reference (e.g. System.out::println)");
       System.exit(1);
     }
-    methodReference = args[0];
+    String methodReference = args[0];
```
fmeum, line 66: This shadows the instance variable of the same name. Can we remove it?

src/main/java/com/code_intelligence/jazzer/autofuzz/FuzzTarget.java
```diff
@@ -51,8 +52,9 @@ public final class FuzzTarget {
   private static Meta meta;
   private static String methodReference;
   private static Executable[] targetExecutables;
-  private static Map<Executable, Class<?>[]> throwsDeclarations;
-  private static Set<SimpleGlobMatcher> ignoredExceptionMatchers;
+  private static Object targetInstance;
+  private static Map<Executable, Class<?>[]> throwsDeclarations = Collections.emptyMap();
```
fmeum, line 56: Do we need the initializer here? We would probably want these to be set explicitly in any case and could fail instead if they are `null`.

src/main/java/com/code_intelligence/jazzer/autofuzz/FuzzTarget.java
```diff
@@ -210,13 +208,28 @@ public static void fuzzerInitialize(String[] args) {
...
             .collect(Collectors.toMap(method
                 -> method,
                 method
                 -> Stream.concat(Arrays.stream(method.getExceptionTypes()), alwaysIgnore.stream())
                        .toArray(Class[] ::new)));
+
+    fuzzerInitialize(methodReference, new Meta(targetClass), targetExecutables, null);
+  }
+
+  public static void fuzzerInitialize(
```
fmeum, line 223: Since `fuzzerInitialize` has a very specific signature, could we call this something else instead, e.g. just `initialize` or `setTarget`?

We could also add a comment explaining that this is currently only called from JUnit (and internally in this class).

src/main/java/com/code_intelligence/jazzer/autofuzz/FuzzTarget.java
```diff
@@ -288,7 +307,8 @@ private static void fuzzerTestOneInput(
       Throwable cause = e.getCause();
       Class<?> causeClass = cause.getClass();
       // Do not report exceptions declared to be thrown by the method under test.
-      for (Class<?> declaredThrow : throwsDeclarations.get(targetExecutable)) {
+      for (Class<?> declaredThrow :
+          throwsDeclarations.getOrDefault(targetExecutable, new Class[0])) {
```
fmeum, line 311: I would prefer explicilty setting `throwsDeclaration` to `new Class[0]` in the initialize functions - we could then also add a comment explaining why this is the right choice for JUnit.

src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetHolder.java
```diff
@@ -22,25 +22,22 @@
...
-              ()
-                  -> {
-                com.code_intelligence.jazzer.autofuzz.FuzzTarget.fuzzerInitialize(
-                    Opt.targetArgs.toArray(new String[0]));
-                return null;
-              },
-              Optional.empty());
+      Method fuzzTestOneInput = com.code_intelligence.jazzer.autofuzz.FuzzTarget.class.getMethod(
+          "fuzzerTestOneInput", FuzzedDataProvider.class);
+      return new FuzzTargetHolder.FuzzTarget(fuzzTestOneInput, newInstance, Optional.empty());
```
fmeum, line 29: `fuzzerTestOneInput` instead of `fuzzTestOneInput` for consistency.

src/main/java/com/code_intelligence/jazzer/autofuzz/FuzzTarget.java
```diff
@@ -210,13 +208,28 @@ public static void fuzzerInitialize(String[] args) {
...
                 -> method,
                 method
                 -> Stream.concat(Arrays.stream(method.getExceptionTypes()), alwaysIgnore.stream())
                        .toArray(Class[] ::new)));
+
+    fuzzerInitialize(methodReference, new Meta(targetClass), targetExecutables, null);
+  }
+
+  public static void fuzzerInitialize(
+      String methodReference, Meta meta, Executable[] targetExecutables, Object targetInstance) {
```
fmeum, line 224: We could check that all `targetExecutables` are defined in the same class and then use that class to construct the `Meta` object. This would reduce the parameter count and potentially catch implementation bugs.

src/main/java/com/code_intelligence/jazzer/autofuzz/FuzzTarget.java
bertschneider: Renamed the shadowed variables.

src/main/java/com/code_intelligence/jazzer/autofuzz/FuzzTarget.java
bertschneider: They are not needed in the JUnit case, at least not now, and initializing them in the overloaded `fuzzerInitialize` function seemed unnecessary complex. Nevertheless, I moved the initialization there to see how it would look like. WDYT?

src/main/java/com/code_intelligence/jazzer/autofuzz/FuzzTarget.java
bertschneider: I like `setTarget` changed.

src/main/java/com/code_intelligence/jazzer/autofuzz/FuzzTarget.java
bertschneider: Done.

src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetHolder.java
bertschneider: Done.

src/main/java/com/code_intelligence/jazzer/autofuzz/FuzzTarget.java
fmeum: Looks good!

fmeum: Is the `throwsDeclarations` initializer still needed?

fmeum (approved):

PR #689, opened 2023-03-29
mutator: secure enum mutator read

bertschneider:


src/main/java/com/code_intelligence/jazzer/mutation/combinator/MutatorCombinators.java
```diff
@@ -232,7 +232,7 @@ public static SerializingMutator<Integer> mutateIndices(int length) {
     return new SerializingMutator<Integer>() {
       @Override
       public Integer read(DataInputStream in) throws IOException {
-        return in.readInt() % length;
+        return Math.abs(in.readInt()) % length;
```
fmeum, line 235: You could use `Math#floorMod` instead.

bertschneider: Done.

fmeum (approved):

PR #690, opened 2023-03-29
map mutator improvements

0xricksanchez:
This changes the map size handling similar to what was done in #675 to prevent the oscillation around `maxSize` by doing a fair random pick of the mutation operator

### DIFF
--- a/src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
+++ b/src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
@@ -33,8 +33,12 @@ import java.io.IOException;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.AnnotatedType;
 import java.util.AbstractMap;
+import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
@@ -96,7 +100,8 @@ final class MapMutatorFactory extends MutatorFactory {
       for (int i = 0; i < size; i++) {
         map.put(keyMutator.read(in), valueMutator.read(in));
       }
-      // Wrap in an immutable view for additional protection against accidental mutation in fuzz
+      // Wrap in an immutable view for additional protection against accidental
+      // mutation in fuzz
       // tests.
       return toImmutableMapView(map);
     }
@@ -112,7 +117,8 @@ final class MapMutatorFactory extends MutatorFactory {
 
     @Override
     protected Map<K, V> makeDefaultInstance() {
-      // Wrap in an immutable view for additional protection against accidental mutation in fuzz
+      // Wrap in an immutable view for additional protection against accidental
+      // mutation in fuzz
       // tests.
       return toImmutableMapView(new LinkedHashMap<>(maxInitialSize()));
     }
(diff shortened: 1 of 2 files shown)


src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
```diff
@@ -126,32 +128,56 @@ public void initInPlace(Map<K, V> reference, PseudoRandom prng) {
       }
     }
 
+    private static enum Action {
```
bertschneider, line 131: Add `Chunk` operations as well? I kind of like those.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
```diff
@@ -126,32 +130,105 @@ public void initInPlace(Map<K, V> reference, PseudoRandom prng) {
       }
     }
 
+    private void eraseRandomChunk(Map<K, V> map, PseudoRandom prng, int mapSize) {
+      int upperLimit = Math.max(1, Math.min(mapSize - minSize, mapSize / 2));
+      int chunkSize = upperLimit == 1 ? 1 : prng.closedRange(1, upperLimit);
+      int chunkOffset = prng.closedRange(0, mapSize - chunkSize);
+      int elementsRemoved = 0;
+      while (elementsRemoved < chunkSize) {
+        map.entrySet()
```
fmeum, line 139: This iterates over the map in every iteration. Instead, you could collect the keys to remove in a temporary list and then remove them via `map.keySet().removeAll()`.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
```diff
@@ -126,32 +130,101 @@ public void initInPlace(Map<K, V> reference, PseudoRandom prng) {
       }
     }
 
+    private void eraseRandomChunk(Map<K, V> map, PseudoRandom prng, int mapSize) {
+      int upperLimit = Math.max(1, Math.min(mapSize - minSize, mapSize / 2));
+      int chunkSize = upperLimit == 1 ? 1 : prng.closedRange(1, upperLimit);
+      int chunkOffset = prng.closedRange(0, mapSize - chunkSize);
+      List<K> tmpRemStore = new ArrayList<>(chunkSize);
+      map.entrySet().stream().skip(chunkOffset).limit(chunkSize).forEach(entry -> {
```
fmeum, line 138: Map iteration order will in general be non-deterministic. We could force the maps we generated to be `LinkedHashMap`s, which would make the mutations deterministic. 

@bertschneider What do you think?

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
```diff
@@ -126,32 +130,101 @@ public void initInPlace(Map<K, V> reference, PseudoRandom prng) {
...
+        tmpRemStore.add(entry.getKey());
```
fmeum, line 139: You can use `.collect(Collectors.toCollection(new ArrayList<>(chunkSize)))` instead of creating the list and then using `forEach`.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
```diff
@@ -126,32 +130,101 @@ public void initInPlace(Map<K, V> reference, PseudoRandom prng) {
       }
     }
 
+    private void eraseRandomChunk(Map<K, V> map, PseudoRandom prng, int mapSize) {
+      int upperLimit = Math.max(1, Math.min(mapSize - minSize, mapSize / 2));
```
fmeum, line 134: This can also use `clamp`.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
```diff
@@ -126,32 +130,101 @@ public void initInPlace(Map<K, V> reference, PseudoRandom prng) {
...
+      }
+    }
+
+    private void changeRandomChunk(Map<K, V> map, PseudoRandom prng, int mapSize) {
+      int chunkOffset = prng.indexIn(mapSize);
+      int chunkSize =
+          Math.min(prng.closedRange(1, mapSize - chunkOffset), (int) Math.ceil(mapSize / 10.0));
+      int elementsChanged = 0;
+      while (elementsChanged < chunkSize) {
+        map.entrySet()
```
fmeum, line 157: This is also potentially quadratic in the map size. You can collect the entries to mutate once in a temporary structure and then mutate them in a loop.

fmeum, line 157: This is also potentially quadratic in the map size. You can collect the entries to mutate once in a temporary structure and then mutate them in a loop.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
bertschneider: True, `LinkedHashMap` would help with ordering and shouldn't introduce too much of a performance penalty. I think it's a good idea to switch.

0xricksanchez: @fmeum @bertschneider Is that something to be done in this PR or is that "meta" enough to extract this into a separate one?

fmeum: I can do it as a follow-up.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
```diff
@@ -153,12 +159,16 @@ private void changeRandomChunk(Map<K, V> map, PseudoRandom prng, int mapSize) {
...
+      Iterator<Map.Entry<K, V>> iterator = map.entrySet().iterator();
+      for (int i = 0; i < chunkOffset; i++) {
+        if (iterator.hasNext()) {
+          iterator.next();
+        }
+      }
+
+      while (iterator.hasNext() && elementsChanged < chunkSize) {
+        Map.Entry<K, V> entry = iterator.next();
+        mutateElement(entry, map, prng);
```
fmeum, line 171: You can't (safely) change a map while iterating over it in this way: https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html#entrySet--:~:text=If%20the%20map%20is%20modified%20while%20an%20iteration%20over%20the%20set%20is%20in%20progress%20(except%20through%20the%20iterator%27s%20own%20remove%20operation%2C%20or%20through%20the%20setValue%20operation%20on%20a%20map%20entry%20returned%20by%20the%20iterator)%20the%20results%20of%20the%20iteration%20are%20undefined.

You can collect a list of keys in a preparatory step and then just call `mutateElement` on the corresponding entry.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
```diff
@@ -126,32 +132,107 @@ public void initInPlace(Map<K, V> reference, PseudoRandom prng) {
...
+        initElement(map, prng);
+      }
+    }
+
+    private void changeRandomChunk(Map<K, V> map, PseudoRandom prng, int mapSize) {
+      int chunkOffset = prng.indexIn(mapSize);
+      int chunkSize =
+          Math.min(prng.closedRange(1, mapSize - chunkOffset), (int) Math.ceil(mapSize / 10.0));
+
+      List<K> keys = new ArrayList<>(map.keySet());
```
fmeum, line 162: You don't need to copy all keys, just the keys from `chunkOffset` to `chunkOffset + chunkSize` would suffice.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
```diff
@@ -126,32 +132,107 @@ public void initInPlace(Map<K, V> reference, PseudoRandom prng) {
...
+      int elementsChanged = 0;
+
+      for (int i = chunkOffset; i < chunkOffset + chunkSize && i < keys.size(); i++) {
```
fmeum, line 165: Is the comparison with `keys.size()` necessary? `chunkOffset + chunkSize` should be at most the initial map size and `keys` doesn't change in this loop.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
```diff
@@ -127,32 +132,110 @@ public void initInPlace(Map<K, V> reference, PseudoRandom prng) {
...
+
+    private void changeRandomChunk(Map<K, V> map, PseudoRandom prng) {
+      int mapSize = map.size();
+      int chunkOffset = prng.indexIn(mapSize);
+      int chunkSize =
+          Math.min(prng.closedRange(1, mapSize - chunkOffset), (int) Math.ceil(mapSize / 10.0));
+
+      List<K> keys = new ArrayList<>(
+          map.keySet().stream().skip(chunkOffset).limit(chunkSize).collect(Collectors.toList()));
+      int elementsChanged = 0;
```
fmeum, line 166:
```suggestion
```

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
```diff
@@ -127,32 +132,110 @@ public void initInPlace(Map<K, V> reference, PseudoRandom prng) {
...
+    }
+
+    private void changeRandomChunk(Map<K, V> map, PseudoRandom prng) {
+      int mapSize = map.size();
+      int chunkOffset = prng.indexIn(mapSize);
+      int chunkSize =
+          Math.min(prng.closedRange(1, mapSize - chunkOffset), (int) Math.ceil(mapSize / 10.0));
+
+      List<K> keys = new ArrayList<>(
+          map.keySet().stream().skip(chunkOffset).limit(chunkSize).collect(Collectors.toList()));
```
fmeum, line 165:
```suggestion
      List<K> keys = map.keySet().stream().skip(chunkOffset).limit(chunkSize).collect(Collectors.toList());
```

+ reformat

fmeum (approved):

PR #695, opened 2023-04-04
Print invalid corpus file warning for mutator framework

bertschneider:
The experimental mutator framework can detect if a given corpus input could be read cleanly. A failure indicates a changed fuzz test signature, which could lead to issues reproducing prior findings. 

In this case a warning is printed in fuzzing and regression mode.


src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetRunner.java
```diff
@@ -166,7 +170,21 @@ private static int runOne(long dataPtr, int dataLength) {
       //  1. Let libFuzzer pass in a null pointer if the byte array hasn't changed since the last
       //     call to our custom mutator and skip the read entirely.
       //  2. Implement a InputStream backed by Unsafe to avoid the copyToArray overhead.
-      mutator.read(new ByteArrayInputStream(copyToArray(dataPtr, dataLength)));
+      byte[] buf = copyToArray(dataPtr, dataLength);
+      boolean readCompletely = mutator.read(new ByteArrayInputStream(buf));
```
fmeum, line 174: I haven't looked into the implementation yet, but is this `readCompletely` or rather `readExactly`?

src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetRunner.java
```diff
@@ -166,7 +170,21 @@ private static int runOne(long dataPtr, int dataLength) {
...
+
+      // Detect special libFuzzer inputs which can not be processed by the mutator framework.
+      boolean libFuzzerIndicator =
+          buf.length == 0 || (buf.length == 1 && buf[0] == LIBFUZZER_END_OF_CORPUS);
+
+      // If the mutator framework is used all constructed inputs are valid and read completely.
+      // Existing corpus files may not be valid for the current fuzz target anymore, though. In
+      // this case, print a warning once.
+      if (!(invalidCorpusFilesPresent || libFuzzerIndicator || readCompletely)) {
```
fmeum, line 183: `invalidCorpusFilesPresent` seems to be the flag that ensures we only warn once. Maybe call it `invalidCorpusFileWarningShown`? That still reads well in `invalidCorpusFilesPresent()`, but is less confusing here.

src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetRunner.java
```diff
@@ -166,7 +170,21 @@ private static int runOne(long dataPtr, int dataLength) {
...
       //  2. Implement a InputStream backed by Unsafe to avoid the copyToArray overhead.
-      mutator.read(new ByteArrayInputStream(copyToArray(dataPtr, dataLength)));
+      byte[] buf = copyToArray(dataPtr, dataLength);
+      boolean readCompletely = mutator.read(new ByteArrayInputStream(buf));
+
+      // Detect special libFuzzer inputs which can not be processed by the mutator framework.
+      boolean libFuzzerIndicator =
+          buf.length == 0 || (buf.length == 1 && buf[0] == LIBFUZZER_END_OF_CORPUS);
+
+      // If the mutator framework is used all constructed inputs are valid and read completely.
```
fmeum, line 180: I would find this easier to understand as "All inputs constructed by the mutator framework are valid and read completely.". 

What is "valid" referring to in this context? Aren't we using "read completely/exactly" as a proxy for that?

src/main/java/com/code_intelligence/jazzer/mutation/support/InputStreamSupport.java
```diff
@@ -193,5 +193,67 @@ public void close() throws IOException {
     }
   }
 
+  /**
+   * Wraps a given stream with the functionality to detect if it was consumed exactly.
+   * To do so, the stream must provide an accurate implementation of {@link
```
fmeum, line 198: Should we enforce this by limiting it to accepting `ByteArrayInputStream`s? For input streams doing actual IO we can't really guarantee this property.

examples/junit/src/test/java/com/example/MutatorFuzzTest.java
```diff
@@ -0,0 +1,34 @@
...
+import com.code_intelligence.jazzer.junit.FuzzTest;
+import com.code_intelligence.jazzer.mutation.annotation.NotNull;
+import java.util.List;
+
+class MutatorFuzzTest {
+  @FuzzTest
+  void mutatorFuzz(List<@NotNull String> list) {
+    assertEquals(FuzzTargetRunner.mutatorDebugString(), "Arguments[Nullable<List<String>>]");
+    if (list != null && list.size() > 3 && list.get(2).equals("mutator")) {
+      throw new IllegalStateException("Found expected JUnit mutator test issue");
```
fmeum, line 33: I would recommend using a different exception type as this one may be reasonably be thrown as the result of a bug in our mutator framework.

src/main/java/com/code_intelligence/jazzer/junit/FuzzTestArgumentsProvider.java
```diff
@@ -128,7 +132,10 @@ private Stream<? extends Arguments> adaptInputsForFuzzTest(
         Object[] args;
         if (argumentsMutator.isPresent()) {
           ArgumentsMutator mutator = argumentsMutator.get();
-          mutator.read(new ByteArrayInputStream(e.getValue()));
+          boolean readCompletely = mutator.read(new ByteArrayInputStream(e.getValue()));
+          if (!readCompletely) {
```
fmeum, line 136: This could also be called `readExactly`.

src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetRunner.java
```diff
@@ -166,7 +170,21 @@ private static int runOne(long dataPtr, int dataLength) {
       //  1. Let libFuzzer pass in a null pointer if the byte array hasn't changed since the last
       //     call to our custom mutator and skip the read entirely.
       //  2. Implement a InputStream backed by Unsafe to avoid the copyToArray overhead.
-      mutator.read(new ByteArrayInputStream(copyToArray(dataPtr, dataLength)));
+      byte[] buf = copyToArray(dataPtr, dataLength);
+      boolean readCompletely = mutator.read(new ByteArrayInputStream(buf));
+
+      // Detect special libFuzzer inputs which can not be processed by the mutator framework.
+      boolean libFuzzerIndicator =
+          buf.length == 0 || (buf.length == 1 && buf[0] == LIBFUZZER_END_OF_CORPUS);
```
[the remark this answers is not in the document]
bertschneider: Good idea. Done.

src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetRunner.java
bertschneider: Renamed.

src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetRunner.java
bertschneider: Rephrased the comment a little, better?

examples/junit/src/test/java/com/example/MutatorFuzzTest.java
bertschneider: Done.

src/main/java/com/code_intelligence/jazzer/junit/FuzzTestArgumentsProvider.java
bertschneider: Renamed.

fmeum (approved):

src/main/java/com/code_intelligence/jazzer/mutation/support/InputStreamSupport.java
bertschneider: I didn't want to change the signature of `ArgumentsMutator#read`, which would be necessary to do that. Come to think of it, `ArgumentsMutator` hides the restriction and could provide an incorrect return value. We could pull out the `extendWithReadExactly` to the call sites. WDYT?

fmeum: We could also restrict the signature of `read` to `ByteArrayInputStream` for now and consider alternatives when we want to pass in other types of streams.

bertschneider: It's not as nice as accepting `InputStream` but for sure the easier solution for now. Done.

fmeum (approved):

PR #703, opened 2023-04-12
mutation: Simplify the `detach` concept and make lists mutable

fmeum:
Detaching confusingly meant two things at once, but failed at both:
1. It was meant to rid fuzz test arguments of state shared with the mutator so that fuzz test body and future mutations couldn't affect each other. This wasn't working correctly for lists, which provided a shallowly immutable view on potentially mutable contents.
2. It also promised "mutable values where possible", which wasn't really well-defined and also doesn't match common use cases: If a function accepts a `List<Foo>`, it should generally be able to expect the list to permit modifications.

Instead, `detach` now always returns an object that shares no mutable state, not even transitively, with the original object. Apart from simplifying the implementation, this allows mutators to copy values by applying `detach`, which previously could have resulted in shared state that couldn't be mutated independently.

Since we are currently reading in the fuzz test arguments from the raw byte representation in every iteration, there is no need to detach and we can also return mutable lists directly. If we get rid of the re-reads, we would have to detach, which is now verified by a new consistency check in `ArgumentsMutator`.

We might be able to optimize e.g. `List<String>` to avoid all copies in the future via a new annotation (e.g. `@Immutable`). The current implementation would already avoid copying the strings and would only recreate the list, which is most likely negligible overhead.


src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/ListMutatorFactory.java
```diff
@@ -100,7 +94,7 @@ public void initInPlace(List<T> reference, PseudoRandom prng) {
 
     @Override
     public void mutateInPlace(List<T> reference, PseudoRandom prng) {
-      List<T> list = underlyingMutableList(reference);
+      List<T> list = reference;
```
0xricksanchez, line 97: Is having the local variable really necessary now?

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/ListMutatorFactory.java
```diff
@@ -83,15 +79,13 @@ public void write(List<T> list, DataOutputStream out) throws IOException {
...
-      // tests.
-      return toImmutableListView(new ArrayList<>(maxInitialSize()));
+      return new ArrayList<>(maxInitialSize());
     }
 
     @Override
     public void initInPlace(List<T> reference, PseudoRandom prng) {
       int targetSize = prng.closedRange(minInitialSize(), maxInitialSize());
-      List<T> list = underlyingMutableList(reference);
+      List<T> list = reference;
```
0xricksanchez, line 88: See local variable comment below

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
```diff
@@ -117,16 +114,15 @@ public void write(Map<K, V> map, DataOutputStream out) throws IOException {
...
+      // Use a LinkedHashMap to ensure deterministic iteration order, which makes chunk-based
+      // mutations deterministic. The additional overhead compared to HashMap should be minimal.
+      return new LinkedHashMap<>(maxInitialSize());
     }
 
     @Override
     public void initInPlace(Map<K, V> reference, PseudoRandom prng) {
       int targetSize = prng.closedRange(minInitialSize(), maxInitialSize());
-      Map<K, V> map = underlyingMutableMap(reference);
+      Map<K, V> map = reference;
```
0xricksanchez, line 125: see above

0xricksanchez (commented):
Looks good to me. Just minor clarifications needed. Overall this simplifies the list/map handling quite a bit :)

0xricksanchez (approved):

PR #704, opened 2023-04-12
mutation: Improve list adapter for repeatable fields

fmeum:
Ensures that individual bulk operations never require more than linear time.

Also improves the test by asserting that the underlying builder is modified rather than just the view.

### DIFF
--- a/src/main/java/com/code_intelligence/jazzer/mutation/mutator/proto/BuilderAdapters.java
+++ b/src/main/java/com/code_intelligence/jazzer/mutation/mutator/proto/BuilderAdapters.java
@@ -16,38 +16,87 @@
 
 package com.code_intelligence.jazzer.mutation.mutator.proto;
 
-import static java.lang.String.format;
+import static java.util.Collections.singletonList;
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/code_intelligence/jazzer/mutation/mutator/proto/BuilderAdapters.java
```diff
@@ -16,64 +16,130 @@
...
+
+      // O(size() + other.size())
+      public boolean addAll(int index, Collection<? extends U> other) {
+        // This was benchmarked against the following implementation and found to be faster in all
+        // cases (up to 4x on lists of size 1000):
+        //
+        // for (U element : other) {
+        //   builder.addRepeatedField(field, element);
+        // }
+        // Collections.rotate(subList(index, size()), other.size());
```
0xricksanchez, line 71: I like the perf annotations a lot! That should make things easier when trying to improve the mutator framework in the long run :).

0xricksanchez (approved):

PR #706, opened 2023-04-13
mutation: Improve ListMutator

fmeum:
Adds chunk-based mutations to the list mutator mimicking fuzztest's mutation.

The size of the chunks is chosen according to a Zipf distribution, which ensures that chunks tend to be small (and often have length 1), which should result in more stable mutations and also makes separate individual element mutations unnecessary.

### DIFF
--- a/src/main/java/com/code_intelligence/jazzer/mutation/engine/SeededPseudoRandom.java
+++ b/src/main/java/com/code_intelligence/jazzer/mutation/engine/SeededPseudoRandom.java
@@ -19,6 +19,7 @@ package com.code_intelligence.jazzer.mutation.engine;
 import static com.code_intelligence.jazzer.mutation.support.Preconditions.require;
 
 import com.code_intelligence.jazzer.mutation.api.PseudoRandom;
+import com.code_intelligence.jazzer.mutation.support.Preconditions;
 import com.code_intelligence.jazzer.mutation.support.RandomSupport;
 import java.util.List;
 import java.util.SplittableRandom;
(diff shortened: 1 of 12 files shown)


src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/ChunkMutations.java
```diff
@@ -0,0 +1,112 @@
...
+
+    list.subList(chunkOffset, chunkOffset + chunkSize).clear();
+  }
+
+  static <T> void insertRandomChunk(
+      List<T> list, int maxSize, SerializingMutator<T> elementMutator, PseudoRandom prng) {
+    Preconditions.require(list.size() < maxSize);
+
+    int chunkSize = prng.closedRangeBiasedTowardsSmall(1, maxSize - list.size());
+    int chunkOffset = prng.closedRange(0, list.size());
```
0xricksanchez, line 47: nit: calling list.size() 3 times in a row

0xricksanchez (approved):

src/main/java/com/code_intelligence/jazzer/mutation/engine/SeededPseudoRandom.java
```diff
@@ -213,4 +214,57 @@ public double closedRange(double lowerInclusive, double upperInclusive) {
   public void bytes(byte[] bytes) {
     RandomSupport.nextBytes(random, bytes);
   }
+
+  @Override
+  public int closedRangeBiasedTowardsSmall(int upperInclusive) {
+    if (upperInclusive == 0) {
+      return 0;
+    }
+    Preconditions.require(upperInclusive >= 0);
```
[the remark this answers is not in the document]
fmeum: It is still needed as the parameter could be negative, but I made it more precise (`>` instead of `>=`).

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/ChunkMutations.java
fmeum: Fixed.

PR #708, opened 2023-04-13
Initialize Android Runtime env before starting fuzzers.

marktefftech:


src/main/java/com/code_intelligence/jazzer/Jazzer.java
```diff
@@ -226,7 +230,7 @@ private static String prepareArgv0(Map<String, String> additionalEnvironment) th
                      .map(e -> e.getKey() + "='" + e.getValue() + "'")
                      .collect(joining(" "));
     String command = Stream
-                         .concat(Stream.of(javaBinary().toString()), javaBinaryArgs())
```
TheCoryBarker, line 229: Why change this?

src/main/java/com/code_intelligence/jazzer/Jazzer.java
```diff
@@ -236,11 +240,15 @@ private static String prepareArgv0(Map<String, String> additionalEnvironment) th
 
     // argv0 is printed by libFuzzer during reproduction, so have the launcher basename contain
     // "jazzer".
+    String launcherContent;
```
TheCoryBarker, line 243: is this already defined a couple lines up?

src/main/java/com/code_intelligence/jazzer/Jazzer.java
```diff
@@ -249,21 +257,28 @@ private static String prepareArgv0(Map<String, String> additionalEnvironment) th
...
   }
 
   private static Stream<String> javaBinaryArgs() {
-    Stream<String> stream = Stream.of("-cp", System.getProperty("java.class.path"),
+    Stream<String> stream;
+    if (isAndroid()) {
+      stream = Stream.of("app_process", "-Djdk.attach.allowAttachSelf=true", "/system/bin", Jazzer.class.getName());
+      // ManagementFactory wont work with Android
+      return stream;
+    } else {
```
TheCoryBarker, line 280: nit: this else isn't needed

src/main/java/com/code_intelligence/jazzer/Jazzer.java
marktefftech: Changing this because Android doesn't actually execute a Java binary, and `javaBinary()` returns a `Path` that is then converted to a string here.

Since we execute a binary named `app_process`, we simply need to return the linux execute command `exec`.

src/main/java/com/code_intelligence/jazzer/Jazzer.java
marktefftech: Yes, fixed.

src/main/java/com/code_intelligence/jazzer/Jazzer.java
marktefftech: Updated.

PR #709, opened 2023-04-14
bazel: Fix macOS 13.3 build

fmeum:


br-lewis:
It initially didn't work but that seemed to be due to zlib being in a different path than the one you put. I fixed it locally and pushed the change. Hopefully that also works for @0xricksanchez

fmeum:
Thanks for the fix. Now this just needs an approval :-)

br-lewis (approved):

PR #710, opened 2023-04-14
Change junit integration to look use different directories for each test method

br-lewis:
This changes jazzer to create and look for crash files in different directories for each tested method. It adds a new `<test name>` directory to the searched `resources/<classpath>/<class name>Inputs` and `.cifuzz-corpus/<class name>/<test name>` paths so each test has its own separated crash files.

Now inputs are gathered from:
- non-recursively getting all files in `resources/<classpath>/<class name>Inputs`
- recursively getting all files under `resources/<classpath>/<class name>Inputs/<test name>`

Classes can share common test cases between tests while allowing each test to have its own individual inputs. The inputs under `.cifuzz-corpus` do not have this shared vs non-shared inputs behavior because libfuzzer doesn't support that.

I also wrote up some docs on how jazzer works with junit which could use a look over since there are some parts that I'm not sure about it which should be clear in the text. I'm also not sure if it should be in a separate markdown file or as part of the `FuzzTest` javadoc comment.


src/main/java/com/code_intelligence/jazzer/junit/FuzzTestArgumentsProvider.java
```diff
@@ -155,42 +175,122 @@ private Stream<? extends Arguments> adaptInputsForFuzzTest(
...
+   * @param path the path to search to files
+   * @return a stream of all files (without directories) within {@code path}. If {@code path} is
+   *     {@code null}, then an
+   * empty stream is returned.
+   * @throws IOException can be thrown by the underlying call to {@link Files#find}
+   */
+  private static Stream<Map.Entry<String, byte[]>> walkClassInputs(Path path) throws IOException {
+    // this check is technically redundant thanks to the null check near the start of `walkInputs`
+    // however since these are only run once per tested method, I think it shouldn't degrade
+    // performance too much
```
fmeum, line 269: Instead of performing these checks, you could also try not checking and instead catching `FileNotFoundException`. Not that it would present any realistic problems here, but checking file existence separately from the access to the file could lead to unexpected behavior when the file is deleted between the two operations.

src/test/java/com/code_intelligence/jazzer/junit/TestMethod.java
```diff
@@ -0,0 +1,52 @@
...
+import java.lang.reflect.Method;
+
+/**
+ * Small class that allows us to capture the methods that we're using as test data. We need similar
+ * but slightly different data at various points:
+ * 1. the method name with parameters for finding the method initially and for referring to it in
+ * JUnit
+ * 2. the method name without parameters for the findings directories
+ */
+public class TestMethod {
```
fmeum, line 30: That's great cleanup work! :-)

docs/junit-integration.md
```diff
@@ -0,0 +1,91 @@
...
+
+Because these same files and functions are involved in two mostly separate sets of functionality, this will look at the
+flow of the different methods involved in integrating with JUnit in fuzzing mode (when `JAZZER_FUZZ` is set to any
+non-empty value) and in regression mode (when `JAZZER_FUZZ` is not set) separately.
+
+# Fuzzing Flow
+
+## `evaluateExecutionCondition`
+
+It appears that, when running an entire class of tests, JUnit will first call this function on all tests marked
```
fmeum, line 18: That's correct.

docs/junit-integration.md
```diff
@@ -0,0 +1,91 @@
...
+fuzzing.
+
+Crashes are saved in `resources/<package>/<test file name>Inputs/<test method name>` and results that are interesting to
+libfuzzer are saved in `.cifuzz-corpus` though where that's done is a bit of a mystery to me.
+
+# Regression Flow
+
+## `evaluateExecutionCondition`
+
+Similar to the first execution of this in fuzzing mode, this will look for the first fuzz test to return that it's
```
fmeum, line 49: In regression test mode we should be running all fuzz tests (or have a bug somewhere). Since regression test doesn't run libFuzzer, we don't have to worry about its global state.

docs/junit-integration.md
```diff
@@ -0,0 +1,91 @@
...
+  under it are used as inputs for only the test of the same name.
+
+JUnit will use the file's name as the name of the test case for its reporting. It also accepts .jar files where it will
+search with the given directory in the jar.
+
+### CIFuzz Corpus
+
+The corpus kept in `.cifuzz-corpus/<test class name>/<test method name>` holds any inputs that libfuzzer found worth
+saving and not necessarily just inputs that caused a crash. I'm not entirely sure what the criteria for this are but
+this appears to be something that's entirely managed by libfuzzer's code and not Jazzer. Unlike with the resources test
```
fmeum, line 79: That's correct, Jazzer only puts anything there through libFuzzer.

docs/junit-integration.md
br-lewis: I had the usage of this method by JUnit wrong. It's called once on the test itself to determine if the test should be run at all and if so then it will be called again for each input it receives from `provideArguments`. So it goes:

1. `evaluateExecutionCondition` to determine if this test is enabled at all
2. `provideArguments` to get the set of arguments that this test should be called with
3. For each argument set from `provideArguments`, `evaluateExecutionCondition` is called again to check if that particular argument set is enabled.

In fuzzing mode, that would go:
```
fuzzTest1 - enabled
fuzzTest1 Fuzzing... - enabled
fuzzTest2 - disabled
fuzzTest3 - disabled
```
and in regression:
```
fuzzTest1 - enabled
fuzzTest1 <empty input> - enabled
fuzzTest1 crash-1 - enabled
fuzzTest2 - enabled
fuzzTest2 <empty input> - enabled
fuzzTest2 crash-2 - enabled
...
```

Which meant that it looked like it was short circuiting because it called `evaluateExecutionCondition` until it found an enabled test, then called `provideArguments`, then `evaluateExecutionCondition` again but before I wasn't printing enough context to know why those additional calls were happening.

br-lewis:
Something to note here is that this change could be considered breaking in that previously generated corpora will not be used by in fuzzing mode. They will still be used as test cases in regression mode but fuzzing will start from scratch unless the test cases are copied/symlinked into the new test-specific directories.

fmeum (approved):
Thanks for putting so much effort into the docs, including diagrams!

PR #712, opened 2023-04-17
mutator: fix a rare test case

oetr:
The rare case happens for proto float/double fields that are initiated with -0.0/-0.0f; detached to 0.0/0.0f; and mutated back to -0.0/-0.0f. This previously caused a rare random failure of the StressTest.java.

### DIFF
--- a/src/test/java/com/code_intelligence/jazzer/mutation/mutator/StressTest.java
+++ b/src/test/java/com/code_intelligence/jazzer/mutation/mutator/StressTest.java
@@ -456,6 +456,7 @@ public class StressTest {
       testReadWriteExclusiveRoundtrip(mutator, fixedValue);
 
       initValues.add(mutator.detach(value));
+      value = fixFloatingPointsForProtos(value);
 
       for (int mutation = 0; mutation < NUM_MUTATE_PER_INIT; mutation++) {
         Object detachedOldValue = mutator.detach(value);


src/test/java/com/code_intelligence/jazzer/mutation/mutator/StressTest.java
```diff
@@ -456,6 +456,7 @@ void genericMutatorStressTest(AnnotatedType type, String mutatorTree,
       testReadWriteExclusiveRoundtrip(mutator, fixedValue);
 
       initValues.add(mutator.detach(value));
```
fmeum, line 458: Instead of changing `value`, could we instead pass all the arguments to asserts through `fixFloatingPointsForProto`? That could have made this issue more obvious and also ensures that we don't interfere with the mutator - it should be able to assume that it is only passed values that it created itself.

oetr: One problem is that if `value` is `some_field = -0.0`, then mutation of the underlying float/double value starts from `0.0`, so the mutator changes `value` to 0.0 when it converts the proto into float (or double). This seems to happen only because of the generic `Message:toBuilder()` function. Going the usual route `((FloatField3)value).getSomeField()` returns -0.0.
Our modification of `value` in the test mirrors this  proto-related modification and changes it to `some_field = 0.0` (or actually empty string because it's the default value).

The other problem is that when `detachedOldValue` is 0.0 and the mutated `value` is -0.0, `fixFloatingPointsForProto` changes `value` to 0.0 and the assertion fails as well.

Any ideas?

fmeum: Actually no ;-) This bug is plain terrible to work around. I will ping upstream and we can leave this as is.

fmeum (approved):

PR #713, opened 2023-04-18
mutation: Use chunk-based mutation approach for Map 

fmeum:
The previous Map mutator performed both single element as well as chunk
mutations, which isn't necessary now that we can bias chunks towards
being small.

The old implementation didn't check that `initInPlace` and `read` obeyed
the specified `minSize`. Since non-zero `minSize`s can be problematic in
`read`, they are not supported for now.

There were also a few issues with key mutations: Instead of retrying key
creation separately for each key, which can result in a large amount of
wasted attempts if the key set is almost saturated, all such attempts
now share a single failure counter. Keys are now properly detached
before being mutated, which is necessary if map keys are mutable.

The fallback to value mutation if key mutation failed also didn't
behave correctly: It called `setValue` on a newly allocated `Map.Entry`,
which is a noop.

### DIFF
--- a/src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/ChunkMutations.java
+++ b/src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/ChunkMutations.java
@@ -18,21 +18,29 @@ package com.code_intelligence.jazzer.mutation.mutator.collection;
 
 import com.code_intelligence.jazzer.mutation.api.PseudoRandom;
 import com.code_intelligence.jazzer.mutation.api.SerializingMutator;
+import com.code_intelligence.jazzer.mutation.api.ValueMutator;
 import com.code_intelligence.jazzer.mutation.support.Preconditions;
 import java.util.AbstractList;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
 
 // Based on (Apache-2.0)
 // https://github.com/google/fuzztest/blob/f81257ed70ec7b9c191b633588cb6e39c42da5e4/fuzztest/internal/domains/container_mutation_helpers.h
 final class ChunkMutations {
+  private static final int MAX_FAILED_INSERTION_ATTEMPTS = 100;
+
   private ChunkMutations() {}
 
   static <T> void deleteRandomChunk(List<T> list, int minSize, PseudoRandom prng) {
     int oldSize = list.size();
-    Preconditions.require(oldSize > minSize);
-
     int minFinalSize = Math.max(minSize, oldSize / 2);
     int chunkSize = prng.closedRangeBiasedTowardsSmall(1, oldSize - minFinalSize);
     int chunkOffset = prng.closedRange(0, oldSize - chunkSize);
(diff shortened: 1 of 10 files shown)


fmeum:
I submitted https://github.com/bazel-contrib/rules_jvm/pull/175 to get better logs and noticed that there is an OOM issue that I will need to investigate.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/ChunkMutations.java
```diff
@@ -58,15 +86,120 @@ static <T> void insertRandomChunk(
...
+    };
+    Supplier<K> nextCandidate = () -> {
+      // Mutate the next candidate in the queue.
+      K candidate = keyMutator.mutate(unboxNull(keysToMutate.removeFirst()), prng);
+      keysToMutate.addFirst(boxNull(candidate));
+      return candidate;
+    };
+
+    growBy(map.keySet(), addIfNew, chunkSize, nextCandidate);
+    // Remove the original keys that were successfully mutated into new keys.
```
bertschneider, line 151: Isn't the current behavior that the amount of added keys is removed from the map, regardless if a specific key could be mutated into a valid new one?

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
```diff
@@ -91,6 +93,12 @@ public Optional<SerializingMutator<?>> tryCreate(AnnotatedType type, MutatorFact
       require(minSize <= maxSize,
           format("WithSize#min %d needs to be smaller or equal than WithSize#max %d", minSize,
               maxSize));
+
+      // TODO: Add support for min > 0 to map. If min > 0, then #read can fail to construct
+      //       sufficiently many distinct keys, but the mutation framework currently doesn't offer
+      //       a way to handle this situation gracefully. It is also not clear what behavior users
+      //       could reasonably expect in this situation in both regression test and fuzzing mode.
+      check(minSize == 0, "@WithSize#min != 0 is not yet supported for Map");
```
bertschneider, line 101: Remove `minSize` check above.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/ChunkMutations.java
```diff
@@ -40,6 +50,24 @@ static <T> void deleteRandomChunk(List<T> list, int minSize, PseudoRandom prng)
     list.subList(chunkOffset, chunkOffset + chunkSize).clear();
   }
 
+  static <T> void deleteRandomChunk(Collection<T> collection, int minSize, PseudoRandom prng) {
+    int oldSize = collection.size();
+    Preconditions.require(oldSize > minSize);
```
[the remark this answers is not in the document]
fmeum: Good point. I checked that removing them would still run into failed preconditions further down, so there isn't much value in keeping them around.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/ChunkMutations.java
fmeum: Since the old keys are kept in the map, all added keys are automatically valid new keys. I added this to the comment, please take another look and let me know whether you agree.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
fmeum: Done.

bertschneider (approved):
LGTM

PR #725, opened 2023-04-24
driver: Support build with clang-16

bertschneider:


src/main/native/com/code_intelligence/jazzer/driver/fuzz_target_runner.h
```diff
@@ -17,6 +17,7 @@
 #pragma once
 
 #include <jni.h>
+#include <stdint.h>
```
fmeum, line 20: This header seems to be unused here. You could also drop `jni.h`.
```suggestion
```

fmeum (approved):

PR #728, opened 2023-04-25
mutation: Add support for `DynamicMessage `

fmeum:


src/test/java/com/code_intelligence/jazzer/mutation/mutator/StressTest.java
```diff
@@ -302,6 +311,13 @@ public static Stream<Arguments> protoStressTestCases() {
         arguments(new TypeHolder<@NotNull TestProtobuf>() {}.annotatedType(),
             "{Builder.Nullable<Boolean>, Builder.Nullable<Integer>, Builder.Nullable<Integer>, Builder.Nullable<Long>, Builder.Nullable<Long>, Builder.Nullable<Float>, Builder.Nullable<Double>, Builder.Nullable<String>, Builder.Nullable<Enum<Enum>>, Builder.Nullable<{Builder.Nullable<Integer>, Builder via List<Integer>} -> Message>, Builder via List<Boolean>, Builder via List<Integer>, Builder via List<Integer>, Builder via List<Long>, Builder via List<Long>, Builder via List<Float>, Builder via List<Double>, Builder via List<String>, Builder via List<Enum<Enum>>, Builder via List<(cycle) -> Message>, Builder.Map<Integer,Integer>, Builder.Nullable<FixedValue(OnlyLabel)>, Builder.Nullable<{<empty>} -> Message>, Builder.Nullable<Integer> | Builder.Nullable<Long> | Builder.Nullable<Integer>} -> Message",
             manyDistinctElements(), manyDistinctElements()),
+        arguments(
```
bertschneider, line 314: Add a dedicated example or extend `ExperimentalMutatorFuzzer` to show how this would look like in a real test?

bertschneider (approved):
Looks good!

src/test/java/com/code_intelligence/jazzer/mutation/mutator/StressTest.java
fmeum: I added a dedicated example based on `ExperimentalMutatorComplexProtoFuzzer`, could you take another look?

fmeum: On second thought, I'll make it more dynamic than returning an already existing descriptor.

fmeum: @bertschneider I pushed a new version in which I am retrieving the descriptor from a field rather than a method. DynamicMessage doesn't like equal but not identical descriptors, so the factory method approach was error prone.

bertschneider: Ah, I see. The new example is quite dynamic :smile: Looks good!

bertschneider (approved):

PR #730, opened 2023-04-25
Test Android in CI

fmeum:


marktefftech:
I'm looking into this and will circle back around shortly.

marktefftech:
@fmeum Does PR#734 - `Fixed Android build breaks #734` Address your concerns with JDK versioning, or should we revisit this open PR?

fmeum:
@marktefftech I would still like to get the current PR merged if it doesn't break the Android build. Ideally, we wouldn't need a separate `android ` config at all.

.bazelrc
```diff
@@ -40,10 +40,8 @@ build --tool_java_runtime_version=remotejdk_17
 build --experimental_java_classpath=bazel
 
 # Android
-build:android_arm --incompatible_enable_android_toolchain_resolution
-build:android_arm --android_platforms=//:android_arm64
-build:android_arm --copt=-D_ANDROID
-build:android_arm --java_runtime_version=8
```
TheCoryBarker, line 46: When I tested locally, everything worked except this value still needed to be "=local_jdk" to find the correct toolchain

fmeum: Thanks for checking. That probably means that this is blocked onhttps://github.com/bazelbuild/bazel/pull/18262. I will keep it open and we can test it again after that change has landed in Bazel.

marktefftech: @fmeum Thanks so much for following up with this, we tried many toolchain workarounds for `--java_runtime_version` flag but couldn't get it to work. Hopefully the above PR will unblock this feature.

In the meantime, we are still hoping to add a basic CI test to ensure the critical targets that we have will continue to build.

Currently we rely heavily on the following two commands:

```
bazelisk build launcher/android:jazzer_android --config=android_arm
bazelisk build src/main/java/com/code_intelligence/jazzer/android:jazzer_standalone_android --config=android_arm
```

Can you please offer some guidance on how we can incorporate these commands into the current pipeline? Apologies for the rudimentary question, but I don't have much bazel testing experience yet.

We really just need to make sure that those two commands pass to ensure that Android is in a good state in Jazzer. I think it is important to note that even though the below `test` commands still execute as expected and the test passes, trying to build these targets individually will still fail:

```
bazelisk test --config=ci --remote_header=x-buildbuddy-api-key=*** --java_runtime_version=local_jdk_8 --disk_cache=/home/runner/.cache/bazel-disk //launcher/android:jazzer_android --build_tag_filters="-no-linux-jdk8,-no-jdk8" --test_tag_filters="-no-linux-jdk8,-no-jdk8" //...
```

Thanks in advance, I am happy to implement the test, just need some guidance on where to start.

fmeum: @marktefftech Could you try adding `--config=android_arm` as well as the missing label to https://github.com/CodeIntelligenceTesting/jazzer/blob/73e764196a85fa82b3a8b782f6091723ffb4df0f/.github/workflows/run-all-tests.yml#L42?

fmeum:
Superseded by https://github.com/CodeIntelligenceTesting/jazzer/pull/807

PR #731, opened 2023-04-25
Sanitizes launcher tests using AddressSanitizer

hadi88:


launcher/jvm_tooling.cpp
```diff
@@ -64,16 +64,21 @@ std::string getExecutablePath() {
   char buf[655536];
 #if defined(__APPLE__)
   uint32_t buf_size = sizeof(buf);
-  if (_NSGetExecutablePath(buf, &buf_size) != 0) {
+  ssize_t read_bytes = buf_size - 1;
+  bool failed _NSGetExecutablePath(buf, &buf_size) != 0;
```
fmeum, line 68:
```suggestion
  bool failed = (_NSGetExecutablePath(buf, &buf_size) != 0);
```

launcher/jvm_tooling.cpp
```diff
@@ -64,16 +64,21 @@ std::string getExecutablePath() {
   char buf[655536];
 #if defined(__APPLE__)
   uint32_t buf_size = sizeof(buf);
-  if (_NSGetExecutablePath(buf, &buf_size) != 0) {
+  ssize_t read_bytes = buf_size - 1;
+  bool failed = (_NSGetExecutablePath(buf, &buf_size) != 0);
 #elif defined(_WIN32)
-  if (GetModuleFileNameA(NULL, buf, sizeof(buf)) == 0) {
+  ssize_t read_bytes = GetModuleFileNameA(NULL, buf, sizeof(buf));
```
fmeum, line 70: Windows CI is failing with:
```
launcher/jvm_tooling.cpp(70,3): error: unknown type name 'ssize_t'
  ssize_t read_bytes = GetModuleFileNameA(NULL, buf, sizeof(buf));
  ^
```

`ssize_t` isn't available on Windows, but you could use this "polyfill" for it: https://stackoverflow.com/a/35368387/297261

hadi88: Yes. I think we can use the function signature. In this case, `DWORD` is the return type.

fmeum (approved):

PR #735, opened 2023-05-04
mutation: Do not initialize recursive proto fields

fmeum:
When recursive proto fields are initialized, the resulting messages have an expected nesting depth on the order of the inverse of the frequency with which a nullable value is non-null, which tends to run into StackOverflowErrors quickly.

This is fixed by checking whether a given proto field is recursive and if so, only initializing it "layer by layer" in mutations rather than all at once during initialization.


fmeum:
@hadi88 This should fix the SOE you have seen.

PR #736, opened 2023-05-05
junit: Use marker arguments to signal start of fuzzing

fmeum:
This is intended to be a pure refactoring that enables us to allow arbitrary `ArgumentsProvider`s to provide arguments for fuzz tests that are picked up in fuzzing mode.

### DIFF
--- a/src/main/java/com/code_intelligence/jazzer/junit/Utils.java
+++ b/src/main/java/com/code_intelligence/jazzer/junit/Utils.java
@@ -14,18 +14,33 @@
 
 package com.code_intelligence.jazzer.junit;
 
+import static java.util.Arrays.stream;
+import static java.util.Collections.newSetFromMap;
+import static org.junit.jupiter.api.Named.named;
+import static org.junit.jupiter.params.provider.Arguments.arguments;
+
+import com.code_intelligence.jazzer.utils.UnsafeProvider;
+import com.code_intelligence.jazzer.utils.UnsafeUtils;
+import java.lang.invoke.MethodType;
 import java.lang.management.ManagementFactory;
+import java.lang.reflect.Array;
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.lang.reflect.Proxy;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.Duration;
 import java.util.Arrays;
+import java.util.IdentityHashMap;
 import java.util.Optional;
+import java.util.Set;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.ReflectiveInvocationContext;
+import org.junit.jupiter.params.provider.Arguments;
 
 class Utils {
   /**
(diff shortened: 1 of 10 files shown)


src/main/java/com/code_intelligence/jazzer/utils/UnsafeUtils.java
```diff
@@ -0,0 +1,76 @@
...
+    ClassWriter cw = new ClassWriter(0);
+    String superClassName = abstractClass.getName().replace('.', '/');
+    // Only the package of the class name matters, the actual name is generated. defineHiddenClass
+    // requires the package of the new class to match the one of the lookup.
+    String className = UnsafeUtils.class.getPackage().getName().replace('.', '/') + "/Anonymous";
+    cw.visit(Opcodes.V1_8, 0, className, null, superClassName, null);
+    cw.visitEnd();
+
+    try {
+      Optional<Method> defineHiddenClass =
```
bertschneider, line 51: In which cases are these not available?

src/main/java/com/code_intelligence/jazzer/utils/UnsafeUtils.java
```diff
@@ -0,0 +1,76 @@
...
+  public static <T> Class<? extends T> defineAnonymousConcreteSubclass(Class<T> abstractClass) {
+    if (!Modifier.isAbstract(abstractClass.getModifiers())) {
+      throw new IllegalArgumentException(abstractClass + " is not abstract");
+    }
+
+    ClassWriter cw = new ClassWriter(0);
+    String superClassName = abstractClass.getName().replace('.', '/');
+    // Only the package of the class name matters, the actual name is generated. defineHiddenClass
+    // requires the package of the new class to match the one of the lookup.
+    String className = UnsafeUtils.class.getPackage().getName().replace('.', '/') + "/Anonymous";
```
bertschneider, line 46: Don't we have to implement abstract methods to instantiate the subclass? If not, why couldn't the given abstract class be instantiated directly?

src/main/java/com/code_intelligence/jazzer/utils/UnsafeUtils.java
fmeum: Added a comment.

src/main/java/com/code_intelligence/jazzer/utils/UnsafeUtils.java
fmeum: Turns out that we don't have to implement them - they will throw exceptions when called, but they aren't called. `Unsafe#allocateInstance` does not support abstract classes, which is why we need to provide a concrete subclass.

src/main/java/com/code_intelligence/jazzer/junit/Utils.java
```diff
@@ -127,12 +143,89 @@ static boolean permissivelyParseBoolean(String value) {
...
+        return true;
+      }
+      throw new IllegalStateException(
+          "Some, but not all arguments were marked in invocation of " + invocationContext);
+    } else {
+      return false;
+    }
+  }
+
+  private static final ClassValue<Object> uniqueInstanceCache = new ClassValue<Object>() {
```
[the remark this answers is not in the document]
fmeum: It's essentially a fixed version of `WeakConcurrentHashMap<Class<?>, T>` if that were to exist - if the GC collects a class, its entry in the `ClassValue` will go away. Getting this right with a regular `Map` is apparently quite tricky since the value will usually hold a reference to the class.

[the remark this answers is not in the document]
bertschneider: Thanks for the clarification! If the value still holds a strong reference to the class or an instance of it, the class can not be unloaded/gced, can it? Regardless, good to know that `ClassValue` exists :smiley:

src/main/java/com/code_intelligence/jazzer/utils/UnsafeUtils.java
bertschneider: I can't see a comment. Did you push that change?

src/main/java/com/code_intelligence/jazzer/utils/UnsafeUtils.java
bertschneider: I expected some kind of verification error defining, and even instantiating, a class not implementing all abstract methods. The tests show that it works, still strange, though.

bertschneider (approved):
LGTM

src/main/java/com/code_intelligence/jazzer/utils/UnsafeUtils.java
fmeum: Pushed to the wrong fork :-)

PR #741, opened 2023-05-12
junit: Record parameterized test arguments as seeds

fmeum:
The seeds are serialized to files in a temporary directory that is passed to libFuzzer as an additional seed directory.


src/main/java/com/code_intelligence/jazzer/junit/SeedSerializer.java
```diff
@@ -0,0 +1,142 @@
...
+
+  @Override
+  public boolean allReadsValid() {
+    return allReadsValid;
+  }
+
+  @Override
+  public byte[] write(Object[] args) {
+    ByteArrayOutputStream out = new ByteArrayOutputStream();
+    mutator.writeAny(out, args);
```
bertschneider, line 113: Is this safe? Could user provided argument arrays vary in size?

bertschneider (approved):
Looks really nice! 

I wonder if and how we could print a message with instructions on how to add a new corpus entry to the test source file in case a finding is raised. This would probably need some bigger changes in the mutator frameworks, though. 

Could you mention this feature in the quick start section of the readme and update the JUnit integration docs?

src/main/java/com/code_intelligence/jazzer/junit/SeedSerializer.java
fmeum: JUnit checks the parameter count and types, which will catch such issues before they reach this method (I added comments).

fmeum:
We should consider adding code generation in the future (fuzztest has this and Autofuzz does as well).

I added another test that exercises the `byte[]` signature case with argument conversion. Even with this though, we should probably wait for the entire mutator feature to land without a flag before we add this to the quick start section. When we do, we can even consider switching to recommending this as the primary way to add seeds.

bertschneider (approved):

PR #744, opened 2023-05-24
junit: Generate temporary seeds with deterministic names

fmeum:
This removes a potential source of non-deterministic behavior while fuzzing with a fixed `-seed`.

### DIFF
--- a/src/main/java/com/code_intelligence/jazzer/junit/FuzzTestExecutor.java
+++ b/src/main/java/com/code_intelligence/jazzer/junit/FuzzTestExecutor.java
@@ -33,7 +33,11 @@ import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
+import java.util.Base64;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


src/main/java/com/code_intelligence/jazzer/junit/FuzzTestExecutor.java
```diff
@@ -201,8 +205,21 @@ static FuzzTestExecutor fromContext(ExtensionContext extensionContext) {
   }
 
   public void addSeed(byte[] bytes) throws IOException {
-    Path seed = Files.createTempFile(javaSeedsDir, "seed", null);
-    Files.write(seed, bytes);
+    Path tmpSeed = Files.createTempFile(javaSeedsDir, "tmp-seed-", null);
```
bertschneider, line 208: Should we clean up these tmp files? Something like `seed.toFile().deleteOnExit();`?

bertschneider (approved):
Just one remark, but overall this LGTM.

src/main/java/com/code_intelligence/jazzer/junit/FuzzTestExecutor.java
fmeum: It's moved below, so there is no need to clean it up.

bertschneider: `javaSeedsDir` is already temporary, didn't notice that.

PR #746, opened 2023-05-25
Refactor and centralize configuration handling

br-lewis:
This refactors how we handle configuration options to allow all options to be specified by CLI args, environment variables, or manifest file options. It replaces `Opt` with a new class `Config` in the `utils` package with the idea that we can remove calls to `setProperty` from the codebase with references to variables under `Config`, similar to `Opt` but also handling the setting of the values in addition to the reading of them.

There are a few things which I didn't have time to do:
- [ ] remove all calls to `setProperty`. The remaining ones are mostly in the junit integration.
- [ ] switch `ConfigItem` to use `Optional` instead of needing to check `isSet`
- [ ] we discussed further centralizing the places that modify properties to allow `Config` to set properties once and then never allow them to change
- [ ] the [changes in `getLibFuzzerSeed`](https://github.com/CodeIntelligenceTesting/jazzer/pull/746/files#diff-34d7892aaf9629e082ad974e6bb389057d339e583d4abbda79152cca4cbf898aR257-R264) need to be removed because the `api` package should have have any dependency on the rest of jazzer
- [ ] `Opt`, `OptParser`, and `OptTest` all still exist but currently only reference each other
- [ ] `ConfigItem.StrList` does not handle escaping on the strings (see `OptTest` for the relevant code bits)
- [ ] `ConfigItemTest` does not test each type to ensure they're serialized and deserialized as expected
- [ ] I created but did not implement a test in `ConfigTest` to make sure that loading from a manifest file works. Though this might be unnecessary as other tests will use manifest files to set configuration options
- [ ] I left some useless code commented out that should be removed

Other notes:
- `FuzzTargetRunner` must only be loaded after a call to `Config.loadConfig`. This appears to be case currently but it would be nice to make this more explicit if possible.

### DIFF
--- a/src/main/java/com/code_intelligence/jazzer/driver/Driver.java
+++ b/src/main/java/com/code_intelligence/jazzer/driver/Driver.java
@@ -109,9 +109,6 @@ public class Driver {
       args.add(getDefaultRssLimitMbArg());
     }
 
-    // Do not modify properties beyond this point, loading Opt locks in their values. The agent will
-    // cause Opt to be loaded again, this time in the bootstrap class loader, but since all its
-    // fields are immutable that should not cause confusion.
     AgentInstaller.install(Config.hooks.get());
 
     if (!Config.instrumentOnly.get().isEmpty()) {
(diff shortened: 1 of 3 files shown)


br-lewis:
Closing in favor of #767

PR #751, opened 2023-06-01
Log coverage input directories on startup

zgtm:


bertschneider (approved):

src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetRunner.java
```diff
@@ -367,6 +367,14 @@ public static int startLibFuzzer(List<String> args) {
       // https://github.com/llvm/llvm-project/blob/da3623de2411dd931913eb510e94fe846c929c24/compiler-rt/lib/fuzzer/FuzzerFlags.def#L19
       args.add("-len_control=100");
     }
+
+    for (String arg : args.subList(1, args.size())) {
+      if (arg.length() > 0 && arg.charAt(0) != '-') {
```
fmeum, line 372: Assuming that empty args should never appear in practice, this could be easier to read:
```suggestion
      if (!arg.startsWith("-")) {
```

src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetRunner.java
```diff
@@ -367,6 +367,14 @@ public static int startLibFuzzer(List<String> args) {
...
+        System.out.print("INFO: seed corpus: input file: ");
```
fmeum, line 373: The arguments may be directories, not just "input files". This also usually includes the generated corpus (the first non-dash arg), which is not generally called "seed corpus". Is this output format something we could still change?

zgtm: Sure, we can still change that! Do you have a proposal for the format?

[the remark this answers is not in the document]
zgtm: Ahh, good point, that will also log to the correct output stream (stderr)!

fmeum: Maybe `"using input(s) from: " + arg`?

That avoids fuzzing terminology as well as having to determine whether this points to a file or a directory.

@bertschneider What do you think?

bertschneider: Fine by me :+1:

zgtm: Done! I removed the parentheses for easier parsing.

src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetRunner.java
```diff
@@ -368,6 +368,15 @@ public static int startLibFuzzer(List<String> args) {
       // https://github.com/llvm/llvm-project/blob/da3623de2411dd931913eb510e94fe846c929c24/compiler-rt/lib/fuzzer/FuzzerFlags.def#L19
       args.add("-len_control=100");
     }
+
+    for (String arg : args.subList(1, args.size())) {
+      if (!arg.startsWith("-")) {
+        String message = "using inputs from: ";
+        message += arg;
+        Log.info(message);
```
fmeum, line 376:
```suggestion
        Log.info("using inputs from: " + arg);
```

fmeum: @zgtm Small cleanup, PTAL

zgtm: Ah, yes, thanks for the hint … I have written too much rust it seems! ;)

PR #752, opened 2023-06-02
all: Only load Opt in the system class loader

fmeum:
We can detect whether we are running on Android via system properties and thus don't need to have this bit passed in via a flag.

As a result, the runtime no longer depends on Opt, which is thus no longer loaded in the bootstrap class loader in addition to the system class loader.

### DIFF
--- a/src/main/java/com/code_intelligence/jazzer/BUILD.bazel
+++ b/src/main/java/com/code_intelligence/jazzer/BUILD.bazel
@@ -117,6 +117,7 @@ java_library(
     ],
     deps = [
         "//src/main/java/com/code_intelligence/jazzer/driver",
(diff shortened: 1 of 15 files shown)


bertschneider (commented):
You missed one `jazzer.android` in `Jazzer#isAndroid`. Besides that, this LGTM.

bertschneider (approved):
LGTM

fmeum:
CC @marktefftech

marktefftech:
Overall this change LGTM-- @TheCoryBarker and myself will need to update our open PR's (PR#737, PR#743) with similar changes. Cory, please confirm.

FWIW I think it makes sense to submit this change before the above PR's so we can rebase accordingly.

marktefftech (approved):

TheCoryBarker:
LGTM

PR #757, opened 2023-06-06
mutator: Set required fields to defaults when reading incomplete proto

fmeum:
A proto message deserialized from bytes can be incomplete, which results in an error when building it if required fields are missing. To prevent this, walk the proto and initialize all unset required fields to defaults recursively.


fmeum:
FYI @hadi88

PR #759, opened 2023-06-06
Replace `@DescriptorSource` with `@WithDefaultInstance`

fmeum:
A default instance is a more direct representation of a protobuf base message and getting it via a method allows for more flexible use, e.g., by making the choice of proto type dependent on command-line arguments.


fmeum:
@hadi88 Could you verify that this is sufficiently flexible for your needs?

hadi88:
Looks good to me! Thank you!

PR #761, opened 2023-06-09
Honor explicitly stated corpus directory

bertschneider:
The JUnit integration always created corpus files in `.cifuzz-corpus` subdirectories, even if an explicit corpus directory was stated via command line arguments. 

Now explicit corpus parameters are honored.


src/main/java/com/code_intelligence/jazzer/junit/FuzzTestExecutor.java
```diff
@@ -94,7 +95,16 @@ public static FuzzTestExecutor prepare(ExtensionContext context, String maxDurat
     ArrayList<String> libFuzzerArgs = new ArrayList<>();
     libFuzzerArgs.add(argv0);
 
-    // Store the generated corpus in a per-class directory under the project root, just like cifuzz:
+    // Add passed in corpus directories (and files) at the beginning of the arguments list,
+    // as those have precedence over the JUnit specific ones added later on.
```
fmeum, line 99: "precedence" is a bit vague. It could make sense to explain that the first directory is where libFuzzer stores new inputs, whereas all the other ones just supply additional seeds and aren't written into - their particular order shouldn't really matter.

src/main/java/com/code_intelligence/jazzer/junit/FuzzTestExecutor.java
```diff
@@ -94,7 +95,16 @@ public static FuzzTestExecutor prepare(ExtensionContext context, String maxDurat
...
+    List<String> corpusDirs = originalLibFuzzerArgs.stream()
```
fmeum, line 100: Instead of mutating the list further down, you could use [`groupingBy`](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#groupingBy-java.util.function.Function-) to split the list into two parts.

src/test/java/com/code_intelligence/jazzer/junit/CorpusDirectoryTest.java
```diff
@@ -0,0 +1,183 @@
...
+    Path artifactsDirectory = baseDir.resolve(Paths.get("src", "test", "resources", "com",
+        "example", "CorpusDirectoryFuzzTestInputs", "corpusDirectoryFuzz"));
+    Files.createDirectories(artifactsDirectory);
+
+    // Explicitly stated corpus directory has precedence over the generated one.
+    Path corpusDirectory = baseDir.resolve(Paths.get("corpus"));
+    Files.createDirectories(corpusDirectory);
+
+    // Generated corpus directory should not be used, if explicitly stated one is available.
+    Path generatedCorpus = baseDir.resolve(
```
fmeum, line 80: Maybe call this `defaultGeneratedCorpus` and the one above `explicitGeneratedCorpus`? Both would contain generated entries, but only one is used for this purpose here.

src/test/java/com/code_intelligence/jazzer/junit/CorpusDirectoryTest.java
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  @Test
+  public void fuzzingDisabled() throws IOException {
+    assumeTrue(System.getenv("JAZZER_FUZZ").isEmpty());
+
+    Path corpusDirectory = baseDir.resolve(Paths.get("corpus"));
+    Files.createDirectories(corpusDirectory);
+    // noinspection ResultOfMethodCallIgnored
+    corpusDirectory.resolve("corpus_entry").toFile().createNewFile();
```
fmeum, line 135: Would `Files.createFile()` work here?  We could stay in `Path` land that way.

fmeum (approved):

src/test/java/com/code_intelligence/jazzer/junit/CorpusDirectoryTest.java
bertschneider: Yes! Changed.

src/test/java/com/code_intelligence/jazzer/junit/CorpusDirectoryTest.java
bertschneider: Good idea, renamed.

src/main/java/com/code_intelligence/jazzer/junit/FuzzTestExecutor.java
bertschneider: I tried that out originally, but didn't like the resulting code. Mutating the original list seemed simpler than handling the map returned by `groupingBy`.

src/main/java/com/code_intelligence/jazzer/junit/FuzzTestExecutor.java
bertschneider: Rephrased the comment. Is it clearer now?

fmeum (approved):

PR #763, opened 2023-06-12
Ignore JetBrains classes during instrumentation

bertschneider:
Some JetBrains products interact badly with Jazzer's agent / instrumentation, e.g. their coverage agent. It's reasonable to ignore all JetBrains classes.

### DIFF
--- a/src/main/java/com/code_intelligence/jazzer/agent/RuntimeInstrumentor.kt
+++ b/src/main/java/com/code_intelligence/jazzer/agent/RuntimeInstrumentor.kt
@@ -181,7 +181,10 @@ class RuntimeInstrumentor(
             .acceptClasses(className)
             .scan()
             .use {
-                it.getClassInfo(className).resource.load()
+                it.getClassInfo(className)?.resource?.load() ?: run {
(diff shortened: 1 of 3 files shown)


src/main/java/com/code_intelligence/jazzer/agent/RuntimeInstrumentor.kt
```diff
@@ -181,7 +181,10 @@ class RuntimeInstrumentor(
             .acceptClasses(className)
             .scan()
             .use {
-                it.getClassInfo(className).resource.load()
+                it.getClassInfo(className)?.resource?.load() ?: run {
```
fmeum, line 184: Could you mention this change in the commit message or split it out into a separate commit?

fmeum (approved):

src/main/java/com/code_intelligence/jazzer/agent/RuntimeInstrumentor.kt
bertschneider: done.

PR #780, opened 2023-06-22
Add new ANDROID_ONLY compatibility option

TheCoryBarker:
This is to correct the Android build targets compatibility configuration. This makes the Android build targets only compatible when building for Android target.

This will also add a check that will fail with a message when building for Android on a non-linux host machine. This is because:
1. Android modules aren't supported on Windows
2. the genrule command needed for jvmti.h doesn't work on Mac

### DIFF
--- a/.github/workflows/run-all-tests.yml
+++ b/.github/workflows/run-all-tests.yml
@@ -39,7 +39,6 @@ jobs:
           - os: ubuntu-20.04
             arch: "linux"
             cache: "/home/runner/.cache/bazel-disk"
-            bazel_args: "//launcher/android:jazzer_android"
           - os: ubuntu-20.04
             jdk: 20
             # Workaround for https://github.com/bazelbuild/bazel/issues/14502
--- a/launcher/android/BUILD.bazel
+++ b/launcher/android/BUILD.bazel
@@ -1,4 +1,4 @@
-load("//bazel:compat.bzl", "SKIP_ON_WINDOWS")
+load("//bazel:compat.bzl", "ANDROID_ONLY", "SKIP_ON_WINDOWS")
 
 android_library(
     name = "jazzer_android_lib",
@@ -7,7 +7,7 @@ android_library(
         "//src/main/java/com/code_intelligence/jazzer/android:jazzer_standalone_android.apk",
     ],
     tags = ["manual"],
-    target_compatible_with = SKIP_ON_WINDOWS,
+    target_compatible_with = ANDROID_ONLY,
 )
 
 android_binary(
--- a/src/main/native/com/code_intelligence/jazzer/driver/android_tooling.cpp
+++ b/src/main/native/com/code_intelligence/jazzer/driver/android_tooling.cpp
@@ -47,8 +47,7 @@ Java_com_code_1intelligence_jazzer_android_AndroidRuntime_registerNatives(
       dlsym(handle, "registerFrameworkNatives"));
   const char *dlsym_error = dlerror();
   if (dlsym_error) {
-    std::cerr << "ERROR: Unable to invoke registerFrameworkNatives."
-              << std::endl;
+    std::cerr << "ERROR: Unable to find registerFrameworkNatives." << std::endl;
     exit(1);
   }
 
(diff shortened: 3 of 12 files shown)


TheCoryBarker:
With this change ```//launcher/android:jazzer_android``` can only be built with ```--config=android_arm``` - until we can get rid of --config=android_arm

TheCoryBarker:
I also updated the CI test for Android to work with this new configuration

br-lewis:
I see that the springboot example is still failing. Are you working on that or do need some assistance from our side on it?

TheCoryBarker:
Thanks @br-lewis, looks like I finally got it.

I ended up also needing to fix some of the Android build targets, and update the build test for Android.

src/main/native/com/code_intelligence/jazzer/driver/android_tooling.cpp
```diff
@@ -12,7 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#if defined(_ANDROID)
```
fmeum, line 15: Assuming this file is only ever useful on Android, why do we need this define? If the file is compiled on other platforms, we could instead try to prevent that.

br-lewis:
@TheCoryBarker have you had a chance to address @fmeum's comments?

.github/workflows/run-all-tests.yml
```diff
@@ -79,6 +78,10 @@ jobs:
         run: .github/scripts/echoBuildBuddyConfig.sh ${{ secrets.BUILDBUDDY_API_KEY }} >> $GITHUB_ENV
         shell: bash
 
+      - name: Build Andorid
+        if: contains(matrix.os, 'ubuntu')
+        run: bazelisk build //launcher/android:jazzer_android --config=android_arm
```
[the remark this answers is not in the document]
TheCoryBarker: Removed the Android CI test for now so these fixes can get in

src/main/native/com/code_intelligence/jazzer/driver/android_tooling.cpp
TheCoryBarker: update this to not be included in non-Android builds

fmeum:
@br-lewis CI runs from forks don't have the CI Sense access token that is needed for the CI Fuzz Action. Can we configure the action so that it always passes in that case?

.github/workflows/run-all-tests.yml
```diff
@@ -39,7 +39,6 @@ jobs:
           - os: ubuntu-20.04
             arch: "linux"
             cache: "/home/runner/.cache/bazel-disk"
-            bazel_args: "//launcher/android:jazzer_android"
```
fmeum, line 42: Are we still building the Android targets in CI without this?

launcher/android/BUILD.bazel
```diff
@@ -7,15 +7,15 @@ android_library(
...
+    target_compatible_with = ANDROID_ONLY,
 )
 
 android_binary(
     name = "jazzer_android",
     manifest = ":android_manifest",
     min_sdk_version = 26,
     tags = ["manual"],
-    target_compatible_with = SKIP_ON_WINDOWS,
+    target_compatible_with = ANDROID_ONLY,
```
fmeum, line 18: Since this is a top-level android_binary target, does it need this restriction? Without it, we wouldn't have to worry about this being run in CI.

fmeum (approved):

br-lewis:
@TheCoryBarker pulling in the recent changes should fix the CI failure

.github/workflows/run-all-tests.yml
TheCoryBarker: no we wont be building the Android targets without this. This change is correcting the Android build, but does break this CI. Android build wont be able to be built now without the --config=android_arm flag

launcher/android/BUILD.bazel
TheCoryBarker: android_binary wont build on Windows, so it's good to have the SKIP_ON_WINDOWS, do you think I should change it back to that?

fmeum: That's a good idea, please do.

PR #783, opened 2023-06-23
Separate Sanitizers for Android

TheCoryBarker:
This is to create a separate list of Android sanitizers. These sanitizers will be packaged up with the rest of the sanitizers for a normal Jazzer build, and then the user can specify to use these sanitizers at instrumentation time with the flag ```--use_android_sanitizers```. This flag will be used for Android for offline instrumentation, like so:

``` > ./jazzer --instrument_only=<fuzz_target>.jar --dump_classes_dir=/tmp --use_android_sanitizers```

and then you'll have a jar file that is instrumented for Android. The Jazzer built for Android wont need all of the sanitizers, so only the Android sanitizers will be included in the Android build.

The reasons to keep these lists separate are:
- Some sanitizers will cause crashes on Android if used on Android code (RegexInjection.java) so we want to make sure these are excluded for all Android runs. There may be more in the future that are added that need to not be added to Android
- ALthough the one sanitizer I added specific for Android (converted SQL Injection sanitizer) wont cause any issues for a normal JVM if it is used at runtime, it is possible a future one will cause issues, so it's best to just keep them on a separate list of sanitizers

### CODE REVIEW

File: src/main/java/com/code_intelligence/jazzer/driver/Opt.java [hunk 1]
```diff
@@ -149,9 +149,8 @@ public final class Opt {
   public static final boolean dedup =
       boolSetting("dedup", hooks, "Compute and print a deduplication token for every finding");
 
-  public static final String androidBootclassJarPath = stringSetting("android_bootclass_jar_path",
```
TheCoryBarker, line 152: This is not used so I removed it

sanitizers/BUILD.bazel
```diff
@@ -7,9 +7,9 @@ java_library(
 )
 
 java_library(
-    name = "offline_only_sanitizers",
```
TheCoryBarker, line 10: This is also dead code from before sanitizers were injected into bootstrap classloader on Android

TheCoryBarker: ... the offline_only_sanitizers is dead code, but the new android_sanitizers is needed

TheCoryBarker:
> This looks fine to me, but I wonder whether we really need this to be configurable at runtime. Could we instead get by with `select` on `@platforms//os:android` and, if necessary, a [`bool_flag`](https://github.com/bazelbuild/bazel-skylib/blob/288731ef9f7f688932bd50e704a91a45ec185f9b/rules/common_settings.bzl#L48)?
> 
> We already have more runtime flags than we should, which makes testing more difficult.

I like this idea, I'll make this change

PR #805, opened 2023-07-24
Add fuzzing hooks for clojure stdlib functions

zgtm:
This PR adds the capability to hook functions written in clojure.

Since calling clojure functions just calles the general `clojure.lang.IFn.invoke()`, we don't know what function we're dealing with when hooking this function call.

In order to know the function name, we also hook the call to `clojure.lang.Var.getRawRoot()` call, which does return information about the function name (at least for non-anonymous functions). We mark the functions objects we're interested in, by adding them to a `WeakHashMap markedObjects` and looking them up on `clojure.lang.IFn.invoke()`.


fmeum:
From the perspective of maintenance, it would be great if you could, now that we see the hooks work, port them over to use only the [public API for hooks](https://codeintelligencetesting.github.io/jazzer-docs/jazzer-api/com/code_intelligence/jazzer/api/Jazzer.html) and move them to the `sanitizers` directory.

If any public API is missing (e.g. for integer compares), let's add it.

zgtm:
> From the perspective of maintenance, it would be great if you could, now that we see the hooks work, port them over to use only the [public API for hooks](https://codeintelligencetesting.github.io/jazzer-docs/jazzer-api/com/code_intelligence/jazzer/api/Jazzer.html) and move them to the `sanitizers` directory.
> 
> If any public API is missing (e.g. for integer compares), let's add it in a separate commit.

Yes, I'll try to do that!

For the integer comparisons I did not write any new hooks, but just used existing ones. Should those still be copied to `sanitizers/` or can they stay where they are right now?

src/main/java/com/code_intelligence/jazzer/runtime/TraceCmpHooks.java
```diff
@@ -69,6 +73,14 @@ final public class TraceCmpHooks {
     TraceDataFlowNativeCallbacks.traceCmpLong(thisObject, (long) arguments[0], hookId);
   }
 
+  @MethodHook(type = HookType.BEFORE, targetClassName = "clojure.lang.Numbers", targetMethod = "isZero")
+  @MethodHook(type = HookType.BEFORE, targetClassName = "clojure.lang.Numbers", targetMethod = "isPos")
+  @MethodHook(type = HookType.BEFORE, targetClassName = "clojure.lang.Numbers", targetMethod = "isNeg")
+  public static void staticLongCompareZero(
```
fmeum, line 79: You can keep these hooks here, they are very simple and similar to existing hooks.

src/main/java/com/code_intelligence/jazzer/runtime/TraceCmpHooks.java
```diff
@@ -69,6 +73,14 @@ final public class TraceCmpHooks {
...
+          MethodHandle method, Object thisObject, Object[] arguments, int hookId) {
+    TraceDataFlowNativeCallbacks.traceCmpLong((long) arguments[0], (long) 0, hookId);
```
fmeum, line 81: Every hook should either use `targetMethodDescriptor` to specify the exact function signature it matches *or* check the types of arguments before casting. Doing neither can result in `ClassCastException`s if unexpected (or new) overloads are invoked.

src/main/java/com/code_intelligence/jazzer/runtime/TraceDivHooks.java
```diff
@@ -35,6 +35,7 @@ final public class TraceDivHooks {
       targetMethod = "divideUnsigned", targetMethodDescriptor = "(JJ)J")
   @MethodHook(type = HookType.BEFORE, targetClassName = "java.lang.Long",
       targetMethod = "remainderUnsigned", targetMethodDescriptor = "(JJ)J")
+  @MethodHook(type = HookType.BEFORE, targetClassName = "clojure.lang.Numbers", targetMethod = "divide")
```
fmeum, line 38: See above, please add `targetMethodDescriptor`.

src/main/java/com/code_intelligence/jazzer/runtime/ClojureLangHooks.java
```diff
@@ -0,0 +1,47 @@
...
+package com.code_intelligence.jazzer.runtime;
+
+import com.code_intelligence.jazzer.api.HookType;
+import com.code_intelligence.jazzer.api.MethodHook;
+import java.lang.invoke.MethodHandle;
+import java.util.*;
+
+@SuppressWarnings("unused")
+final public class ClojureLangHooks {
+    static ThreadLocal<WeakHashMap<Object, String>> markedObjects = ThreadLocal.withInitial(() -> new WeakHashMap<Object, String>());
```
fmeum, line 24: Note that using a `WeakHashMap` is only correct if the key type does not override `equals`. Could you add a comment pointing to the definition of the type we expect to be stored in this map?

src/main/java/com/code_intelligence/jazzer/runtime/ClojureLangHooks.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @MethodHook(type = HookType.AFTER, targetClassName = "clojure.lang.Var", targetMethod = "getRawRoot")
+    public static void clojureMarkContains(MethodHandle method, Object thisObject, Object[] arguments, int hookId, Object result) {
+        if ("clojure.string$includes_QMARK_".equals(result.getClass().getCanonicalName())
```
fmeum, line 28: A `static final HashSet` would be cleaner here, probably also more performant.

src/main/java/com/code_intelligence/jazzer/runtime/ClojureLangHooks.java
```diff
@@ -0,0 +1,47 @@
...
+                || "clojure.string$index_of_QMARK_".equals(result.getClass().getCanonicalName())
+                || "clojure.string$last_index_of_QMARK_".equals(result.getClass().getCanonicalName())) {
+            markedObjects.get().put(result, result.getClass().getCanonicalName());
+        }
+    }
+
+    @MethodHook(type = HookType.BEFORE, targetClassName = "clojure.lang.IFn", targetMethod = "invoke")
+    public static void clojureMarkedContains(
+            MethodHandle method, Object thisObject, Object[] args, int hookId) {
+        if (markedObjects.get().containsKey(thisObject)) {
```
fmeum, line 40: Are we using the value associated with the key in any way? If not, you can use `Collections.newSetFromMap` to craft a `WeakHashSet`.

src/main/java/com/code_intelligence/jazzer/runtime/ClojureLangHooks.java
fmeum, line 24: This could use a more descriptive name such as `stringContainsFuncs`.

zgtm: Yes, will do!

src/main/java/com/code_intelligence/jazzer/runtime/ClojureLangHooks.java
zgtm: Not for now, but I could imagine that in the future we would like to hook different types of functions with different hook-behaviour. In this case, storing the name of the function in the HashMap would be helpful.

src/main/java/com/code_intelligence/jazzer/runtime/ClojureLangHooks.java
zgtm: … or we use different `HashMaps` / `HashSets` for different functions! :D

src/main/java/com/code_intelligence/jazzer/runtime/TraceCmpHooks.java
fmeum: They aren't overloaded *yet* - if new versions of Clojure are released, this might change and we don't want code to break until Jazzer is updated. Adding a `targetMethodDescriptor` ensures that this won't happen and has no runtime cost.

That said, if the signature really is `(java.lang.Number)Z`, then the casts to `(long)` aren't correct if the argument is e.g. an `int`. You can cast to `Number` instead and use `longValue`. 

Also please add a test that exercises these hooks with different types. Instead of compiling Clojure, which may be tedious to set up, you can add the clojure stdlib as a Maven dep and just call the Java functions directly.

zgtm: Ah, yes, thanks for the hint!

I've now moved these hooks into their own functions concerning `java.lang.Number` objects. A test is added too.

src/main/java/com/code_intelligence/jazzer/runtime/TraceDivHooks.java
zgtm: I have added the `targetMethodDescriptor`  for all methods _except for_ the `clojure.lang.Util.eqiv` hook (`genericStaticEquals`) because here I think we actually want to [hook everything](https://github.com/clojure/clojure/blob/master/src/jvm/clojure/lang/Util.java).

src/main/java/com/code_intelligence/jazzer/runtime/ClojureLangHooks.java
zgtm: Done!

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ClojureLangHooks.java
```diff
@@ -0,0 +1,66 @@
...
+   * Actual hook for fuzzer guidance relying on identified objects by the
+   * clojure.lang.Var.getRawRoot hook.
+   */
+  @MethodHook(type = HookType.BEFORE, targetClassName = "clojure.lang.IFn", targetMethod = "invoke")
+  public static void clojureMarkedContains(
+      MethodHandle method, Object thisObject, Object[] args, int hookId) {
+    if (stringContainsFuncs.get().contains(thisObject)) {
+      try {
+        Jazzer.guideTowardsContainment((String) args[0], (String) args[1], hookId);
+      } catch (ClassCastException ignored) {
```
fmeum, line 62: This is a bit of a code smell as it could hide bugs. Do you know when we expect this to be thrown? Could we get around this by checking that both objects have the same class?

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ClojureLangHooks.java
```diff
@@ -0,0 +1,66 @@
...
+  /**
+   * This hook checks the type of the returned clojure.lang.IFn objects and puts them into
+   * stringContainsFuncs if they match know string-contains functions.
+   */
+  @MethodHook(
+      type = HookType.AFTER, targetClassName = "clojure.lang.Var", targetMethod = "getRawRoot")
+  public static void
+  clojureMarkContains(
+      MethodHandle method, Object thisObject, Object[] arguments, int hookId, Object result) {
+    if ("clojure.string$includes_QMARK_".equals(result.getClass().getCanonicalName())
```
fmeum, line 43: I still think this should be a `Set`, both for readability and (likely, but untested) performance. `Stream.of(...).collect(toSet())` assigned to a `static final` is all the setup that would take.

sanitizers/src/test/java/com/example/ClojureTests.java
```diff
@@ -0,0 +1,40 @@
...
+import clojure.lang.Numbers;
+import com.code_intelligence.jazzer.api.FuzzedDataProvider;
+import java.sql.Connection;
+import java.sql.SQLException;
+import org.h2.jdbcx.JdbcDataSource;
+
+class ClojureFindingException extends Exception {}
+
+public class ClojureTests {
+  static void insecureInsertUser(Long lnumber, Integer inumber, Long divisor) throws Exception {
```
fmeum, line 26: Hello, 1234 5678 12! :⁠-⁠)

src/main/java/com/code_intelligence/jazzer/agent/Agent.kt
```diff
@@ -96,13 +96,13 @@ fun installInternal(
         }
     }
     val includedHookNames = instrumentationTypes
-        .mapNotNull { type ->
```
fmeum, line 99: This diff doesn't seem needed anymore

src/main/java/com/code_intelligence/jazzer/runtime/TraceDivHooks.java
fmeum: I don't see this hook in the diff, am I missing something? There is also `compare`, which can likely be handled similarly.

fmeum (approved):

src/main/java/com/code_intelligence/jazzer/runtime/TraceDivHooks.java
zgtm: Ah, yes! This was already implemented in a [previous PR](https://github.com/CodeIntelligenceTesting/jazzer/pull/765).

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ClojureLangHooks.java
zgtm: Hm, the reason I added this, was that I did not want to throw an exception, when the function is called with something that can not be converted to `String`, but that will still work on the clojure side. In this case we would just loose the guiding towards containment, but not crash with strange `ClassCastException`. Is this reasonable? I can add a comment, too.

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ClojureLangHooks.java
fmeum: Just out of curiosity I created a benchmark for this:

```java
@State(Scope.Benchmark)
@Fork(value = 1, warmups = 1)
public class TestBenchmark {
  private static final Set<String> haystack =
      Stream.of(
              "clojure.string$includes_QMARK_",
              "clojure.string$starts_with_QMARK_",
              "clojure.string$ends_with_QMARK_",
              "clojure.string$index_of_QMARK_",
              "clojure.string$last_index_of_QMARK_")
          .collect(Collectors.toSet());

  @Param({"foobar", "clojure.string$last_index_of_QMARK_", "clojure.string$last_index_of_QMARK__"})
  String needle;

  @Benchmark
  public boolean equalityChecks() {
    return "clojure.string$includes_QMARK_".equals(needle)
        || "clojure.string$starts_with_QMARK_".equals(needle)
        || "clojure.string$ends_with_QMARK_".equals(needle)
        || "clojure.string$index_of_QMARK_".equals(needle)
        || "clojure.string$last_index_of_QMARK_".equals(needle);
  }

  @Benchmark
  public boolean setContains() {
    return haystack.contains(needle);
  }
}
```

```
Benchmark                                                 (needle)   Mode  Cnt          Score          Error  Units
TestBenchmark.equalityChecks                                foobar  thrpt    5  197581255.803 ±  6287389.319  ops/s
TestBenchmark.equalityChecks   clojure.string$last_index_of_QMARK_  thrpt    5  129550623.215 ±  2934774.598  ops/s
TestBenchmark.equalityChecks  clojure.string$last_index_of_QMARK__  thrpt    5  197707791.842 ±  4276520.409  ops/s
TestBenchmark.setContains                                   foobar  thrpt    5  517792671.251 ± 13372898.282  ops/s
TestBenchmark.setContains      clojure.string$last_index_of_QMARK_  thrpt    5  152513807.705 ±  1303357.002  ops/s
TestBenchmark.setContains     clojure.string$last_index_of_QMARK__  thrpt    5  499847231.177 ±  3257284.885  ops/s
```

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ClojureLangHooks.java
fmeum: You can check whether something can be converted to `String` via `instanceof`. We do that in many of the other hooks and if it works here as well, it would be cleaner than catching the exception.

PR #808, opened 2023-08-01
Set fuzzing action to run for the featured project as well

br-lewis:
Switches the fuzzing action to use a matrix to run fuzzing in both our jazzer project and the featured projects one. The project used in `cifuzz.yaml` will remain the same since that would be executing jobs manually and would require the executor be authorized to do run the job.


fmeum:
@br-lewis Our secret lacks permission:
```
Did not succeed uploading the fuzzing artifact: rpc error: code = PermissionDenied desc = Insufficient privilege for project
```

fmeum (approved):

fmeum (approved):

PR #810, opened 2023-08-02
bazel: Remove usages of host JDK

fmeum:
With this change, all tests should pass even without a local JDK.

### DIFF
--- a/deploy/BUILD.bazel
+++ b/deploy/BUILD.bazel
@@ -96,10 +96,11 @@ java_export(
         srcs = [artifact + "_artifact_test.sh"],
         args = [
             "$(rootpath :%s)" % artifact,
+            "$(JAVABASE)",
         ],
         data = [
             ":" + artifact,
-            "@local_jdk//:bin/jar",
+            "@bazel_tools//tools/jdk:current_java_runtime",
         ],
         tags = [
             # Coverage instrumentation necessarily adds files to the jar that we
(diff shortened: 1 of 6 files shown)


deploy/BUILD.bazel
```diff
@@ -96,17 +96,21 @@ java_export(
         srcs = [artifact + "_artifact_test.sh"],
         args = [
             "$(rootpath :%s)" % artifact,
+            "$(JAVABASE)",
         ],
         data = [
             ":" + artifact,
-            "@local_jdk//:bin/jar",
+            "@bazel_tools//tools/jdk:current_host_java_runtime",
```
bertschneider, line 103: I could not find good docs about `@bazel_tools//tools/jdk:current_host_java_runtime`, but expect that it still refers to a locally installed JDK, doesn't it.

What's the difference to `@bazel_tools//tools/jdk:current_java_runtime"?

fmeum: I agree that docs are too sparse on this - I will look into adding some upstream.

Both `current_java_runtime` and `current_host_java_runtime` rely on toolchain resolution to pick an available JDK (which we restrict to `remotejdk_17`, hence we always get a downloaded one). The difference is that `current_java_runtime` selects a runtime that runs on the platform we currently build *for* (which might be Android) whereas `current_host_java_runtime` selects one that runs on the platform we build *on*. 

For tests both platforms usually agree, but it's probably better to switch to `current_java_runtime` here as we run its tools during the test execution, not the build.

bertschneider (approved):

deploy/BUILD.bazel
bertschneider: Thx for the explanation.

PR #819, opened 2023-08-11
junit: Fix CLI runs on individual files 

fmeum:
When all user-provided libFuzzer path arguments are files, the JUnit
executor now correctly starts libFuzzer in reproduction mode by not
adding any directories to the command line. In particular, this fixes
a crash when Jazzer is invoked on a single file to reproduce a
particular finding.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Do not perform LF/CRLF conversion during checkin or checkout.
+* -text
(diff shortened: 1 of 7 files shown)


fmeum:
@br-lewis @bertschneider The finding is an NPE in `Helpers#assertMutator`, which makes me think that the fuzz test may not be executed with the experimental mutator. The experimental mutator would honor the `@NotNull` annotation, Autofuzz doesn.t

fmeum:
This is ready for review again, I fixed the Windows CI issue.

fmeum:
Will do, thanks for the pointer!

bertschneider (approved):
LGTM

PR #821, opened 2023-08-14
Add structured mutator benchmark and fix structured mutator issue

fmeum:


bertschneider:
Benchmarks and output looks good. Is there a way to force a rerun of the benchmarks, though?

fmeum:
> Benchmarks and output looks good. Is there a way to force a rerun of the benchmarks, though?

The individual benchmark runs are deterministic and cached just like any other Bazel test. If you run the `<name>.stats` target, the statistics are computed again based on the (cached or not) logs of the tests. Is there any other kind of rerunning that you had in mind?

bertschneider:
They still should be impacted by the current system load and so on. It's probably neglectable, due to the multiple runs, but I still wondered if I could force a rerun. Normally `run` executes the target again and does not use the cache, right?

fmeum:
> They still should be impacted by the current system load and so on. It's probably neglectable, due to the multiple runs, but I still wondered if I could force a rerun. 

The runtime would definitely be affected by system load, but the benchmark only measures the number of runs it takes to get the finding. We could also measure overall time, but I thought that measuring mutator overhead separately in a JMH benchmark would be be less noisy.

> Normally `run` executes the target again and does not use the cache, right?

Yes, it also does here, but the target we run uses the logs of the tests as inputs and those are cached. If you really want to rerun the tests, you can use `--cache_test_results=no`.
`

bertschneider:
Oh, yes, it measures the number of runs not the time :facepalm: Forget everything I said.

bertschneider (approved):

fmeum:
I will merge the benchmarks now and submit a separate PR that fixes the issues and enables the benchmarks in CI.

PR #824, opened 2023-08-21
Fix subprocess detection

fmeum:


src/main/java/com/code_intelligence/jazzer/Jazzer.java
```diff
@@ -209,6 +208,19 @@ private static SimpleImmutableEntry<String, String> parseSingleArg(String arg) {
...
   }
 
+  /**
+   * Returns whether the given libFuzzer option which isn't enabled by default is enabled by the
+   * given command line.
+   */
+  static boolean isLibFuzzerOptionEnabled(String option, List<String> args) {
+    return args.stream()
+        .filter(arg -> arg.startsWith(String.format("-%s=", option)))
+        .map(arg -> arg.split("=")[1])
```
bertschneider, line 218: This would break with an `ArrayIndexOutOfBounds` exception on invalid arguments like `-merge=`, wouldn't it? Should we provide a better error message?

fmeum: We filtered out arguments that aren't of the form `-option=` in the `filter` above and so `"-option=".split("=")` has at least length 2. I added test coverage for this and also now handle this case just like libFuzzer would.

bertschneider (approved):
LGTM

PR #826, opened 2023-08-22
driver: Enhance method-not-found errors when not using JUnit 5

zgtm:
This adds a new note to the error message that appears when the `@FuzzTest` is used, but JUnit 5 is not set up / the dependency `junit-jupiter-engine` in missing in version 5.

### DIFF
--- a/src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetFinder.java
+++ b/src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetFinder.java
@@ -95,7 +95,8 @@ class FuzzTargetFinder {
             "%s must define exactly one of the following two functions:%n"
                 + "public static void fuzzerTestOneInput(byte[] ...)%n"
                 + "public static void fuzzerTestOneInput(FuzzedDataProvider ...)%n"
-                + "Note: Fuzz targets returning boolean are no longer supported; exceptions should be thrown instead of returning true.",
+                + "Note: Fuzz targets returning boolean are no longer supported; exceptions should be thrown instead of returning true.%n"
+                + "Note: When using the @FuzzTest annotation, you will need to set up JUnit 5, which can be as simple as adding a dependency on org.junit.jupiter:junit-jupiter-engine.",
             clazz.getName()));
       }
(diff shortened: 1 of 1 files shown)


zgtm:
@fmeum: There is a very similar message in `src/main/java/com/code_intelligence/jazzer/replay/Replayer.java` but I was not sure whether we support the `@FuzzTest` annotation for the replayer as well. Should I add the same change, there?

src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetFinder.java
```diff
@@ -95,7 +95,8 @@ private static FuzzTarget findFuzzTargetByMethodName(Class<?> clazz) {
             "%s must define exactly one of the following two functions:%n"
                 + "public static void fuzzerTestOneInput(byte[] ...)%n"
                 + "public static void fuzzerTestOneInput(FuzzedDataProvider ...)%n"
-                + "Note: Fuzz targets returning boolean are no longer supported; exceptions should be thrown instead of returning true.",
+                + "Note: Fuzz targets returning boolean are no longer supported; exceptions should be thrown instead of returning true.%n"
+                + "Note: When using the @FuzzTest annotation, you will need to set up junit5 or add junit-jupiter-engine version 5 as a dependency.",
```
fmeum, line 99: Let's spell it JUnit 5 and refer to `junit-jupiter-engine` via its Maven coordinates `org.junit.jupiter:junit-jupiter-engine` (it's only available in version 5, so we don't have to mention that explicitly).

fmeum:
> @fmeum: There is a very similar message in `src/main/java/com/code_intelligence/jazzer/replay/Replayer.java` but I was not sure whether we support the `@FuzzTest` annotation for the replayer as well. Should I add the same change, there?

No, we do not support the replayer for JUnit 5, `@FuzzTest` has this capability built in.

src/main/java/com/code_intelligence/jazzer/driver/FuzzTargetFinder.java
zgtm: Yes, I have changed the text.

fmeum (approved):

PR #827, opened 2023-08-23
Fix Clojure hooks

fmeum:


fmeum:
@zgtm Could you check whether the other hooks are affected by this problem as well?

fmeum:
@zgtm I reduced the variance in test runtime by adding `-use_value_profile=1`. CI is green now.

zgtm:
> @zgtm Could you check whether the other hooks are affected by this problem as well?

This should also affect the hook `numberUnsignedDivide` in `src/main/java/com/code_intelligence/jazzer/runtime/TraceDivHooks.java`. Though I have not been able  to tweak the test to show different behaviour, here …

zgtm (approved):
Thanks for finding and fixing this! I've added another commit, feel free to squash into your first commit!

fmeum:
> Thanks for finding and fixing this! I've added another commit, feel free to squash into your first commit!

Made the exact same change at a similar time (only with an additional comment) :-)

PR #832, opened 2023-08-28
Add `@FuzzTest#maxExecutions` and improve `LifecycleFuzzTest`

fmeum:


src/main/java/com/code_intelligence/jazzer/junit/FuzzTest.java
```diff
@@ -103,9 +103,22 @@
...
+  /**
+   * If set to a positive number, the fuzz test function will be executed at most this many times
+   * during fuzzing.
+   *
+   * <p>Prefer this element over {@link #maxDuration()} if you want to ensure comparable levels of
+   * fuzzing across machine's with different performance characteristics.
+   *
+   * <p>This option has no effect during regression testing.
+   */
+  long maxExecutions() default 0;
```
fmeum, line 121: Happy to bikeshed this name. I found "runs" to clash with our concept of "fuzzing runs" while "executions" goes well with "exec/s" metrics in libFuzzer output.

bertschneider, line 121: We normally set `-runs=0` to only execute corpus entries. Should the annotation have  a different semantic and couldn't this  be confusing? I'm fine either way, though, but would add `0` as infinite runs to the description.

src/main/java/com/code_intelligence/jazzer/junit/AgentConfiguringArgumentsProvider.java
```diff
@@ -37,7 +37,8 @@ public Stream<? extends Arguments> provideArguments(ExtensionContext extensionCo
     // FIXME(fmeum): Calling this here feels like a hack. There should be a lifecycle hook that runs
     //  before the argument discovery for a ParameterizedTest is kicked off, but I haven't found
     //  one.
-    FuzzTestExecutor.configureAndInstallAgent(extensionContext, fuzzTest.maxDuration());
+    FuzzTestExecutor.configureAndInstallAgent(
```
bertschneider, line 40: Could you please add a comment why the agent needs to be installed here and in `FuzzTestExtensions`? We discussed this, but I forgot again 😬

src/main/java/com/code_intelligence/jazzer/junit/FuzzTest.java
fmeum: We do have the regression test mode for this, so I don't think `-runs=0` has much use with JUnit-based fuzz tests. But it should still be supported (e.g. for use in cifuzz) as command-line arguments take precedence over this value.

I added an explanation of the default case to the docs.

examples/junit/src/test/java/com/example/LifecycleFuzzTest.java
```diff
@@ -16,72 +16,125 @@
...
+      // TODO: Fuzz tests currently don't run before each and after each methods between fuzz test
+      // invocations.
+      expectedEvents.addAll(Collections.nCopies(3, "lifecycleFuzz"));
+      expectedEvents.addAll(expectedAfterEachEvents);
+    }
+
+    expectedEvents.add("afterAll");
+
+    assertThat(events).containsExactlyElementsIn(expectedEvents).inOrder();
     throw new IOException();
```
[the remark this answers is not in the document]
fmeum: I am doing that now and also added a comment explaining why this isn't fully straightforward.

src/main/java/com/code_intelligence/jazzer/junit/AgentConfiguringArgumentsProvider.java
fmeum: Good point, done :-)

bertschneider (approved):
Nice addition!

PR #834, opened 2023-08-28
Fuzz FuzzedDataProviderImpl

br-lewis:
This adds a very large fuzz test for the `FuzzedDataProviderImpl` and modifies building `fuzzed_data_provider.cpp` to make it possible for the normal and shaded `FuzzedDataProvider` to at least load properly by duplicating the selffuzz symbols into `fuzzed_data_provider.cpp` manually.

I also removed the segfault fix I did just to make sure our tooling can catch it again.


selffuzz/src/test/java/com/code_intelligence/selffuzz/driver/FuzzedDataProviderImplFuzzTest.java
```diff
@@ -0,0 +1,433 @@
...
+
+public class FuzzedDataProviderImplFuzzTest {
+  @FuzzTest
+  void fuzzedDataProviderTest(byte @NotNull[] buf) {
+    try (FuzzedDataProviderImpl data = FuzzedDataProviderImpl.withJavaData(buf)) {
+      ArrayList<Consumer<FuzzedDataProvider>> actionList = getActionList();
+
+      while (data.remainingBytes() > 0) {
+        int actionIndex = data.consumeInt(0, actionList.size() - 1);
+        Consumer<FuzzedDataProvider> action = actionList.get(actionIndex);
```
fmeum, line 34: Could this use `pickValue`?

selffuzz/src/test/java/com/code_intelligence/selffuzz/driver/FuzzedDataProviderImplFuzzTest.java
```diff
@@ -0,0 +1,433 @@
...
+
+  void testBoolean(FuzzedDataProvider data) {
+    data.consumeBoolean();
+  }
+
+  void testBooleans(FuzzedDataProvider data) {
+    try {
+      data.consumeBooleans(data.consumeInt());
+    } catch (IllegalArgumentException e) {
+      // catch negative length
```
fmeum, line 95: This can hide actual bugs. Could we instead just return early if `data.consumeInt() < 0`?

The same applies to the other cases of `catch (IllegalArgumentException e)` below.

selffuzz/src/test/java/com/code_intelligence/selffuzz/driver/FuzzedDataProviderImplFuzzTest.java
```diff
@@ -0,0 +1,433 @@
...
+
+    data.pickValue(collection);
+  }
+
+  void testPickValueArray(FuzzedDataProvider data) {
+    int length = data.consumeShort();
+    Integer[] collection;
+    try {
+      collection = new Integer[length];
+    } catch (NegativeArraySizeException e) {
```
fmeum, line 293: Instead of catching, just return early if `length` is negative.

selffuzz/src/test/java/com/code_intelligence/selffuzz/driver/FuzzedDataProviderImplFuzzTest.java
```diff
@@ -0,0 +1,433 @@
...
+        int actionIndex = data.consumeInt(0, actionList.size() - 1);
+        Consumer<FuzzedDataProvider> action = actionList.get(actionIndex);
+        action.accept(data);
+      }
+    }
+  }
+
+  ArrayList<Consumer<FuzzedDataProvider>> getActionList() {
+    ArrayList<Consumer<FuzzedDataProvider>> actionList = new ArrayList<>();
+    actionList.add(this::testBoolean);
```
fmeum, line 42: That's a great list, thanks for the effort!

You can return `Arrays.asList(...)` to make this less verbose, but there is no need to change this now if you don't want to.

Changing the return type to `List<...>` and wrapping the result into `Collections.unmodifiableList(...)` would be welcome though.

selffuzz/src/test/java/com/code_intelligence/selffuzz/driver/FuzzedDataProviderImplFuzzTest.java
```diff
@@ -0,0 +1,433 @@
...
+    }
+  }
+
+  void testRemainingAsString(FuzzedDataProvider data) {
+    data.consumeRemainingAsString();
+  }
+
+  void testAsciiString(FuzzedDataProvider data) {
+    try {
+      data.consumeAsciiString(data.consumeInt());
```
fmeum, line 267: We could verify that all the characters are ASCII here.

src/main/native/com/code_intelligence/jazzer/driver/com_code_intelligence_selffuzz_jazzer_driver_FuzzedDataProviderImpl.h
```diff
@@ -0,0 +1,356 @@
...
+   jazzer_fuzzed_data_provider_impl) and copied here and modified
+
+    Specifically it was taken from
+    `bazel-bin/src/main/java/com/code_intelligence/jazzer/driver/fuzzed_data_provider_impl.hdrs.h/com_code_intelligence_jazzer_driver_FuzzedDataProviderImpl.h`
+    and modified such that all symbol names now include `selffuzz` in order to
+   be usable by our selffuzzing library. Due to the shading of our class files,
+   the selffuzz code will attempt to load symbols with the new classpath which
+   otherwise don't exist.
+
+    Normally the built jazzer jar will have end up a dylib (on Mac) in
```
fmeum, line 27:
```suggestion
    Normally the built jazzer jar will have a dylib (on Mac) in
```

src/main/native/com/code_intelligence/jazzer/driver/com_code_intelligence_selffuzz_jazzer_driver_FuzzedDataProviderImpl.h
```diff
@@ -0,0 +1,356 @@
...
+    `/com/code_intelligence/jazzer/driver/jazzer_fuzzed_data_provider_macos_aarch64/libjazzer_fuzzed_data_provider.dylib`
+
+    FuzzedDataProviderImpl will load that will a call to
```
fmeum, line 30:
```suggestion
    FuzzedDataProviderImpl will load that with a call to
```

src/main/native/com/code_intelligence/jazzer/driver/fuzzed_data_provider.cpp
```diff
@@ -96,9 +98,10 @@ ConsumeIntegralArray(JNIEnv &env, jobject self, jint max_length) {
   const auto *dataPtr =
       reinterpret_cast<const uint8_t *>(env.GetLongField(self, gDataPtrField));
   jint remainingBytes = env.GetIntField(self, gRemainingBytesField);
+  jint requested_bytes = static_cast<jint>(sizeof(T)) * max_length;
 
   jint max_num_bytes =
-      std::min(static_cast<jint>(sizeof(T)) * max_length, remainingBytes);
+      std::max(std::min(requested_bytes, remainingBytes), static_cast<jint>(0));
```
fmeum, line 104: While this fixes a crash, it can result in somewhat weird behavior: If `max_length` grows, `max_num_bytes` may shrink.

Instead, we could perform the multiplication above on `uint64_t` to prevent an overflow and then truncate to `jint` here. This would result in saturation at `remainingBytes` rather than fluctuation. What do you think?

fmeum (approved):

fmeum:
We shouldn't include this in the current PR, but I found the following patch to work locally in cifuzz. I do get coverage feedback from the native methods backing `FuzzedDataProvider` with it. I also tried this out with UBSan, but that fails with an obscure error that is probably a bootstrapping problem. 

It runs with `--asan` and `-fsanitize=address`, but didn't find a planted use-after-free directly. This could be interesting to look into more.

```patch
diff --git a/.github/workflows/fuzzing-featured.yaml b/.github/workflows/fuzzing-featured.yaml
index a5680d5b..6c8c9fb8 100644
--- a/.github/workflows/fuzzing-featured.yaml
+++ b/.github/workflows/fuzzing-featured.yaml
@@ -57,7 +57,10 @@ jobs:
         shell: bash
       - id: build-jazzer
         name: Build Jazzer deps
-        run: cd $CHECKOUT_DIR && bazel build //deploy:jazzer-junit-project //deploy:jazzer-project //deploy:jazzer-api-project //selffuzz:jazzer_selffuzz //selffuzz:jazzer_api_selffuzz
+        run: |
+          cd $CHECKOUT_DIR
+          bazel build //deploy:jazzer-junit-project //deploy:jazzer-project //deploy:jazzer-api-project 
+          bazel build --fsanitize=fuzzer-no-link //selffuzz:jazzer_selffuzz //selffuzz:jazzer_api_selffuzz
 
       - id: build-fuzzers
         name: Build Fuzzers
diff --git a/.github/workflows/fuzzing.yaml b/.github/workflows/fuzzing.yaml
index 13a8ef5c..10306606 100644
--- a/.github/workflows/fuzzing.yaml
+++ b/.github/workflows/fuzzing.yaml
@@ -49,7 +49,10 @@ jobs:
         shell: bash
       - id: build-jazzer
         name: Build Jazzer deps
-        run: cd $CHECKOUT_DIR && bazel build //deploy:jazzer-junit-project //deploy:jazzer-project //deploy:jazzer-api-project //selffuzz:jazzer_selffuzz //selffuzz:jazzer_api_selffuzz
+        run: |
+          cd $CHECKOUT_DIR
+          bazel build //deploy:jazzer-junit-project //deploy:jazzer-project //deploy:jazzer-api-project 
+          bazel build --copt=-fsanitize=fuzzer-no-link //selffuzz:jazzer_selffuzz //selffuzz:jazzer_api_selffuzz
 
       - id: build-fuzzers
         name: Build Fuzzers
diff --git a/selffuzz/cifuzz.yaml b/selffuzz/cifuzz.yaml
index 95f3af76..b4f61d55 100644
--- a/selffuzz/cifuzz.yaml
+++ b/selffuzz/cifuzz.yaml
@@ -38,6 +38,7 @@
 ## See https://llvm.org/docs/LibFuzzer.html#options
 engine-args:
  - --experimental_mutator
+ - --native
 
 ## Maximum time to run fuzz tests. The default is to run indefinitely.
 #timeout: 30m

```

src/main/native/com/code_intelligence/jazzer/driver/fuzzed_data_provider.cpp
br-lewis: I'm not sure that using a `uint64_t` here would help much but I may not know enough about C++ conversion options. I tried `static_cast<int64_t>(<a uint64_t bigger than int64_t::max>)` and it resulted in a negative. Is there a saturating conversion function available?
Like
```c++
    uint64_t bytes = static_cast<uint64_t>(std::numeric_limits<int32_t>::max()) + 12340;
    int32_t bytes_int = static_cast<int32_t>(bytes);
    std::cout << "bytes: " << bytes << std::endl;
    std::cout << "int  : " << bytes_int << std::endl;
```

Otherwise, I would probably keep this `jint` territory and do a check if it's negative.

src/main/native/com/code_intelligence/jazzer/driver/fuzzed_data_provider.cpp
```diff
@@ -96,9 +98,10 @@ ConsumeIntegralArray(JNIEnv &env, jobject self, jint max_length) {
   const auto *dataPtr =
       reinterpret_cast<const uint8_t *>(env.GetLongField(self, gDataPtrField));
   jint remainingBytes = env.GetIntField(self, gRemainingBytesField);
```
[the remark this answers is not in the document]
br-lewis: Do you want all the changes here to be a separate commit or just the ones about fixing the issues I found in it while keeping the ones related to getting it running together with the rest?

src/main/native/com/code_intelligence/jazzer/driver/fuzzed_data_provider.cpp
fmeum: Sorry, my comment wasn't fully accurate: You have to perform the multiplication *and* `std::min` on `uint64_t`, and only then truncate to `jint`.

src/main/native/com/code_intelligence/jazzer/driver/fuzzed_data_provider.cpp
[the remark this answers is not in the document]
fmeum: I would prefer separating based on "fixing bugs" and "getting a fuzz test to run". The former can break users who relied on the buggy behavior, the latter should be invisible to them.

src/main/native/com/code_intelligence/jazzer/driver/fuzzed_data_provider.cpp
br-lewis: Pushed a fixed version. If you think that's good, I'll fix up the commits

src/main/native/com/code_intelligence/jazzer/driver/fuzzed_data_provider.cpp
```diff
@@ -96,9 +99,14 @@ ConsumeIntegralArray(JNIEnv &env, jobject self, jint max_length) {
...
-
+  uint64_t requested_bytes = sizeof(T) * max_length;
+  // cap to max int32 so we don't request negative bytes
+  uint64_t capped_bytes =
+      std::min(requested_bytes,
+               static_cast<uint64_t>(std::numeric_limits<int32_t>::max()));
   jint max_num_bytes =
-      std::min(static_cast<jint>(sizeof(T)) * max_length, remainingBytes);
+      std::min(static_cast<jint>(capped_bytes), remainingBytes);
+
```
fmeum, line 109: It seems easier to just cap to `static_cast<uint32_t>(remaining_bytes)` directly, then cast to `jint`. The separate cap to `std::numeric_limits<int32_t>::max()` should be unnecessary as `remaining_bytes` is always at most that large.

src/main/native/com/code_intelligence/jazzer/driver/fuzzed_data_provider.cpp
fmeum: I left one comment, but if I missed something about that, this version is fine too.

fmeum:
Looks good, thanks! I can't approve again as that results in a 500 from GitHub, just assume I did.

PR #840, opened 2023-09-01
junit: Do not depend on `junit-jupiter-engine`

fmeum:
The user has to add and control this dependency themselves if they want
to use JUnit 5. Even if we added it, Maven resolution mechanics mean
that we do not control the version anyway.

This fixes an issue introduced by 1ca007d04325014d4fa0e48d239745f3ecc8fbcf.

See https://github.com/junit-team/junit5/discussions/3441#discussioncomment-6884855

Also require fixing an issue in rules_jvm_external that results in
neverlink deps not appearing on the javadoc classpath.

### DIFF
--- a/repositories.bzl
+++ b/repositories.bzl
@@ -64,6 +64,9 @@ def jazzer_dependencies(android = False):
             # Fixes an incompatibility with latest Bazel introduced in
             # https://github.com/bazelbuild/bazel/commit/d0e29582a2e788e8acdaf53fe30ab7f7dc592df3
             "//third_party:rules_jvm_external-add-toolchain-to-maven-project-jar.patch",
+            # https://github.com/bazelbuild/rules_jvm_external/pull/952
+            # Fixes javadoc generation when using neverlink dependencies.
+            "//third_party:rules_jvm_external-add-neverlink-deps-to-javadoc-classpath.patch",
         ],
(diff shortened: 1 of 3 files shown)


bertschneider (approved):
Is this considered a breaking change?

fmeum:
> Is this considered a breaking change?

It fixes a regression introduced by https://github.com/CodeIntelligenceTesting/jazzer/commit/1ca007d04325014d4fa0e48d239745f3ecc8fbcf, I added this to the PR description and commit message. Since that commit hasn't been released yet, it isn't breaking (luckily :-)).

bertschneider:
I thought that the dependency was introduced earlier and only updated in the mentioned PR.

fmeum:
> I thought that the dependency was introduced earlier and only updated in the mentioned PR.

I checked and it does exist on `main`, but not on `v0.20.1`. You can run `bazel build //deploy:jazzer-junit-pom && cat bazel-bin/deploy/jazzer-junit-pom.xml` to check.

bertschneider:
Locally on `main` `org.junit.jupiter:junit-jupiter-engine` is present, but `org.junit.platform:junit-platform-engine` is missing in `jazzer-junit-pom.xml`, which I expected to be there. This is probably due to its definition as `runtime_deps` in `fuzz_test`.

Do we need to treat `junit-jupiter-engine` differently to begin with? User versions will be preferred anyway.

PR #843, opened 2023-09-05
mutation: Fix and simplify cross-over insertion logic 

fmeum:


bertschneider (approved):
This looks good, minus the pipeline issues :smile:

If I understand it correctly, the additional `Math.min(list.size(), otherList.size())` caused issues. I wonder why the size of the initial list was considered  for inserts to begin with.

fmeum:
> If I understand it correctly, the additional `Math.min(list.size(), otherList.size())` caused issues. I wonder why the size of the initial list was considered for inserts to begin with.

That plus the logic of the form `int toPos = prng.closedRange(0, list.size() - chunkSize)`, which in the case of a map could actually crash as the second argument can become negative.

I think that we just didn't notice that insertion differs from all the other actions in this aspect.

PR #850, opened 2023-09-11
Simplify Javadoc generation

fmeum:


zgtm (approved):
Works like a charm, as can be seen here: https://github.com/CodeIntelligenceTesting/jazzer-docs/commits/main

Thanks for building this and creating the upstream PR!

PR #852, opened 2023-09-14
Release in CI pipeline

zgtm:
This PR automates the release process by creating
 - Github releases
 - git tags
 - Maven artifacts
 - Docker images
 - documentation
in the CI pipeline.

More Information in first commit message.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,3 +20,82 @@ jobs:
 
       - name: Push docker containers
         run: docker/push_all.sh
+
+  generate_docs:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: checkout
+        uses: actions/checkout@v3
+
+      - name: Build documentation
+        run: |
+          bazel build //deploy:jazzer-docs //deploy:jazzer-api-docs //deploy:jazzer-junit-docs
+          cp $(bazel cquery --output=files //deploy:jazzer-docs) ./jazzer-docs.jar
+          cp $(bazel cquery --output=files //deploy:jazzer-api-docs) ./jazzer-api-docs.jar
+          cp $(bazel cquery --output=files //deploy:jazzer-junit-docs) ./jazzer-junit-docs.jar
+
+      - name: Upload jars
+        uses: actions/upload-artifact@v3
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -20,3 +20,90 @@ jobs:
...
+
+      - name: Build Jazzer API docs
+        run: |
+          bazel build //deploy:jazzer-api-docs
+          cp bazel-bin/deploy/jazzer-api-docs.jar .
+
+      - name: Build Jazzer docs
+        run: |
+          bazel build //deploy:jazzer-docs
+          cp bazel-bin/src/main/java/com/code_intelligence/jazzer/jazzer-docs.jar .
```
fmeum, line 39: Instead of hard-coding the paths, we should use `bazel cquery --output=files`.

We could probably also merge these three separate builds into a single one with the three targets specified.

fmeum (approved):
Great work, this will make our lives much easier!

.github/workflows/release.yml
zgtm: > Instead of hard-coding the paths, we should use `bazel cquery --output=files`.

Yeah, that sounds sensible! Will change!

> We could probably also merge these three separate builds into a single one with the three targets specified.

While I have no strong opinions on this, I've split this up so that different steps would fail depending on which docs would fail to build. OTOH, you can still figure that out, if we unify them into one build step …

.github/workflows/release.yml
```diff
@@ -2,85 +2,108 @@ name: Release
 
 on:
   workflow_dispatch:
```
bertschneider, line 4: I wonder if manual invocation should still be allowed.

bertschneider (approved):

.github/workflows/release.yml
zgtm: I've left it in for now for cases where we might still want to run this manually (eg. if the deployment failed and we want to add a commit for fixing the deployment). But if it turns out we never need it, we can get rid of it later! :)

bertschneider: Good point!

PR #855, opened 2023-09-20
junit: Only create `.cifuzz-corpus` if it is the generated corpus

fmeum:
If users add custom corpus directories, the first of those will be used as the generated corpus instead of the default `.cifuzz-corpus` directory. We now no longer create this directory if it is going to stay empty because it isn't used as the generated corpus directory.


bertschneider (approved):
LGTM

PR #862, opened 2023-10-05
Add FuzzTest dictionary support

br-lewis:
This adds dictionary support to JUnit fuzz tests via 2 annotations: `WithDictionary` and `WithDictionaryFile` that allow dictionaries to be specified either as static arrays of tokens or by referring to a dictionary file. Multiple instances of both annotations are allowed and all values will be merged in the dictionary given to libfuzzer.


src/main/java/com/code_intelligence/jazzer/junit/FuzzerDictionary.java
```diff
@@ -0,0 +1,159 @@
...
+  public static Optional<String> createDictionaryFile(
+      List<WithDictionary> inline, List<WithDictionaryFile> files) throws IOException {
+    int sources = inline.size() + files.size();
+    if (sources == 0) {
+      return Optional.empty();
+    }
+
+    Stream<String> joined = Stream.concat(getInlineTokens(inline), getFileTokens(files));
+
+    File f = File.createTempFile(DICTIONARY_PREFIX, DICTIONARY_SUFFIX);
```
fmeum, line 65: `Files.createTempFile` is preferred as it builds on the more modern `Path` class and also uses more restrictive permissions. `path.toFile().deleteOnExit()` still works.

src/main/java/com/code_intelligence/jazzer/junit/FuzzerDictionary.java
```diff
@@ -0,0 +1,159 @@
...
+    f.deleteOnExit();
+    Log.info(String.format("Creating merged dictionary from %d sources", sources));
+
+    try (OutputStream out = Files.newOutputStream(f.toPath())) {
+      joined.forEach(
+          (token) -> {
```
fmeum, line 71:
```suggestion
          token -> {
```

src/main/java/com/code_intelligence/jazzer/junit/FuzzerDictionary.java
```diff
@@ -0,0 +1,159 @@
...
+      return Optional.empty();
+    }
+
+    Stream<String> joined = Stream.concat(getInlineTokens(inline), getFileTokens(files));
+
+    File f = File.createTempFile(DICTIONARY_PREFIX, DICTIONARY_SUFFIX);
+    f.deleteOnExit();
+    Log.info(String.format("Creating merged dictionary from %d sources", sources));
+
+    try (OutputStream out = Files.newOutputStream(f.toPath())) {
```
fmeum, line 69: This uses unbuffered I/O. `Files.newBufferedWriter(path, StandardCharsets.UTF_8)` gives you buffered IO plus more convenient methods.

src/main/java/com/code_intelligence/jazzer/junit/FuzzerDictionary.java
```diff
@@ -0,0 +1,159 @@
...
+      joined.forEach(
+          (token) -> {
+            try {
+              // the tokens will come in without newlines attached, so we append them here before
+              // writing
+              String line = token.concat("\n");
+              out.write(line.getBytes());
```
fmeum, line 76: Not that it matters too much, but this allocates a new String just to then encode it to bytes. This would be more efficient (using the `BufferedWriter`):
```suggestion
              out.append(line).append('\n');
```

src/main/java/com/code_intelligence/jazzer/junit/FuzzerDictionary.java
```diff
@@ -0,0 +1,159 @@
...
+   * @return stream of all lines from each of the files
+   */
+  private static Stream<String> getFileTokens(List<WithDictionaryFile> files) {
+    return files.stream()
+        .map(WithDictionaryFile::resourcePath)
+        .map(FuzzerDictionary::tokensFromFile)
+        .flatMap(List::stream);
+  }
+
+  private static List<String> tokensFromFile(String path) {
```
fmeum, line 112: This makes it more clear that this method operates on resources:
```suggestion
  private static List<String> tokensFromResource(String absoluteResourcePath) {
```

src/main/java/com/code_intelligence/jazzer/junit/FuzzerDictionary.java
```diff
@@ -0,0 +1,159 @@
...
+    try (InputStream resourceFile = ClassLoaderUtils.class.getResourceAsStream(path)) {
```
fmeum, line 113: AFAIK `ClassLoaderUtils` is JUnit-internal. You can use `FuzzerDictionary.class.getClassLoader().getResourceAsStream()` instead. See https://www.baeldung.com/java-class-vs-classloader-getresource for the difference to `class.getResource()`: We want to accept absolute paths only here, relative paths could result in lookups relative to *our* classes.

src/main/java/com/code_intelligence/jazzer/junit/FuzzerDictionary.java
```diff
@@ -0,0 +1,159 @@
...
+      if (resourceFile == null) {
+        throw new FileNotFoundException(path);
+      }
+      BufferedReader reader = new BufferedReader(new InputStreamReader(resourceFile));
+      // I think returning just reader.lines() results in the file stream being closed before it's
```
fmeum, line 118: You can wrap the `reader` in another try-with-resources statement to close it explicitly. It is indeed pretty common to read files into intermediate in-memory data structures instead of keeping them open through a `return`.

src/main/java/com/code_intelligence/jazzer/junit/FuzzerDictionary.java
```diff
@@ -0,0 +1,159 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.code_intelligence.jazzer.junit;
+
+import com.code_intelligence.jazzer.utils.Log;
+import java.io.*;
```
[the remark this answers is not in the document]
fmeum: A formatter can't do this as determining whether `Foo` requires the import of `java.io.Foo` or `com.example.Foo` requires knowledge of the entire compilation classpath (and even then may be ambiguous). If you enable the `google-java-format` plugin in IntelliJ, it should perform this change automatically when you reorganize imports.

src/main/java/com/code_intelligence/jazzer/junit/FuzzTestExecutor.java
```diff
@@ -363,6 +369,8 @@ public Optional<Throwable> execute(
           });
     }
 
+    System.out.println("STARTING LIBFUZZER");
+    System.out.printf("WITH ARGS %s%n", Arrays.toString(libFuzzerArgs.toArray()));
```
fmeum, line 373: More debug prints.

fmeum (changes requested):

fmeum:
Looks good to me after the debug prints have been removed!

fmeum (approved):

PR #874, opened 2023-10-18
BREAKING: JUnit dictionary improvements

fmeum:


src/main/java/com/code_intelligence/jazzer/junit/DictionaryFile.java
```diff
@@ -23,13 +23,13 @@
...
- * <a href="https://llvm.org/docs/LibFuzzer.html#dictionaries">libfuzzer's dictionary syntax</a>.
+ * Attaches a dictionary file located in resources to a fuzz test. The file should follow <a
+ * href="https://llvm.org/docs/LibFuzzer.html#dictionaries">libfuzzer's dictionary syntax</a>.
  */
 @Target({ElementType.METHOD, ElementType.ANNOTATION_TYPE})
 @Retention(RetentionPolicy.RUNTIME)
 @Repeatable(DictionaryFiles.class)
 public @interface DictionaryFile {
+  /** The resource path of the dictionary file. */
   String resourcePath();
```
br-lewis, line 34: Should this also be changed to `value` so you don't need to include `resourcePath`?

fmeum: I thought about that, but found `@DictionaryFile(resourcePath = 'foo')` to be more descriptive. It also makes it easier for us to introduce `filePath` or a similar element in case we should need to later.

br-lewis (approved):

PR #909, opened 2024-09-12
Adjust license to clarify the usage within OSS-Fuzz

serj:


kyakdan (approved):

evverx:
@oliverchang @jonathanmetzman I think https://google.github.io/clusterfuzzlite/ (where it was advertised that it can be used to test projects that aren't integrated into OSS-Fuzz because they aren't critical enough) should be updated. It would also be great if the OSS-Fuzz documentation was also updated to make the license situation more clear.

README.md
```diff
@@ -96,8 +96,10 @@ See [the README](https://github.com/bazelbuild/rules_fuzzing#java-fuzzing) for i
 [Code Intelligence](https://code-intelligence.com) and Google have teamed up to bring support for Java, Kotlin, and other JVM-based languages to [OSS-Fuzz](https://github.com/google/oss-fuzz), Google's project for large-scale fuzzing of open-source software.
 Read [the OSS-Fuzz guide](https://google.github.io/oss-fuzz/getting-started/new-project-guide/jvm-lang/) to learn how to set up a Java project.
 
-**Note**: projects accepted in OSS-Fuzz can use Jazzer for free and benefit from the OSS-Fuzz infrastructure and 
-licensing for automated analysis and continuous integration.
+**Note**: Open source projects can use Jazzer for free and benefit from the 
+OSS-Fuzz infrastructure, including ClusterFuzzLite and OSS-Fuzz-Gen for
+automated analysis and continuous integration. There is no risk of accidental 
+license violation as long as Jazzer is used for testing open-source code.
```
evverx, line 102: I'm not sure it's accurate. Recently I tried to turn on a coverage service whose TOS allows testing open-source projects for free on their infrastructure via their GitHub integration and that got blocked by GitHub.

Personally I'm not going to download or build those OSS-Fuzz images but I still think the OSS-Fuzz/ClusterFuzzLite documentation/license should clearly mention all those things.

Just to be clear I understand and respect the decision and I'm not trying to convince anyone to relicense it back. It's just that I think the goal should be to minimize the risk by at least clearly stating all those things in all the places.

serj: @evverx I understand your concerns. Together with OSS-Fuzz, we're in the process of clarifying this in the OSS-Fuzz / ClusterFuzzLite docs.

evverx: @serj what I was trying to say is that I don't think this paragraph should be included in README.md. My understanding is that some things depend on who tests open-source projects and where they do it and this license can't control those things. The CodeQL TOS doesn't say anything like that for example (and since it's mostly used on GitHub GitHub can more or less control who turns the CodeQL action on and where it's run). With OSS-Fuzz images that can be downloaded to reproduce bugs locally and ClusterFuzzLite (which is supposed to be used in various places including private GitHub repositories for example) all those lines are kind of blurry.

PR #915, opened 2024-12-17
Add a path traversal sanitizer

tballison:
This adds a FilePathTraversal sanitizer. 

Harnesses set directories under which the application may read and write files via an environment variable.

This does not currently handle symlinks.

### DIFF
--- a/sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/FilePathTraversal.java
+++ b/sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/FilePathTraversal.java
@@ -1,0 +1,349 @@
+/*
+ * Copyright 2025 Code Intelligence GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.code_intelligence.jazzer.sanitizers;
+
+import com.code_intelligence.jazzer.api.FuzzerSecurityIssueCritical;
+import com.code_intelligence.jazzer.api.HookType;
+import com.code_intelligence.jazzer.api.Jazzer;
+import com.code_intelligence.jazzer.api.MethodHook;
+import java.io.File;
+import java.lang.invoke.MethodHandle;
+import java.nio.file.InvalidPathException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * This tests for a file read or write of a specific file path whether relative or absolute.
+ *
+ * <p>This checks only for literal, absolute, normalized paths. It does not process symbolic links.
+ *
+ * <p>The default target is {@link FilePathTraversal#DEFAULT_TARGET_STRING}
+ *
+ * <p>Users may customize a customize the target by setting the full path in the environment
+ * variable {@link FilePathTraversal#FILE_PATH_TARGET_KEY}
+ *
+ * <p>This does not currently check for reading metadata from the target file.
+ */
+public class FilePathTraversal {
+  public static final String FILE_PATH_TARGET_KEY = "jazzer.file_path_traversal_target";
+  public static final String DEFAULT_TARGET_STRING = "../jazzer-traversal";
+
+  private static final Logger LOG = Logger.getLogger(FilePathTraversal.class.getName());
+
+  private static Path RELATIVE_TARGET;
+  private static Path ABSOLUTE_TARGET;
+  private static boolean IS_DISABLED = false;
+  private static boolean IS_SET_UP = false;
+
+  private static void setUp() {
+    String customTarget = System.getProperty(FILE_PATH_TARGET_KEY);
+    if (customTarget != null && !customTarget.isEmpty()) {
+      LOG.log(Level.FINE, "custom target loaded: " + customTarget);
+      setTargets(customTarget);
+    } else {
+      // check that this isn't being run at the root directory
+      Path cwd = Paths.get(".").toAbsolutePath();
+      if (cwd.getParent() == null) {
+        LOG.warning(
+            "Can't run from the root directory with the default target. "
+                + "The FilePathTraversal sanitizer is disabled.");
+        IS_DISABLED = true;
+      }
+      setTargets(DEFAULT_TARGET_STRING);
+    }
+  }
+
+  private static void setTargets(String targetPath) {
+    Path p = Paths.get(targetPath);
+    Path pwd = Paths.get(".");
+    if (p.isAbsolute()) {
+      ABSOLUTE_TARGET = p.toAbsolutePath().normalize();
+      RELATIVE_TARGET = pwd.toAbsolutePath().relativize(ABSOLUTE_TARGET).normalize();
+    } else {
+      ABSOLUTE_TARGET = pwd.resolve(p).toAbsolutePath().normalize();
+      RELATIVE_TARGET = p.normalize();
+    }
+  }
+
+  // intentionally skipping createLink and createSymbolicLink
+  @MethodHook(
+      type = HookType.BEFORE,
+      targetClassName = "java.nio.file.Files",
+      targetMethod = "createDirectory")
+  @MethodHook(
+      type = HookType.BEFORE,
+      targetClassName = "java.nio.file.Files",
+      targetMethod = "createDirectories")
+  @MethodHook(
+      type = HookType.BEFORE,
+      targetClassName = "java.nio.file.Files",
+      targetMethod = "createFile")
+  @MethodHook(
+      type = HookType.BEFORE,
+      targetClassName = "java.nio.file.Files",
+      targetMethod = "createTempDirectory")
+  @MethodHook(
+      type = HookType.BEFORE,
+      targetClassName = "java.nio.file.Files",
+      targetMethod = "createTempFile")
+  @MethodHook(
+      type = HookType.BEFORE,
+      targetClassName = "java.nio.file.Files",
+      targetMethod = "delete")
+  @MethodHook(
+      type = HookType.BEFORE,
+      targetClassName = "java.nio.file.Files",
+      targetMethod = "deleteIfExists")
+  @MethodHook(
+      type = HookType.BEFORE,
+      targetClassName = "java.nio.file.Files",
+      targetMethod = "lines")
+  @MethodHook(
+      type = HookType.BEFORE,
+      targetClassName = "java.nio.file.Files",
+      targetMethod = "newByteChannel")
+  @MethodHook(
+      type = HookType.BEFORE,
+      targetClassName = "java.nio.file.Files",
+      targetMethod = "newBufferedReader")
+  @MethodHook(
+      type = HookType.BEFORE,
+      targetClassName = "java.nio.file.Files",
+      targetMethod = "newBufferedWriter")
+  @MethodHook(
+      type = HookType.BEFORE,
+      targetClassName = "java.nio.file.Files",
+      targetMethod = "readString")
+  @MethodHook(
+      type = HookType.BEFORE,
+      targetClassName = "java.nio.file.Files",
+      targetMethod = "newBufferedReader")
+  @MethodHook(
+      type = HookType.BEFORE,
+      targetClassName = "java.nio.file.Files",
+      targetMethod = "readAllBytes")
+  @MethodHook(
+      type = HookType.BEFORE,
+      targetClassName = "java.nio.file.Files",
+      targetMethod = "readAllLines")
+  @MethodHook(
+      type = HookType.BEFORE,
+      targetClassName = "java.nio.file.Files",
+      targetMethod = "readSymbolicLink")
+  @MethodHook(
+      type = HookType.BEFORE,
+      targetClassName = "java.nio.file.Files",
+      targetMethod = "write")
+  @MethodHook(
+      type = HookType.BEFORE,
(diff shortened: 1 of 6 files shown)


oetr:
Awesome work, thanks!

Initially I liked the idea of an allow list for directories. But on a second thought, if the fuzzer is able to access a file that's a few directories up from the current directory, then most likely path traversal is possible. This saves us the complexity of matching the files on the allow list, and the user of providing those, and not forgetting to enable this sanitizer.

The second aspect that is missing in this PR is to guide the fuzzer towards path traversal, which is difficult to achieve with an allow list. 

Since you have not opened a branch in this repo, I have added my suggestions in the comments, instead of pushing my changes directly, so apologies for that!

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/FilePathTraversal.java
```diff
@@ -0,0 +1,367 @@
...
+            if (candidateNormalized.startsWith(allowedNormalized) &&
+                ! candidateNormalized.equals(allowedNormalized)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+}
+
```
oetr, line 367: - replaced allow list by the target ../../jazzer-traversal
- moved type handling into one place

```suggestion
package com.code_intelligence.jazzer.sanitizers;

import com.code_intelligence.jazzer.api.FuzzerSecurityIssueCritical;
import com.code_intelligence.jazzer.api.HookType;
import com.code_intelligence.jazzer.api.Jazzer;
import com.code_intelligence.jazzer.api.MethodHook;
import java.io.File;
import java.lang.invoke.MethodHandle;
import java.nio.file.Path;
import java.nio.file.Paths;

/**
 * This tests for a file read or write of a specific file name.
 *
 * <p>This checks only for literal, absolute, normalized paths. It does not process symbolic links.
 */
public class FilePathTraversal {
  public static final String TARGET =
      Paths.get("..", "..", "jazzer-traversal").normalize().toString();

  // intentionally skipping createLink and createSymbolicLink
  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.nio.file.Files",
      targetMethod = "createDirectory")
  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.nio.file.Files",
      targetMethod = "createDirectories")
  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.nio.file.Files",
      targetMethod = "createFile")
  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.nio.file.Files",
      targetMethod = "createTempDirectory")
  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.nio.file.Files",
      targetMethod = "createTempFile")
  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.nio.file.Files",
      targetMethod = "delete")
  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.nio.file.Files",
      targetMethod = "deleteIfExists")
  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.nio.file.Files",
      targetMethod = "lines")
  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.nio.file.Files",
      targetMethod = "newByteChannel")
  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.nio.file.Files",
      targetMethod = "newBufferedReader")
  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.nio.file.Files",
      targetMethod = "newBufferedWriter")
  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.nio.file.Files",
      targetMethod = "readString")
  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.nio.file.Files",
      targetMethod = "newBufferedReader")
  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.nio.file.Files",
      targetMethod = "readAllBytes")
  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.nio.file.Files",
      targetMethod = "readAllLines")
  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.nio.file.Files",
      targetMethod = "readSymbolicLink")
  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.nio.file.Files",
      targetMethod = "write")
  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.nio.file.Files",
      targetMethod = "writeString")
  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.nio.file.Files",
      targetMethod = "newInputStream")
  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.nio.file.Files",
      targetMethod = "newOutputStream")
  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.nio.file.probeContentType",
      targetMethod = "open")
  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.nio.channels.FileChannel",
      targetMethod = "open")
  public static void pathFirstArgHook(
      MethodHandle method, Object thisObject, Object[] arguments, int hookId) {
    if (arguments.length > 0) {
      checkPath(arguments[0], hookId);
    }
  }

  /**
   * Checks to confirm that a path that is read from or written to is in an allowed directory.
   *
   * @param method
   * @param thisObject
   * @param arguments
   * @param hookId
   */
  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.nio.file.Files",
      targetMethod = "copy")
  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.nio.file.Files",
      targetMethod = "mismatch")
  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.nio.file.Files",
      targetMethod = "move")
  public static void copyMismatchMvHook(
      MethodHandle method, Object thisObject, Object[] arguments, int hookId) {
    if (arguments.length > 1) {
      checkPath(arguments[0], hookId); // from
      checkPath(arguments[1], hookId); // to
    }
  }

  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.io.FileReader",
      targetMethod = "<init>")
  public static void fileReaderHook(
      MethodHandle method, Object thisObject, Object[] arguments, int hookId) {
    if (arguments.length > 0) {
      checkPath(arguments[0], hookId);
    }
  }

  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.io.FileWriter",
      targetMethod = "<init>")
  public static void fileWriterHook(
      MethodHandle method, Object thisObject, Object[] arguments, int hookId) {
    if (arguments.length > 0) {
      checkPath(arguments[0], hookId);
    }
  }

  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.io.FileInputStream",
      targetMethod = "<init>")
  public static void fileInputStreamHook(
      MethodHandle method, Object thisObject, Object[] arguments, int hookId) {
    if (arguments.length > 0) {
      checkPath(arguments[0], hookId);
    }
  }

  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.io.FileOutputStream",
      targetMethod = "<init>")
  public static void processFileOutputStartHook(
      MethodHandle method, Object thisObject, Object[] arguments, int hookId) {
    if (arguments.length > 0) {
      checkPath(arguments[0], hookId);
    }
  }

  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.util.Scanner",
      targetMethod = "<init>")
  public static void scannerHook(
      MethodHandle method, Object thisObject, Object[] arguments, int hookId) {
    if (arguments.length > 0) {
      checkPath(arguments[0], hookId);
    }
  }

  @MethodHook(
      type = HookType.BEFORE,
      targetClassName = "java.io.FileOutputStream",
      targetMethod = "<init>")
  public static void fileOutputStreamHook(
      MethodHandle method, Object thisObject, Object[] arguments, int hookId) {
    if (arguments.length > 0) {
      checkPath(arguments[0], hookId);
    }
  }

  private static void checkPath(Object f, int hookId) {
    String path;
    if (f instanceof Path) {
      path = ((Path) f).toString();
    } else if (f instanceof File) {
      path = ((File) f).getPath();
    } else if (f instanceof String) {
      path = (String) f;
    } else {
      return;
    }

    if (path.equals(TARGET)) {
      Jazzer.reportFindingFromHook(new FuzzerSecurityIssueCritical("File path traversal: " + path));
    } else {
      Jazzer.guideTowardsContainment(path, TARGET, hookId);
    }
  }
}


```

sanitizers/src/test/java/com/example/BUILD.bazel
```diff
@@ -58,6 +58,19 @@ java_fuzz_target_test(
...
+        "FilePathTraversal.java",
+    ],
+    allowed_findings = [
+        "com.code_intelligence.jazzer.api.FuzzerSecurityIssueCritical",
+    ],
+    target_class = "com.example.FilePathTraversal",
+    #not clear why reproducer doesn't work TODO -- fix this
+    verify_crash_reproducer = False,
+)
+
```
oetr, line 73: - added deps for using Jazzer annotations in the example

```suggestion
java_fuzz_target_test(
    name = "FilePathTraversal",
    srcs = [
        "FilePathTraversal.java",
    ],
    allowed_findings = [
        "com.code_intelligence.jazzer.api.FuzzerSecurityIssueCritical",
    ],
    target_class = "com.example.FilePathTraversal",
    #not clear why reproducer doesn't work TODO -- fix this
    verify_crash_reproducer = False,
    deps = [
        "//src/main/java/com/code_intelligence/jazzer/mutation/annotation",
    ],
)
)

```

sanitizers/src/test/java/com/example/FilePathTraversal.java
```diff
@@ -0,0 +1,52 @@
...
+            try (BufferedReader r = Files.newBufferedReader(p, StandardCharsets.UTF_8)) {
+                r.read();
+            } catch (IOException ignored) {
+                //swallow
+            }
+        } catch (InvalidPathException ignored) {
+            //swallow
+        }
+    }
+}
```
oetr, line 52: - replaced `FuzzedDataProvider` by the new mutator framework, also showing off some of the Jazzer annotations!
- removed the now unused allow list and environmental varibles to turn on the file path traversal sanitizer that is now always enabled by default
```suggestion
package com.example;

import com.code_intelligence.jazzer.mutation.annotation.DoubleInRange;
import com.code_intelligence.jazzer.mutation.annotation.NotNull;
import com.code_intelligence.jazzer.mutation.annotation.WithUtf8Length;
import java.io.BufferedReader;
import java.io.IOException;
import java.nio.charset.StandardCharsets;
import java.nio.file.Files;
import java.nio.file.InvalidPathException;
import java.nio.file.Path;
import java.nio.file.Paths;

public class FilePathTraversal {

  public static void fuzzerTestOneInput(
      @WithUtf8Length(max = 100) @NotNull String pathFromFuzzer,
      @NotNull @DoubleInRange(min = 0.0, max = 1.0) Double fixedPathProbability) {
    // Slow down the fuzzer a bit, otherwise it finds file path traversal way too quickly!
    String path = fixedPathProbability < 0.95 ? "/a/b/c/fixed-path" : pathFromFuzzer;
    try {
      Path p = Paths.get(path);
      try (BufferedReader r = Files.newBufferedReader(p, StandardCharsets.UTF_8)) {
        r.read();
      } catch (IOException ignored) {
        // swallow
      }
    } catch (InvalidPathException ignored) {
      // swallow
    }
  }
}
```

oetr (commented):
Unfortunately, I ran the format script `format.sh` before committing my changes locally, so now every line is different!

oetr:
Getting the paths right is always more difficult than it looks :)
@tballison Thanks for these great points--now we can make this sanitizer even better!

- I suggest, we change the default target to `../jazzer-traversal`. If someone is fuzzing from the root we might want to disable this sanitizer, unless the user provides a valid target.
- Then we derive two paths from the default/user-provided target once during initialization as follows:
  - if the target is relative:
    - we normalize the relative path and save it as `relativeTarget`;
    - and convert it into an absolute path starting from CWD, call it `absoluteTarget`
  - if the target is absolute:
    - we clean it and try to derive a relative path from CWD to it (it is not always possible to derive such path, in which case this sanitizer should only guite and compare with the absolute target) and save it as `relativeTarget`;
    - clean it and call it `absoluteTarget`
- During fuzzing, we dispatch the comparison based on whether the query path is absolute or relative. And here we should also clean/normalize the query path
- Guiding is trickier, but I have an idea
  - if the query path we get is relative, we guide it towards `relativeTarget`
  - otherwise guide towards `absoluteTarget`

This should work from any location. This will error out if the location cannot be resolved relative to CWD, but then it means that the target is user-set, so the user's responsible for fixing it.

WDYT?

tballison:
K. I think I implemented the above. Let me know what you think.

oetr:
@tballison Awesome job, thank you!

From my point of view, following points should be addressed before we can release this:
- the sanitizer should use the `BugDetectors` api (https://github.com/CodeIntelligenceTesting/jazzer/blob/main/src/main/java/com/code_intelligence/jazzer/api/BugDetectors.java) to set the target instead of environmental variables. See [here](https://github.com/CodeIntelligenceTesting/jazzer/blob/7f9015b9ce16f23669d3aca4c136cbe5452d5b56/sanitizers/src/test/java/com/example/SsrfSocketConnectToHost.java#L19) for example usage
- add several tests to check if the bug detector works with JUnit5
- rework target-setting so that it can be used (set and re-set dynamically) in `@BeforeEach` blocks in JUnit
- running from root directory should be ok if the target is absolute; or if relative and it doesn't get truncated
- refactor, such that fuzzer guiding happens in one place
- replace all-caps by camel case

If you like, we can merge this PR "as-is" (after squash+rebase), and I can take over and do the rest. You will certainly get the credit for this bug detector in the release notes :rocket: 
That would be my preferred option, but If you like to be bossed around, we can spend a few more iteration cycles!
Let me know what you think.

tballison:
@oetr go forth! Let me not hinder you any further! :rofl:

oetr:
@tballison great, thanks! Can you rebase on main and squash all commits into one?

tballison:
I _think_ that did it. :rofl: I'm running tests locally now :crossed_fingers:

tballison:
Local tests pass. :rocket:

oetr (approved):
LGTM, thanks! As discussed, this sanitizer will be updated in a followup PR.

Issue #919, opened 2024-12-21
[0.23] jazzer junit switched to a runtime api dependency


ben-manes:
I am a little confused as to why this would be intentional or if I am misusing your library. The [0.22.1 pom](https://search.maven.org/artifact/com.code-intelligence/jazzer-junit/0.22.1/jar) has `jazzer-api` as a compile time dependency, whereas the [0.23 pom](https://search.maven.org/artifact/com.code-intelligence/jazzer-junit/0.23.0/jar) moves all of its dependencies as runtime. Therefore, when upgrading I receive the following error.

```console
> Task :caffeine:compileTestJava
/Users/ben/projects/caffeine/caffeine/src/test/java/com/github/benmanes/caffeine/fuzz/CaffeineSpecFuzzer.java:18: error: package com.code_intelligence.jazzer.api does not exist
import com.code_intelligence.jazzer.api.FuzzedDataProvider;
                                       ^
/Users/ben/projects/caffeine/caffeine/src/test/java/com/github/benmanes/caffeine/fuzz/CaffeineSpecFuzzer.java:31: error: cannot find symbol
  public void parse(FuzzedDataProvider data) {
                    ^
  symbol:   class FuzzedDataProvider
  location: class CaffeineSpecFuzzer
```

A runtime dependency typically means that users should not use those classes themselves as an implementation detail. For example using the JDBC api with a runtime Postgres driver, or SLF4J with a runtime Logback provider. If an API should be used by consumers then that would be a compile time dependency. This does not seem to be the case unless my tests are an incorrect usage,

```java
package com.github.benmanes.caffeine.fuzz;

import com.code_intelligence.jazzer.api.FuzzedDataProvider;
import com.code_intelligence.jazzer.junit.FuzzTest;
import com.github.benmanes.caffeine.cache.CaffeineSpec;

/**
 * @author ben.manes@gmail.com (Ben Manes)
 */
public final class CaffeineSpecFuzzer {

  // These tests require the environment variable JAZZER_FUZZ=1 to try new input arguments

  @FuzzTest(maxDuration = "5m")
  @SuppressWarnings("CheckReturnValue")
  public void parse(FuzzedDataProvider data) {
    try {
      CaffeineSpec.parse(data.consumeRemainingAsString());
    } catch (IllegalArgumentException expected) { /* ignored */ }
  }
}
```

[...]

PR #921, opened 2025-01-14
Fix Maven releases

fmeum:
See individual commit messages. This still requires https://github.com/bazel-contrib/rules_jvm_external/pull/1308, which can be brought in via an `archive_override` after it has been merged or as a patch before that.

Work towards #919


HenrichN:
@fmeum  Thanks a lot!

oetr (approved):
Thank you!

Then we can wait with the next release until https://github.com/bazel-contrib/rules_jvm_external/pull/1308 is released.

PR #924, opened 2025-01-17
Update rules_jvm_external to fix POM

fmeum:
Requires updating a few deps, which in turn adds more checks to ktlint. Most of them can be fixed automatically (see commits 2 and 3, reviewing with "Hide whitespace" makes this more manageable), very few need manual suppression (commit 4).

Work towards #919


oetr (approved):
LGTM, thanks!

PR #932, opened 2025-07-03
Add `Unsafe` array access sanitizer

Marcono1234:
Adds a sanitizer which looks for invalid array access performed by `sun.misc.Unsafe`. Unlike native memory access performed by `Unsafe`, array access can be sanitized in a stateless way just based on the arguments passed to the `Unsafe` method. And multiple Java libraries have used `Unsafe` for arrays in the past to improve performance, see related security advisories GHSA-8wh2-6qhj-h7j9 and GHSA-973x-65j7-xcf4.

Note that fuzzing likely cannot find all such invalid accesses because some of it occurs for numeric overflow respectively multiple MB of processed data, which the fuzzer might be unable to generate.

See related #891

I used #915 as reference for how to implement a sanitizer. However, I am not very familiar with Bazel and this project setup here, and also have / had issues with building it locally on Windows. Therefore I have marked this PR as Draft for now.

Any feedback is appreciated!

### DIFF
--- a/sanitizers/src/test/java/com/example/OsCommandInjectionProcessBuilder.java
+++ b/sanitizers/src/test/java/com/example/OsCommandInjectionProcessBuilder.java
@@ -16,12 +16,11 @@
 
 package com.example;
 
-import com.code_intelligence.jazzer.api.FuzzedDataProvider;
+import com.code_intelligence.jazzer.mutation.annotation.Ascii;
 import java.util.concurrent.TimeUnit;
 
 public class OsCommandInjectionProcessBuilder {
-  public static void fuzzerTestOneInput(FuzzedDataProvider data) {
-    String input = data.consumeRemainingAsAsciiString();
+  public static void fuzzerTestOneInput(@Ascii String input) {
     try {
       ProcessBuilder processBuilder = new ProcessBuilder(input);
       processBuilder.environment().clear();


sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/BUILD.bazel
```diff
@@ -48,6 +48,15 @@ java_library(
     ],
 )
 
+java_library(
+    name = "unsafe_array_out_of_bounds",
```
Marcono1234: The implementation currently actually detects a bit more than just "out-of-bounds access" (e.g. it also detects unaligned object array access, and reading / writing bytes from an object array and the other way around).

So maybe a name like "unsafe_array_invalid_access" or similar would be better?

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/UnsafeArrayOutOfBounds.java
```diff
@@ -0,0 +1,457 @@
...
+    if (componentType == null) {
+      // Not an array
+      return;
+    }
+
+    if (!componentType.isPrimitive()) {
+      // Reading or writing bytes to an array of references; might be possible but seems
+      // rather unreliable and might mess with the garbage collector?
+      report("Reading or writing bytes from a " + objClass.getTypeName());
+    }
```
Marcono1234: Not sure if this is too strict. On the other hand there might be no guarantees on how large object references are so any access where object references are treated as bytes seems error-prone.

sanitizers/src/test/java/com/example/BUILD.bazel
```diff
@@ -218,6 +218,31 @@ java_fuzz_target_test(
...
+java_fuzz_target_test(
+    name = "UnsafeArrayOutOfBounds",
+    srcs = [
+        "UnsafeArrayOutOfBounds.java",
+    ],
+    allowed_findings = [
+        "com.code_intelligence.jazzer.api.FuzzerSecurityIssueCritical",
+    ],
+    target_class = "com.example.UnsafeArrayOutOfBounds",
+    verify_crash_reproducer = False,
```
Marcono1234, line 239: During my testing when reproducer verification was enabled no sanitizer exception was thrown. Are sanitizers generally not enabled when running in reproducer / regression (?) mode?

That might be rather problematic; in the case of `Unsafe` that might crash the JVM, and I guess for the other sanitizers (such as the path traversal one) it could also have undesired consequences (e.g. if due to path traversal files are placed at random paths on the machine).

sanitizers/src/test/java/com/example/BUILD.bazel
```diff
@@ -218,6 +218,31 @@ java_fuzz_target_test(
...
+    srcs = [
+        "UnsafeArrayOutOfBoundsValid.java",
+    ],
+    allowed_findings = [],
+    fuzzer_args = [
+        # Test does not depend on fuzzing input; just run it once
+        "-runs=1",
+    ],
+    target_class = "com.example.UnsafeArrayOutOfBoundsValid",
+)
```
Marcono1234: Not sure if this is a good approach to ensure that the `Unsafe` sanitizer does not report errors for valid access.

sanitizers/src/test/java/com/example/UnsafeArrayOutOfBounds.java
```diff
@@ -0,0 +1,552 @@
...
+      UNSAFE.getObject(new String[2], Unsafe.ARRAY_OBJECT_BASE_OFFSET + 1);
+    }
+  }
+
+  public static void fuzzerTestOneInput(FuzzedDataProvider data) throws Exception {
+    Method[] testMethods = TestMethods.class.getDeclaredMethods();
+    // Since all of these methods are expected to cause a sanitizer exception, pick a random one and
+    // run it
+    // TODO: Is this a proper way to implement this?
+    Method testMethod = testMethods[data.consumeInt(0, testMethods.length - 1)];
```
Marcono1234: Not sure if this is a good / proper way to ensure that all of the methods cause a sanitizer error.

Once a fuzzing test run finds the first expected failing input (in this case here basically _any input_), does it keep reusing that? In that case it would never run any of the other test methods here.

Is there a better way to solve this, other than adding dozens of separate classes each with their own `fuzzerTestOneInput`?

Marcono1234:
Thanks! That only works for Linux though because `deploy/deploy_local.sh` is a Bash script, right? Unfortunately I am on Windows. And even if I try to run the commands there individually, it then fails at a later point, apparently because the `jazzer.publish-publisher` script it generates is again a Bash script.

I will see if I can maybe get it working with WSL 2 or a Docker container. But if that then only produces the Linux and not the Windows native libraries for Jazzer (not completely sure how the native build for Jazzer works) it will also be a bit cumbersome to then use these artifacts.


(Side note: I had to update locally some of the `bazel_dep` entries in `MODULE.bazel` especially `rules_kotlin` to be able to build the project on Windows 11. But I don't think this is related to the `deploy_local` issues I am facing now.)

fmeum:
If you follow the steps at https://bazel.build/install/windows#install-compilers to install MSYS2 and Visual Studio, `sh_binary`s should run just fine on Windows (via `bazel run`) and the native libraries are built for Windows. You can build your own release jars that way.

The publishing script for `java_export` is unfortunately hardcoded to a Bash script without a launcher even on Windows, so it won't work. I will see whether I can fix that in rules_jvm_external.

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/UnsafeArrayOutOfBounds.java
```diff
@@ -0,0 +1,554 @@
...
+      targetMethodDescriptor = "(Ljava/lang/Object;JLjava/lang/Object;JJ)V")
+  public static void copyMemoryHook(
+      MethodHandle method, Object thisObject, Object[] arguments, int hookId) {
+    long size = (long) arguments[4];
+    checkAccess(arguments[0], (long) arguments[1], size);
+    checkAccess(arguments[2], (long) arguments[3], size);
+  }
+
+  private static void report(String message) {
+    Jazzer.reportFindingFromHook(new FuzzerSecurityIssueCritical(message));
```
Marcono1234: Might be nice to report also for which `Unsafe` method this occurred. Since the hooks are run before the actual `Unsafe` method, it does not appear in the stack trace. For code like `unsafe.putLong(dest, destOffset, unsafe.getLong(src, srcOffset))` that makes it a bit more difficult to tell which of the `Unsafe` calls caused the exception, and you cannot debug inside the `Unsafe` call either since sanitization already happens before.

Any suggestions for how reporting could be improved (in case you consider this an issue) are welcome.

Marcono1234:
Thanks for the hints!

> The publishing script for `java_export` is unfortunately hardcoded to a Bash script without a launcher even on Windows, so it won't work. I will see whether I can fix that in rules_jvm_external.

I worked around it now by building within a Docker container, but maybe other users would benefit from that as well, so thanks for your work on this.

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/UnsafeArrayOutOfBounds.java
```diff
@@ -0,0 +1,554 @@
...
+    } else if (c == byte.class) {
+      bytesCount = 1;
+    } else if (c == char.class || c == short.class) {
+      bytesCount = 2;
+    } else if (c == int.class || c == float.class) {
+      bytesCount = 4;
+    } else if (c == long.class || c == double.class) {
+      bytesCount = 8;
+    } else {
+      throw new AssertionError("Unexpected type: " + c);
```
oetr: Sanitizers are only allowed to throw/report findings.
If the program reaches here, the hook should simply return.

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/BUILD.bazel
oetr: Maybe just `"unsafe_sanitizer"` aka `UnsafeSanitizer.java` (since Unsafe is already taken!) and keep it ready for future additions?

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/UnsafeArrayOutOfBounds.java
```diff
@@ -0,0 +1,554 @@
...
+   ====================================================
+  */
+
+  private static int getBytesCount(Class<?> c) {
+    int bytesCount;
+    if (c == boolean.class) {
+      // Assumes that `Unsafe.ARRAY_BOOLEAN_INDEX_SCALE > 0`
+      bytesCount = 1;
+    } else if (c == byte.class) {
+      bytesCount = 1;
```
oetr: Is there a reason to hard-code these? Why not use `Unsafe.ARRAY_BOOLEAN_INDEX_SCALE`, etc.?

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/UnsafeArrayOutOfBounds.java
oetr: Perhaps we could extend our hooks API and forward the original method name to the hook function.
The only workaround that I can think of is to hook each method individually and report the correct method name there :sweat_smile: 
But I don't consider this an issue.

sanitizers/src/test/java/com/example/BUILD.bazel
oetr: > Are sanitizers generally not enabled when running in reproducer / regression (?) mode?

They are, but the bazel rule `java_fuzz_target_test` in the Jazzer repo cannot handle `FuzzerSecurityIssueCritical` findings in regression mode (we correctly get exit code 77, but no finding information). Testing with maven/gradle + JUnit, everything works as expected (each `@FuzzTest` throws findings in regression mode).

sanitizers/src/test/java/com/example/UnsafeArrayOutOfBounds.java
oetr: We cannot leave it up to the fuzzer to randomly find some findings. We want to ensure that all those methods generate a finding. You can do something like this instead:
```starlark
[java_fuzz_target_test(
    name = "UnsafeArrayOutOfBounds_" + method,
    srcs = [
        "UnsafeArrayOutOfBounds.java",
    ],
    fuzzer_args = [
        "-print_final_stats=1",
        "-runs=0",
    ],
    env = {
        "JAZZER_FUZZ": "1",
    },
    allowed_findings = [
        "com.code_intelligence.jazzer.api.FuzzerSecurityIssueCritical",
    ],
    expect_number_of_findings=1,
    target_class = "com.example.UnsafeArrayOutOfBounds",
    target_method = method,
    runtime_deps = [
        "@maven//:org_junit_jupiter_junit_jupiter_engine",
    ],
    deps = [
        "//deploy:jazzer-junit",
    ],
) for method in [
     "compareAndSwapLong_end",
     "compareAndSwapLong",
...
 ]]
```

With the test class looking like this:
```java
public class UnsafeArrayOutOfBounds {
  private static final Unsafe UNSAFE;

  static {
    try {
      Field f = Unsafe.class.getDeclaredField("theUnsafe");
      f.setAccessible(true);
      UNSAFE = (Unsafe) f.get(null);
    } catch (ReflectiveOperationException e) {
      throw new RuntimeException(e);
    }
  }
  @FuzzTest
  public void compareAndSwapLong(Boolean ignored) {
    UNSAFE.compareAndSwapLong(new long[5], Unsafe.ARRAY_LONG_BASE_OFFSET - 1, 0, 1);
  }

  @FuzzTest
  public void compareAndSwapLong_end(Boolean ignored) {
    UNSAFE.compareAndSwapLong(new long[5], Unsafe.ARRAY_LONG_BASE_OFFSET + 5L * Unsafe.ARRAY_LONG_INDEX_SCALE, 0, 1);
  }
...
}

```

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/UnsafeArrayOutOfBounds.java
```diff
@@ -0,0 +1,554 @@
...
+  /**
+   * Checks {@link Unsafe} memory access where an object reference is accessed.
+   *
+   * @param obj the base object for memory access; e.g. for {@link Unsafe#getObject(Object, long)}
+   *     it is the argument at index 0
+   * @param offset the offset for the memory access; e.g. for {@link Unsafe#getObject(Object, long)}
+   *     it is the argument at index 1
+   * @see #checkAccess(Object, long, long)
+   */
+  private static void checkObjectSizedAccess(Object obj, long offset) {
```
oetr: This method looks almost exactly like `private static void checkAccess(Object obj, long offset, long accessSize) {`.
Would it make sense to reduce duplication a bit?

sanitizers/src/test/java/com/example/BUILD.bazel
Marcono1234: Thanks for the clarification! So keeping `verify_crash_reproducer = False` here is fine, right?

A bit related: Maybe the comments "not clear why reproducer doesn't work TODO -- fix this" for the `FilePathTraversal` tests was then also a misunderstanding of how this works?
(though maybe those comments should be removed in a separate PR and not as part of this one)

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/UnsafeArrayOutOfBounds.java
Marcono1234: Do you mean the hook should return and do not perform any sanitization?

Though if this `AssertionError` occurs here it would indicate a bug in this sanitizer implementation. And it is unlikely that it will occur given that the method hooks explicitly define the names of the `Unsafe` methods to intercept; this error would only occur if new `sun.misc.Unsafe` method overloads are added which break the assumptions of this sanitizer implementation.

So is it fine to keep the `AssertionError` here?

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/UnsafeArrayOutOfBounds.java
Marcono1234: I guess technically they should be identical[^1], though this method is more about "how large is an `int`, ...".

So would you prefer usage of the `..._INDEX_SCALE` values here then?

[^1]: If for some reason the `..._INDEX_SCALE` was larger than the corresponding size of the primitive (meaning there are gaps between array elements), then probably lots of `Unsafe` usage would be broken anyway. So I doubt that this is the case for any JDK implementation, and it might not be permitted by the `Unsafe` documentation in the first place.

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/UnsafeArrayOutOfBounds.java
Marcono1234: Since there are some differences between them I wasn't sure if it makes sense to combine them.
I hope the way I combined them now is ok.

sanitizers/src/test/java/com/example/UnsafeArrayOutOfBounds.java
Marcono1234: Thanks for the tip! I have adjusted it now. The method names in this list are mostly lexicographically ordered, except for a few cases where I put the corresponding `_end` method name directly below the other method name.

Marcono1234:
Side note: The current test failure for the GitHub workflow seems to be unrelated to my changes:
```
 [5,521 / 5,709] 123 / 336 tests; Testing //sanitizers/src/test/java/com/example:OsCommandInjectionProcessBuilder; 300s local ... (4 actions running)

TIMEOUT: //sanitizers/src/test/java/com/example:OsCommandInjectionProcessBuilder (Summary)
      C:/users/runneradmin/_bazel_runneradmin/2gbstkcc/execroot/_main/bazel-out/x64_windows-opt/testlogs/sanitizers/src/test/java/com/example/OsCommandInjectionProcessBuilder/test.log
```

sanitizers/src/test/java/com/example/BUILD.bazel
oetr: Yes, setting `verify_crash_reproducer=True` currently has a bug, where the crash file directory differs from the one bazel actually uses, which results in file not found error. Leaving it to false is fine. In the current setup it is checked that the exception is thrown for every test case.

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/UnsafeArrayOutOfBounds.java
oetr: LGTM, thanks!

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/UnsafeArrayOutOfBounds.java
oetr: Alright, let's leave it the way it is.

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/UnsafeArrayOutOfBounds.java
oetr: I initially disliked the idea of breaking countless OSS-Fuzz projects because we forgot a case. 
But after giving it some thought, I now think that having assertion here has a higher chance of any bugs being discovered and reported, resulting in a better sanitizer in the end. So let's keep it and see what it breaks :smiley:

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/UnsafeArrayOutOfBounds.java
oetr: In case users start complaining, we can add a customization layer for cases where primitive- and object arrays are transformed into each other.

oetr (approved):
This looks good to me, thanks! Only the timeout test needs to be fixed like shown in 45405d2e476ebc35be4923d421ab968af016eedc

I don't know why some comments appear twice in different places :see_no_evil:

Marcono1234:
Thanks! I have included 45405d2e476ebc35be4923d421ab968af016eedc now. Should I also squash the commits or will you do that on merge?

Also, maybe the `OsCommandInjectionProcessBuilder` test is in general a bit risky if it invokes arbitrary OS commands. Or are there any security / safety measures in place which prevent the actual execution of the command? (I might have overlooked them)

> I don't know why some comments appear twice in different places 🙈

Do you mean the review comments here on the GitHub UI? I think that happens when you submit review comments on an existing comment thread. Then GitHub shows the comments on that thread but additionally (without any context...) for the review event. Hopefully GitHub improves this some day.

oetr:
> Should I also squash the commits or will you do that on merge?

I am planning to rebase and merge. Can you group meaningful commits together?

> Also, maybe the `OsCommandInjectionProcessBuilder` test is in general a bit risky if it invokes arbitrary OS commands. Or are there any security / safety measures in place which prevent the actual execution of the command? (I might have overlooked them)

It is totally unsafe! I usually run only specific tests locally, and the rest on Github.
An idea to make it safer would be to add custom replace hook that does noops.

Marcono1234:
> Can you group meaningful commits together?

Is it ok if I just squash all my commits into a single one and keep your commit (with slightly different message)? I don't think for my commits it is worth keeping any of the intermediate state as separate commit.

oetr:
> Is it ok if I just squash all my commits into a single one and keep your commit (with slightly different message)? I don't think for my commits it is worth keeping any of the intermediate state as separate commit.

Yes, thank you!

PR #944, opened 2025-08-11
fix: clear last finding before starting a fuzzing run

oetr:


florianGla (approved):

src/test/java/com/code_intelligence/jazzer/junit/FuzzingAfterRegressionCrashTest.java
```diff
@@ -0,0 +1,167 @@
...
+            event(
+                type(FINISHED),
+                test(uniqueIdSubstrings(ENGINE, CLAZZ, CRASH_FUZZ.getDescriptorId(), INVOCATION)),
+                displayName("Fuzzing..."),
+                finishedWithFailure(
+                    instanceOf(FuzzTestFindingException.class),
+                    cause(instanceOf(FuzzerSecurityIssueMedium.class)))));
+
+    try (Stream<Path> entries =
+        Files.list(baseDir).filter(f -> f.getFileName().toString().startsWith("crash-"))) {
```
Marcono1234, line 150: Should this rather list the files in `inputsDirectory` instead of `baseDir`? (or otherwise what is the `inputsDirectory` field used for?)

Maybe to be safe should also assert that `inputsDirectory` actually contains at least one crash file?

oetr: As soon as you set the base dir, as we do here: `.configurationParameter("jazzer.internal.basedir", baseDir.toAbsolutePath().toString())`, Jazzer will write all crash files into it.
The test fails if you remove the fix precisely because there will be a 0-bytes crash file in the `baseDir`.

oetr: But it's a good idea to check in addition that there is exactly one crash file in the `baseDir` with more than 0 bytes. I'll add it.

Marcono1234: Ah ok, but what is the purpose of the `inputsDirectory` field in this test then?

examples/junit/src/test/java/com/example/FuzzTestWithCrashTest.java
```diff
@@ -0,0 +1,32 @@
...
+import com.code_intelligence.jazzer.api.FuzzerSecurityIssueMedium;
+import com.code_intelligence.jazzer.junit.FuzzTest;
+import java.util.Objects;
+
+public class FuzzTestWithCrashTest {
+  @FuzzTest(maxDuration = "10s")
+  void crashFuzz(FuzzedDataProvider data) {
+    String input = data.consumeRemainingAsString();
+    if (Objects.equals(input, "crash")) {
+      throw new FuzzerSecurityIssueMedium();
```
Marcono1234: This test here might not reproduce the originally reported bug.

It is necessary that the fuzz test calls `Jazzer#reportFindingFromHook`[^1] (respectively uses a sanitizer / hook which does this), because only then `JazzerInternal#lastFinding` is set. Directly throwing an exception here will not cause that.

At least that is my understanding of it.

[^1]: Based on the documentation is sounds as if fuzz tests should not directly call `reportFindingFromHook`. In that case maybe one of the simple sanitizer test cases (e.g. for Regex or Unsafe) could be copied here?

src/test/java/com/code_intelligence/jazzer/junit/FuzzingAfterRegressionCrashTest.java
oetr: It's unused, and now I've deleted it, nice!

examples/junit/src/test/java/com/example/FuzzTestWithCrashTest.java
oetr: Good catch! I started out with the regexp file, but converged to manual finding for other reasons, and haven't tested it correctly. I added the original fuzz test back in.

Issue #931, opened 2025-06-09
Bug: `FuzzedDataProvider#consume...` returns wrong results when `max - min == MAX_VALUE`


Marcono1234:
## Version
Jazzer JUnit 0.24.0

## Description
It seems the `FuzzedDataProvider` methods for producing a value within a `[min, max]` range, such as `consumeInt`, return results outside that range when `max - min == MAX_VALUE`.

The simplest case is something like this:
```java
@FuzzTest
void test(FuzzedDataProvider dataProvider) {
    var value = dataProvider.consumeInt(0, Integer.MAX_VALUE);
    if (value < 0) {
        throw new RuntimeException("value: " + value);
    }
}
```
`min` is 0 so the value should never be `< 0`, yet it does return results which are negative.

To highlight that this is not due to numeric overflow or related to `max` being `MAX_VALUE`, consider this example:
```java
@FuzzTest
void test(FuzzedDataProvider dataProvider) {
    int diff = Byte.MAX_VALUE;
    int min = -10;
    int max = min + diff;
    
    var value = dataProvider.consumeByte((byte) min, (byte) max);
    if (value < min) {
        throw new RuntimeException("value: " + value);
    }
}
```
It fails in a similar way, but if you change it to `diff = Byte.MAX_VALUE + 1` or `diff = Byte.MAX_VALUE - 1` it does not fail anymore.

----

The cause _might_ be this check here, not sure why it exists:
https://github.com/CodeIntelligenceTesting/jazzer/blob/efbc6354e412ce221ad3b18a6fdd32bf12241825/src/main/native/com/code_intelligence/jazzer/driver/fuzzed_data_provider.cpp#L114
Maybe this is supposed to prevent overflow for the `result` variable, but contains a bug and should rather check `uint64_t::max()` (`uint64_t` being the type of `result`) instead of `T::max()` (which is the `MAX_VALUE` of the Java type?)?

Anchels:
I have the same issue

PR #945, opened 2025-08-13
fix: make sure ConsumeIntegralInRange is always in range [min; max]

oetr:


oetr:
> Can this change the result in any case in which it previously fell within the given bounds? If possible, we should avoid invalidating existing crashing inputs that behaved as expected.

Ahh, I see what you mean, good point. And yes, it can and it will!
Obviously, this will still invalidate the crashes due to the bug addressed in this PR. 
For ranges [0; N] (the most frequent in my opinion!) my change does nothing different.
For all other ranges it might invalidate the crashing input.

My motivation for leaving the number as is, if it's already within the requested range `[min; max]`, was to make the fuzzer more efficiently use the table of recent compares. 
The way I understand it, libfuzzer keeps recently compared numbers in a table and randomly inserts them during mutation steps.
If these numbers are reduced modulo some range, ToRC becomes mostly useless for ranges other than `[0;N]`.
The further away the range deviates from `[0; N]`, the more useless the ToRC becomes (IMO).

So, my claim is that this change makes fuzzing more efficient, but I don't have the data!

A compromise solution that considers your point and fixes the bug at the same time would be to always squeeze (aka remove the if-statement)
```c++
// if (range != std::numeric_limits<T>::max()) 
// We accept modulo bias in favor of reading a dynamic number of bytes as
// this would make it harder for the fuzzer to mutate towards values from
// the table of recent compares.
result = result % (range + 1);

return static_cast<T>(min + result);
```

> If possible, we should avoid invalidating existing crashing inputs that behaved as expected.

It's inconvenient, but can be addressed by fixing the issues before upgrading to the new Jazzer release, or by running the fuzzer and letting it find the issue again.

WDYT?

src/main/native/com/code_intelligence/jazzer/driver/fuzzed_data_provider.cpp
```diff
@@ -111,13 +111,13 @@ T JNICALL ConsumeIntegralInRange(JNIEnv &env, jobject self, T min, T max) {
...
 
-  if (range != std::numeric_limits<T>::max())
+  if (result < min || result > max) {
     // We accept modulo bias in favor of reading a dynamic number of bytes as
     // this would make it harder for the fuzzer to mutate towards values from
     // the table of recent compares.
-    result = result % (range + 1);
-
-  return static_cast<T>(min + result);
+    result = min + (result % (range + 1));
```
Marcono1234: Is this `range + 1` here a potential problem when users call `consumeLong(Long.MIN_VALUE, Long.MAX_VALUE)`? I assume this is why this `range != ` check existed in the first place.

But as mentioned in #931, the previous use of `std::numeric_limits<T>::max()` might be wrong, and might have been the underlying cause of the bug.

Would a simpler (and more backward compatible) fix be to just change this line:
```diff
- if (range != std::numeric_limits<T>::max())
+ if (range != std::numeric_limits<uint64_t>::max())
```

(not sure if that works, or even compiles, I am not that familiar with C++)

oetr: Numbers are hard! 
` +1` is definitely a mistake in this PR, but not in my first comment of the linked issue. I dediced to refactor it to have a single return :smile: 
Thinking about it a bit more, I think the original intent was this:
```c++
if (range != static_cast<uint64_t>(std::numeric_limits<T>::max()) -
             static_cast<uint64_t>(std::numeric_limits<T>::min()))
   // reduce modulo
```

Or maybe better:
```c++
if (min == std::numeric_limits<T>::min() && max == std::numeric_limits<T>::max())
  // return
```

But I still think that this version is better for fuzzing (it exits before it can run into the `+1` overflow problem):
```c++
if (result >= min && result <= max) {
    return static_cast<T>(result);
}
return static_cast<T>(min + (result % (range + 1)));
```

Marcono1234: Ah nevermind, sorry, I think my concern regarding `result + 1` `uint64_t` overflow for `consumeLong(Long.MIN_VALUE, Long.MAX_VALUE)` is irrelevant due to the `if (result < min || result > max)` check which you added above. Because neither `result < min` nor `result > max` would be true in that case.

oetr: The only thing left to check is whether this change boosts the fuzzing performance so much that it is well worth it to invalidate all corpus and many crash files for all fuzz targets that use `Consume<Byte,Char,Short,Int,Long>`. I need to run some experiments.

oetr: And here is the long-awaited data! I used this simple fuzz test, where I get the parameters from the environmental variables. Each fuzz test is run for at most 1mil inputs.
```java
@FuzzTest
  public void testBytes(FuzzedDataProvider fdp) {
    byte target = getTargetNumberGeneric("FDP_TARGET", Byte.class);
    byte from = getTargetNumberGeneric("FDP_FROM", Byte.class);
    byte to = getTargetNumberGeneric("FDP_TO", Byte.class);
    byte data = fdp.consumeByte(from, to);
    if (data == target) {
      throw new FuzzerSecurityIssueLow("FDP_TARGET matched: " + target);
    }
  }
```

Here is the python loop from which I run this fuzz test:
```python
stats = []
for window_size in range(10, 128, 10):
    for window_start in range(-128, 127-window_size, 10):
        seed = random.randint(1, 100000)
        # sample a random target in the given window (inclusive on both ends)
        target = random.randint(window_start, window_start + window_size)
        stats.append(execTest(seed, target, window_start, window_start + window_size))
```
**Before this PR**:
before-scatter-1755214269


**After this PR**:
after-scatter-1755195537

> 
> The "After" results behave as I would expect---with bigger window, the fuzzer has harder time to get the correct number, because it has to be placed right in the beginning of fuzzer data _and_ be correct number from an increasingly larger range.
> 
>  The "Before" image shows my expectation that `base + x % range` operation renders the table of recent comparse useless. The results Before and After converge at full range, or for ranges [0; N].
> Initially, I had the number of runs set to 10mil, but it took 7 seconds for a data point that still didn't hit the target.
> Basically, in this graph, 1mil is a cutoff, where the fuzz test hasn't guessed the number correctly.
> 
> My hunch is that most ranges will be either untouched, or start from 0, for which this PR will not change anything. But for others it will make fuzzing ridiculously more efficient for targets where the numbers from FDP are compared with other numbers that guard some code.
> 
> @fmeum do you see any other issue wiht this?

Update: I couln't believe that libfuzzer would need more than 10mil tries to get a number in a range of 10. And couldn't reproduce my own experiment. The updated version for `ConsumeByte` is almost indistinguishable.

fmeum: As far as I can tell this will preserve corpus entries for ranges starting at 0? I think that's good enough regarding backwards compatibility. 

Great work on the benchmark 🙂

Not having the mutator constrained by backwards compatibility was one of the major reasons for building the mutator framework. If someone' fuzz test is broken by this PR, we should point them to that.

fmeum (approved):

src/main/native/com/code_intelligence/jazzer/driver/fuzzed_data_provider.cpp
oetr: There is actually a bug in my implementation! It makes no sense comparing `uint64_t result` with `T min/max` because `T`, being a java number, is always signed. The comparison only works if min and max are both positive. `result` should be casted to `T` before comparison. I am curious if copilot review can find it.
Running the benchmark for java integers, we get a clearer picture of which method is better. I will change this PR to 3). And even though 4) looks much better (because now the ToRC can work properly), I am not feeling well completely invalidating everything!

|1) Status quo||
|:-:|:-:|
|||
|2) Buggy because of `result >= min...`||
| ||
|3) Fixed with `(T) result >= min...`||
| ||
|4) Just for fun: always read as many bytes as `sizeof(T)` |and change the byte order of data stream to LSB first|
| ||
|5) Same as above but ylim scaled to 1mil|
| ||

oetr: > Not having the mutator constrained by backwards compatibility was one of the major reasons for building the mutator framework. If someone' fuzz test is broken by this PR, we should point them to that.

I am actually curious how our mutator fares against this. I will set up some experiments as well!

oetr: And here are the results for the mutation framework. I had to hack Jazzer a bit to get min/max configurable post-compilation, because `@InRange` is static. Here is the fuzz test
```java
  @FuzzTest
  public void testBug(int data) {
    int target = getTargetNumberGeneric("FDP_TARGET", Integer.class);
    if (data == target) {
      throw new FuzzerSecurityIssueLow("FDP_TARGET matched: " + target);
    }
  }
```

And the good news is that we are better than FDP. But I am surprised that it takes the same number of runs to reach the target for very small window sizes. Basically the number of runs seems to be independent of window size.
|Mutation framework results||
|:-:|:-:|
|||

PR #952, opened 2025-09-15
fix(mutation): ensure maxSize >= size when using libfuzzer

oetr:
Since users are free to change fuzz test signature, readExclusive might return byte arrays longer than maxLength. Calling libfuzzer's MutateOne with maxSize that is less than the length of the array to be mutated might result in a segfault at the time when the data type is used in Java. Some of libfuzzer's mutatos (Mutate_AddWordFromTORC) are not designed to handle this case and will mutate past the array bounds.


src/main/java/com/code_intelligence/jazzer/mutation/mutator/libfuzzer/LibFuzzerMutatorFactory.java
```diff
@@ -113,8 +113,13 @@ private int maxInitialSize() {
...
     @Override
     public byte[] mutate(byte[] value, PseudoRandom prng) {
-      int maxLengthIncrease = maxLength - value.length;
-      byte[] mutated = LibFuzzerMutate.mutateDefault(value, maxLengthIncrease);
+      // Ensure that maxSize is always >= value.length to avoid negative size increase.
+      // Calling to libFuzzer's Mutate function with maxSize < value.length will eventually result
+      // in a segfault at time of use in Java.
+      byte[] mutated =
+          maxLength < value.length
+              ? LibFuzzerMutate.mutateDefault(Arrays.copyOf(value, maxLength), 0)
```
fmeum: Could we perform this truncation in `readExclusive` instead? It's a simpler mental model if `mutate` can assume a valid value. I would still prefer keeping a `checkState` here to protect against bugs turning into memory access violations.

oetr: I changed the fix, Even though this goes against the original inventor's idea for readExclusive :wink: 
In the long term, I think readExclusive should not be used at all by mutators reading from corpus inputs.
The need for read/writeExclusive will naturally go away if we ditch libfuzzer in favor of a Java-only fuzzer.

oetr: Also modifying readExclusive is a breaking change, that might make some crashing inputs non-reproducible after this.
The first version of this fix only changed the mutation function and was non-breaking.

fmeum: The original inventor may very well have forgotten some details 🙂 One reason to have readExclusive was that it preserved the canonical corpus entry representations of single protos and byte arrays, which may still be relevant after sourcing libfuzzer's mutation function. 

I find it difficult to weigh backwards compatibility against avoiding JVM crashes, but I think I personally slightly prefer the fix in readExclusive. Feel free to arrive at a different conclusion though!

src/main/java/com/code_intelligence/jazzer/mutation/support/InputStreamSupport.java
```diff
@@ -60,6 +60,41 @@ public static byte[] readAllBytes(InputStream stream) throws IOException {
...
+    require(n >= 0, "n must be non-negative");
+    Queue<byte[]> buffers = new ArrayDeque<>();
+    int arrayLength = 0;
+    outer:
+    while (true) {
+      byte[] buffer = new byte[max(8192, stream.available())];
+      buffers.add(buffer);
+      int off = 0;
+      while (off < buffer.length) {
+        int bytesRead = stream.read(buffer, off, buffer.length - off);
```
simonresch: If I understand correctly, this function reads more than `n` bytes from the stream since only the buffer length is passed in the `stream.read()` function. This would lead the stream in an "unpredictable" state since its not clear how many bytes between `[n, num_bytes_in_stream]` have been read.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/libfuzzer/LibFuzzerMutatorFactory.java
```diff
@@ -71,7 +71,7 @@ public byte[] read(DataInputStream in) throws IOException {
 
     @Override
     public byte[] readExclusive(InputStream in) throws IOException {
-      return readAllBytes(in);
+      return readNBytes(in, maxLength);
```
simonresch: The current contract of `readExclusive` states that it consumes the remaining input stream which would not be the case if we read only `maxLength` bytes. Maybe instead of `readNBytes` we could read all bytes as before and truncate the byte array here?

oetr: `readExclusive` means that this is the last operation on given stream. Why should we read all bytes to satisfy some contract if we only need N bytes? The stream being in "unpredictable" state is not our concern, because it will only ever be touched by the garbage collector. Now that we've arrived at this solution, I prefer fuzzing efficiency+updating the contract (readExclusive is the last operation on the stream). WDYT?

simonresch: I suspect the performance overhead would be minimal since we are usually working on a `ByteArrayInputStream` where the data will be read with a single `read` operation. Personally, I would prefer the cleaner API of consuming all the available data in `readExclusive` unless there is a non-negligible performance overhead associated with it, but don't feel too strongly about it.

> The stream being in "unpredictable" state is not our concern, because it will only ever be touched by the garbage collector.

I agree that as the code currently is it will should not break things but having a `readNBytes` function that seems to do the same as https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/InputStream.html#readNBytes(byte%5B%5D,int,int) but in fact reads more bytes (silently dropped) could cause problems in the future if re-used.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/libfuzzer/LibFuzzerMutatorFactory.java
oetr: I always appreciate your feedback, thanks!

simonresch (approved):

PR #958, opened 2025-10-01
improve the docs

oetr:


README.md
```diff
@@ -75,15 +92,50 @@ class ParserTests {
     void fuzzTest(@NotNull String input) {
         assertEquals(input, SomeScheme.decode(SomeScheme.encode(input)));
     }
+	
+    @FuzzTest
+    void fuzzTest(@NotNull @WithUtf8Length(min=10, max=200) String input, @InRange(min=-10, max=10) int n) {
```
HenrichN: Doesn't this function need to have a different name as one function `fuzzTest` already exists?

README.md
```diff
@@ -45,26 +30,58 @@ Jazzer currently supports the following platforms:
...
-   In this mode, only a single fuzz test is executed per test run (see [#599](https://github.com/CodeIntelligenceTesting/jazzer/issues/599) for details).
-5. Run the fuzz test without `JAZZER_FUZZ` set to execute it only on the inputs in the inputs directory.
-   This mode, which behaves just like a traditional unit test, ensures that issues previously found by the fuzzer remain fixed and can also be used to debug the fuzz test on individual inputs.
+## Writing fuzz tests
 
-A simple property-based fuzz test could look like this:
+To write a fuzz test, add a method to your test class and annotate it with [`@FuzzTest`](https://codeintelligencetesting.github.io/jazzer-docs/jazzer-junit/com/code_intelligence/jazzer/junit/FuzzTest.html).
+Jazzer will automatically generate and mutate inputs for your method parameters.
+You can use primitives, strings, arrays, and many standard library classes—see the [mutation framework documentation](docs/mutation-framework.md#supported-types) for details.
+
```
HenrichN, line 80: I would like to add directly here that you need to set the environment variable  `JAZZER_FUZZ` to `1`  to make it very clear and easier for beginners. Otherwise I fear that people will miss it. I know that it is mentioned a few paragraphs  down, but I think it is worth to mention it here as well.

README.md
```diff
@@ -75,15 +92,50 @@ class ParserTests {
...
+### Using JUnit Integration
+To run fuzz tests, use your build system as you would for regular tests.
+Methods annotated with `@FuzzTest` can be executed in two modes: regression mode and fuzzing mode.
+
+Jazzer manages inputs for fuzz tests using an *inputs directory*.
+By default, this directory is derived from the package and class name of the fuzz test.
+For example, if the fuzz test is in the class `src/test/java/com/example/ValidFuzzTestsInputs.java`, method `byteFuzz`, the crash files directory is located in `src/test/resources/com/example/ValidFuzzTestsInputs/byteFuzz`.
+
+If this directory does not exist, Jazzer will save crash inputs in the directory from which the tests are executed.
```
HenrichN: I would move this section below the section fuzzing mode, perhaps even create a seperate section called "Inputs" or "Data directories" or something along the lines.

HenrichN (changes requested):
Thanks for the thorough work! I left some comments especially for the "starting section".

README.md
oetr: It works due to different signatures, but giving them unique names will make this example clearer, so I am changing the names.

README.md
```diff
@@ -75,15 +92,50 @@ class ParserTests {
...
+        assertTrue(n >= -10 && n <= 10);
+    }
 }
 ```
+A complete Maven example project can be found in [examples/junit](examples/junit).
+
+
+## Running Jazzer
+
+Jazzer can be run in two ways: using the JUnit integration or via standalone Jazzer binary.
```
simonresch: That leaves out running the `com.code_intelligence.jazzer.Jazzer.main()` method, right?
IMO that could be the documented way to run Jazzer in "standalone" mode and then in the advanced section we could have documentation on running the native binary directly. Or is it a common enough workflow to run the binary directly?

README.md
```diff
@@ -75,15 +92,50 @@ class ParserTests {
...
+
+### Using JUnit Integration
+To run fuzz tests, use your build system as you would for regular tests.
+Methods annotated with `@FuzzTest` can be executed in two modes: regression mode and fuzzing mode.
+
+Jazzer manages inputs for fuzz tests using an *inputs directory*.
+By default, this directory is derived from the package and class name of the fuzz test.
+For example, if the fuzz test is in the class `src/test/java/com/example/ValidFuzzTestsInputs.java`, method `byteFuzz`, the crash files directory is located in `src/test/resources/com/example/ValidFuzzTestsInputs/byteFuzz`.
```
simonresch: Could change "crash files directory" to "inputs directory" to keep the terminology more consistent?

README.md
```diff
@@ -99,29 +151,218 @@ If you see an error saying that `libjvm.so` has not been found, make sure that `
...
+- `META-INF/MANIFEST.MF` attribute `Jazzer-Some-Opt` on the classpath
+- the `JAZZER_SOME_OPT` environment variable
+- the `jazzer.some_opt` system property
+- the `jazzer.some_opt` JUnit configuration parameter (e.g. in `resources/junit-platform.properties` of the test sources)
+- the `--some_opt` CLI parameter to Jazzer standalone
+
+Some parameters only have an effect when used with standalone Jazzer binary (marked as *standalone only*), or only in fuzzing mode (marked as *fuzzing only*).
+
+
+### Options
```
simonresch: IMO the extensive list of available options could be moved to a dedicated file since many of them should be rarely used. The rest of the main `README.md` can be nicely read from top to bottom but going through every option is probably a bit much.  WDYT?

README.md
```diff
@@ -99,29 +151,218 @@ If you see an error saying that `libjvm.so` has not been found, make sure that `
...
+  
+- **custom_hook_includes** [list, separator=':', default=""]
+  - Glob patterns matching names of classes to instrument with hooks (custom and built-in)
+  
+- **dedup** [bool, default="true"]
+  - Compute and print a deduplication token for every finding
+  
+- **disabled_hooks** [list, separator=':', default=""]
+  - Names of classes from which hooks (custom or built-in) should not be loaded from
+  - Example: to disable the `ServerSideRequestForgery` and `RegexInjection` sanitizers use this environmental variable when running Jazzer:
```
simonresch: `environmental` -> `environment`

README.md
```diff
@@ -75,15 +92,50 @@ class ParserTests {
...
+#### Regression mode
+In regression mode, Jazzer runs each fuzz test on all inputs found in its inputs directory.
+This mode behaves like a traditional unit test: it verifies that previously discovered issues remain fixed and helps debug the fuzz test with specific inputs.
+By default, Jazzer operates in regression mode unless fuzzing mode is explicitly enabled.
 
-A complete Maven example project can be found in [`examples/junit`](examples/junit).
+#### Fuzzing mode
+Enable fuzzing mode by setting the environment variable `JAZZER_FUZZ=1` before running your tests.
+Jazzer will execute a single fuzz test, automatically generating and mutating inputs to maximize code coverage and find bugs.
+Any input that triggers a failure is saved to the corpus directory, which is derived from the fuzz test class and method.
```
simonresch: This seems to contradict the sentence in line 132 that a crashing input is stored in the "inputs directory". You probably meant something like "Any input that triggers new code paths..." or similar?

The two different directories could be a bit confusing to a first time reader. IMO here would be a good place to make it clear that there are two different directories and what files go where.

README.md
```diff
@@ -45,26 +30,58 @@ Jazzer currently supports the following platforms:
...
+You can use primitives, strings, arrays, and many standard library classes—see the [mutation framework documentation](docs/mutation-framework.md#supported-types) for details.
+
+Here is an example that demonstrates fuzzing security-relevant logic:
 
 ```java
+import com.code_intelligence.jazzer.mutation.annotation.InRange;
+import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.*;
```
simonresch, line 90: The `@WithUtf8Length` import seems to be missing. IMO skipping the imports would also be fine.

README.md
```diff
@@ -99,29 +151,218 @@ If you see an error saying that `libjvm.so` has not been found, make sure that `
...
+  - Comma separated list of jar files to instrument.
+  No fuzzing is performed.
+  
+- **instrumentation_excludes** [list, separator=':', default=""]
+  - Glob patterns matching names of classes that should not be instrumented for fuzzing
+  See [here](docs/advanced.md#coverage-instrumentation) for more details.
+  
+- **jvm_args** [list, separator=':', default=""] (*standalone only*)
+  - Arguments to pass to the JVM (separator can be escaped with `\\`).
+  See [here](docs/advanced.md#passing-jvm-arguments) for more details.
```
simonresch: This option seems identical to `additional_jvm_args` from reading the docs. Is there a reason to use one or the other?

README.md
```diff
@@ -99,29 +151,218 @@ If you see an error saying that `libjvm.so` has not been found, make sure that `
...
+  See [here](docs/advanced.md#trace-instrumentation) for more details.
+  
+- **ubsan** [bool, default="false"]
+  - Allow fuzzing of native libraries compiled with `-fsanitize=undefined` used for JUnit fuzz tests.
+  See [here](docs/advanced.md#native-libraries) for more details.
+  
+- **version** [bool, default="false"]
+  - Print version information
+
+## Sanitizers
```
simonresch: What do you think about adding a high level explanation what a sanitizer in Jazzer is? The terminology is overloaded when coming from the C++ language. Also, would it make sense to use the "bug detector" instead for external docs? :thinking: 

But love that they are now getting some exposure in the docs. :+1:

README.md
```diff
@@ -99,29 +151,218 @@ If you see an error saying that `libjvm.so` has not been found, make sure that `
...
+All Jazzer sanitizers can be found [here](sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers).
+Each sanitizer can be disabled using `disabled_hooks`, as shown above in [options](#options).
+
+
+### Configure Sanitizers using BugDetectorsAPI
+Currently only `ServerSideRequestForgery` and `FilePathTraversal` can be configured at runtime using the `BugDetectorsAPI`.
+See the [corresponding javadocs](https://codeintelligencetesting.github.io/jazzer-docs/jazzer-api/com/code_intelligence/jazzer/api/BugDetectors.html) for details.
+
+
+## Custom hooks
```
simonresch: IMO this could go in the `advanced.md` file.

docs/dev-implementation-details.md
```diff
@@ -0,0 +1,122 @@
+## Implementation
```
simonresch: :+1: for splitting this off into a dedicated file.

The heading could mention that this is about the JUnit integration.
Minor: the headings here could also loose one `#`.

README.md
HenrichN: I second that. I was also thinking about moving it to a different file.

oetr: I moved them to the top README for better visibility, but looking at how long they make it, it's probably better to move and reference.
The advantage of moving is also that now the arg description can be as long as it needs to be!

README.md
oetr: The way I read it in the code, they are treated the same way, and concatenated one after another, and passed to the JVM.

README.md
oetr: Moved the description from advanced.md to here.

README.md
```diff
@@ -97,31 +168,70 @@ You can also use GitHub release archives to run a standalone Jazzer binary that
...
+## Directories and files
 
-### Bazel
+Jazzer uses two directories to store inputs: the *generated corpus directory* and the *inputs directory*.
+
+### Generated corpus directory
+
+The *generated corpus directory* is where Jazzer saves inputs that generate new coverage during fuzzing.
+It is located in `.cifuzz-corpus/<package>/<FuzzTestClass>/<fuzzTestMethod>`, where `<package>`, `<FuzzTestClass>`, and `<fuzzTestMethod>` correspond to the package name, class name, and method name of the fuzz test, respectively.
+For example, if the fuzz test is in the class `src/test/java/com/example/ValidFuzzTestsInputs.java`, method `byteFuzz`, the corpus directory is located in `.cifuzz-corpus/com.example.ValidFuzzTestsInputs/byteFuzz`.
```
Marcono1234, line 179: Does it really make sense that the example test class name ends with `Inputs` here? That might only cause confusion with the generated `...Inputs` directory (next section).

```suggestion
For example, if the fuzz test is in the class `src/test/java/com/example/ValidFuzzTests.java`, method `byteFuzz`, the corpus directory is located in `.cifuzz-corpus/com.example.ValidFuzzTests/byteFuzz`.
```

README.md
```diff
@@ -97,31 +168,70 @@ You can also use GitHub release archives to run a standalone Jazzer binary that
...
+
+
+### Inputs directory
 
-Support for Jazzer is available in [rules_fuzzing](https://github.com/bazelbuild/rules_fuzzing), the official Bazel rules for fuzzing.
-See [the README](https://github.com/bazelbuild/rules_fuzzing#java-fuzzing) for instructions on how to use Jazzer in a Java Bazel project.
+Any input that triggers a crash during fuzzing is saved to the *inputs directory*.
+This directory is derived from the package and class name of the fuzz test.
+For example, if the fuzz test is in the class `src/test/java/com/example/ValidFuzzTestsInputs.java`, method `byteFuzz`, the *inputs directory* is located in `src/test/resources/com/example/ValidFuzzTestsInputs/byteFuzz`.
```
Marcono1234, line 186: Is the class name here correct? Or would that lead then to `...InputsInputs/` (and also the Maven Surefire Plugin not recognizing it by default due to the name not ending with `Test` / `Tests`).

```suggestion
For example, if the fuzz test is in the class `src/test/java/com/example/ValidFuzzTests.java`, method `byteFuzz`, the *inputs directory* is located in `src/test/resources/com/example/ValidFuzzTestsInputs/byteFuzz`.
```

README.md
```diff
@@ -97,31 +168,70 @@ You can also use GitHub release archives to run a standalone Jazzer binary that
...
+If this directory does not exist, Jazzer will save crash inputs in the directory from which the tests are executed.
```
Marcono1234, line 187: Should this mention that the input files are binary and care should be taken when commiting them to Git?
Basically a short version of this:
https://github.com/CodeIntelligenceTesting/jazzer/blob/e5057ce5be7b498e1b35ea14717f499bbc1e5dd6/docs/dev-junit-implementation-details.md?plain=1#L86-L90

README.md
oetr: Nice catch! I updated the readme here: https://github.com/CodeIntelligenceTesting/jazzer/pull/965

README.md
oetr: Thanks, I put it more prominently in the main README now: https://github.com/CodeIntelligenceTesting/jazzer/pull/965

PR #959, opened 2025-10-02
feat: add additional number annotations

florianGla:
This PR adds support for these additional number annotations:  `@Positive`, `@Negative`, `@NonPositive`, `@NonNegative`.
Note that it is not checked if multiple annotations contain contradicting information.


src/main/java/com/code_intelligence/jazzer/mutation/support/RangeSupport.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+    return new FloatRange(minValue, maxValue, allowNaN);
+  }
+
+  public static DoubleRange resolveDoubleRange(
+      AnnotatedType type, double defaultMin, double defaultMax, boolean defaultAllowNaN) {
+    double minValue = defaultMin;
+    double maxValue = defaultMax;
+    boolean allowNaN = defaultAllowNaN;
+    for (Annotation annotation : type.getAnnotations()) {
```
oetr, line 133: In this for-loop (and the equivalents for floats and integrals), every annotation gets to move min/maxValue/allowNan. So if you have something like `@InRange(min=-10, max=10) @Positive int n`, n will be in the merged range [1; 10].
I think, each annotation should set /all/ variables associated with the range. In this way only the last annotation (closest to the type) will be actually active.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/PrimitiveArrayMutatorFactory.java
oetr: `inRangeFloat` and `inRangeDouble` are now unused.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/PrimitiveArrayMutatorFactory.java
```diff
@@ -145,38 +149,66 @@ private void extractRange(AnnotatedType type) {
...
+          {
+            LongRange r =
+                RangeSupport.resolveIntegralRange(
+                    type, (long) Short.MIN_VALUE, (long) Short.MAX_VALUE);
+            minRange = r.min;
+            maxRange = r.max;
+          }
           break;
         case "char":
           minRange = inRange.map(InRange::min).orElse((long) Character.MIN_VALUE);
```
oetr, line 168: I think, we should be able to set  `@InRange` for `Character` and `char` types (they are currently not in the list of `@AppliesTo`), so the old code always used Character.MIN/MAX

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/PrimitiveArrayMutatorFactory.java
```diff
@@ -145,38 +149,66 @@ private void extractRange(AnnotatedType type) {
...
+            DoubleRange r =
+                RangeSupport.resolveDoubleRange(
+                    type, Double.NEGATIVE_INFINITY, Double.POSITIVE_INFINITY, false);
+            minDoubleRange = r.min;
+            maxDoubleRange = r.max;
+            allowNaN = r.allowNaN;
+          }
           break;
         case "boolean":
           minRange = inRange.map(InRange::min).orElse(0L);
```
oetr: Existinc code bug: `InRange` is not applicable to booleans, so booleans should always set min and max to 0 and 1.

src/test/java/com/code_intelligence/jazzer/mutation/support/TypeSupportTest.java
```diff
@@ -28,13 +28,7 @@
...
 import static org.junit.jupiter.params.provider.Arguments.arguments;
 
-import com.code_intelligence.jazzer.mutation.annotation.DoubleInRange;
-import com.code_intelligence.jazzer.mutation.annotation.FloatInRange;
-import com.code_intelligence.jazzer.mutation.annotation.InRange;
-import com.code_intelligence.jazzer.mutation.annotation.NotNull;
-import com.code_intelligence.jazzer.mutation.annotation.WithLength;
-import com.code_intelligence.jazzer.mutation.annotation.WithSize;
-import com.code_intelligence.jazzer.mutation.annotation.WithUtf8Length;
+import com.code_intelligence.jazzer.mutation.annotation.*;
```
oetr: This looks like the default editor setting that transforms >N imports from the same class into star-imports. I think explicit imports improve clarity. This is done in all of Jazzer code so far.

oetr (commented):
Nice work!

Since all these new annotations can be used to annotate the same type, I think we really should enforce that only one of them is allowed, because all of them essentially map to `@(Float|Double)InRange` equivalent.
For example, currently, this results in the domain `[0:200]`:
```java
@FuzzTest
  public void fuzz_int(@Positive @InRange(min=-100, max=200) @NonNegative  int i) {
```

However, the Java type system does not let you add two `@InRange` annotations to the same type (because it is not annotated by `@Repeatable`).

TBH, I don't think we should be adding these annotations, because they add large complexity (especially if we want to add uniqueness enforcement for all annotation types that map to some other annotation type) with only little upside for the users (we can already specify these domains using `@InRange`, `@FloatInRange`, and `@DoubleInRange`). But we should discuss that.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/PrimitiveArrayMutatorFactory.java
florianGla: Do you think I should add this change in this PR or create a follow-up? I'm not sure how big this change would be?

src/main/java/com/code_intelligence/jazzer/mutation/support/RangeSupport.java
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  public static LongRange resolveIntegralRange(
+      AnnotatedType type, long defaultMin, long defaultMax) {
+    long minValue = defaultMin;
+    long maxValue = defaultMax;
+    for (Annotation annotation : type.getAnnotations()) {
+      if (annotation instanceof InRange) {
+        InRange inRange = (InRange) annotation;
+        if (inRange.min() != Long.MIN_VALUE) {
```
oetr: Since the other annotations might have changed `minValue` and `maxValue` this example `public void fuzz_char_in_range(@Positive @InRange(max=200) int i) {`
still results in domain `[1:200]`.
We should always assign, so there should either be an else (take `defaultValue`?), or just one assignment.
Do we have to check for `Long.MIN_VALUE` or can we directly assign `minValue = inRange.min()`?
Same for max

src/main/java/com/code_intelligence/jazzer/mutation/support/RangeSupport.java
```diff
@@ -0,0 +1,163 @@
...
+    float maxValue = defaultMax;
+    boolean allowNaN = defaultAllowNaN;
+    for (Annotation annotation : type.getAnnotations()) {
+      if (annotation instanceof FloatInRange) {
+        FloatInRange floatInRange = (FloatInRange) annotation;
+        minValue = floatInRange.min();
+        maxValue = floatInRange.max();
+        allowNaN = floatInRange.allowNaN();
+      } else if (annotation instanceof Positive) {
+        minValue = Math.max(minValue, Float.MIN_VALUE);
```
oetr: Why do we take a max from these two? minValue could have been modified by another annotation before. This happens in several places below as well.
I think we can directly assign `minValue = Float.MIN_VALUE`.

src/main/java/com/code_intelligence/jazzer/mutation/support/RangeSupport.java
```diff
@@ -0,0 +1,163 @@
...
+        maxValue = defaultMax;
+        allowNaN = false;
+      } else if (annotation instanceof Negative) {
+        minValue = defaultMin;
+        maxValue = Math.min(maxValue, -Float.MIN_VALUE);
```
oetr: `-Float.MIN_VALUE` equals `-1.4^-45`
You mean `-Float.MAX_VALUE` here `-3.4^36`

src/main/java/com/code_intelligence/jazzer/mutation/support/RangeSupport.java
```diff
@@ -0,0 +1,163 @@
...
+        minValue = doubleInRange.min();
+        maxValue = doubleInRange.max();
+        allowNaN = doubleInRange.allowNaN();
+      } else if (annotation instanceof Positive) {
+        minValue = Math.max(minValue, Double.MIN_VALUE);
+        maxValue = defaultMax;
+        allowNaN = false;
+      } else if (annotation instanceof Negative) {
+        minValue = defaultMin;
+        maxValue = Math.min(maxValue, -Double.MIN_VALUE);
```
oetr: Same as above, here you mean `-Double.MAX_VALUE`

src/main/java/com/code_intelligence/jazzer/mutation/support/RangeSupport.java
florianGla: Good catch!

src/main/java/com/code_intelligence/jazzer/mutation/support/RangeSupport.java
florianGla: You are right. I will have a look if the check is needed.

PR #961, opened 2025-10-06
fix: update regex for detecting intellij coverage agent

florianGla:


src/main/java/com/code_intelligence/jazzer/junit/Utils.java
```diff
@@ -222,7 +222,8 @@ public FileVisitResult preVisitDirectory(
   }
 
   private static final Pattern COVERAGE_AGENT_ARG =
-      Pattern.compile("-javaagent:.*(?:intellij-coverage-agent|jacoco).*");
+      Pattern.compile(
+          "-javaagent:.*(?:intellij-coverage-agent|intellij.platform.coverage.agent|jacoco).*");
```
simonresch: Minor: Could be cleaner to escape the `.`s in `intellij.platform.coverage.agent` if we don't expect any other characters there.

PR #967, opened 2025-10-10
fix: correct method descriptors

simonresch:
Adds a test to check that `MethodHook` annotations in sanitizers reference existing classes & methods.


sanitizers/src/test/java/com/code_intelligence/jazzer/sanitizers/HookBindingSanityTest.java
```diff
@@ -0,0 +1,186 @@
...
+                  "java.util.regex.Pattern$SingleI",
+                  "java.util.regex.Pattern$SingleS",
+                  "java.util.regex.Pattern$SingleU")
+              .collect(Collectors.toSet()));
+
+  final Set<String> SKIPPED_METHODS_CURRENT_JDK =
+      Collections.unmodifiableSet(
+          Stream.of(
+                  // Available for JDK >= 9
+                  "java.util.regex.Pattern.caseInsensitiveRangeFor",
```
Marcono1234: Isn't it rather the opposite, these classes and methods are **un**available for JDK >= 9?

```suggestion
                  // Unavailable for JDK >= 9
                  "java.util.regex.Pattern$Single",
                  "java.util.regex.Pattern$SingleI",
                  "java.util.regex.Pattern$SingleS",
                  "java.util.regex.Pattern$SingleU")
              .collect(Collectors.toSet()));

  final Set<String> SKIPPED_METHODS_CURRENT_JDK =
      Collections.unmodifiableSet(
          Stream.of(
                  // Unavailable for JDK >= 9
                  "java.util.regex.Pattern.caseInsensitiveRangeFor",
```

sanitizers/src/test/java/com/code_intelligence/jazzer/sanitizers/HookBindingSanityTest.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+    ClassLoader loader = HookBindingSanityTest.class.getClassLoader();
+    Class<?> targetClass =
+        assertDoesNotThrow(
+            () -> Class.forName(targetClassName, false, loader),
+            () -> "class to hook not found: " + targetClassName);
+    String methodName = hook.targetMethod();
+    String methodDesc = hook.targetMethodDescriptor();
+
+    if (SKIPPED_METHODS.contains(String.format("%s.%s", targetClassName, methodName))) {
```
Marcono1234: Would it make sense to include the method descriptor in this check as well? For example for the `Unsafe` methods only certain overloads have been removed in newer JDK versions, but not all.

sanitizers/src/test/java/com/code_intelligence/jazzer/sanitizers/HookBindingSanityTest.java
simonresch: Yes, of course. Thanks!

sanitizers/src/test/java/com/code_intelligence/jazzer/sanitizers/HookBindingSanityTest.java
simonresch: Absolutely right. I have to admit that was a bit of laziness on my part. ;) I revised the test a bit so that the method descriptor is also checked. WDYT?

Marcono1234: Looks really good! Though keep in mind that I am not a project member (maybe I should have clarified this, sorry), so please consider my comments to be at most suggestions.

simonresch: Yes, I'm aware. We will have to wait for one of my colleges for an "official" review. But I always appreciate good feedback. :)

sanitizers/src/test/java/com/code_intelligence/jazzer/sanitizers/HookBindingSanityTest.java
```diff
@@ -0,0 +1,240 @@
...
+
+  static Class<?> getHookClass(String className) {
+    try {
+      return Class.forName(className, false, HookBindingSanityTest.class.getClassLoader());
+    } catch (ClassNotFoundException e) {
+      throw new RuntimeException("Could not find hook class " + className, e);
+    }
+  }
+
+  static MethodHook[] getMethodHooks() throws ClassNotFoundException, NoSuchMethodException {
```
oetr: unused exceptions

sanitizers/src/test/java/com/code_intelligence/jazzer/sanitizers/HookBindingSanityTest.java
```diff
@@ -0,0 +1,240 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.code_intelligence.jazzer.sanitizers;
+
+import static org.junit.jupiter.api.Assertions.*;
```
oetr: Can you change the star import into individual imports?
I think, we need to add a style checker that complains about it.

oetr (approved):
Awesome work, especially with the tests!

sanitizers/src/test/java/com/code_intelligence/jazzer/sanitizers/HookBindingSanityTest.java
simonresch: Now I found the _second_ IntelliJ setting that grouped those together. :upside_down_face:

PR #973, opened 2025-10-10
feat: add linux arm64 support

### LINKED ISSUE #912: Feature Request: Release linux aarch64 binary
Right now it appears the only released Linux binary is for x86_64.  It would be great if the project was also compiled and released for linux aarch64 / arm.  Thank you!

simonresch:
Add support for running Jazzer on linux arm64 platforms.

Closes #912.

### DIFF
--- a/.github/workflows/prerelease.yaml
+++ b/.github/workflows/prerelease.yaml
@@ -13,7 +13,9 @@ jobs:
       matrix:
         include:
           - os: ubuntu-22.04
-            name: linux
+            name: linux-x86-64
+          - os: ubuntu-22.04-arm
+            name: linux-arm64
           - os: macos-14
             name: macos
           - os: windows-2022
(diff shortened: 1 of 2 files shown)


.github/workflows/prerelease.yaml
```diff
@@ -162,5 +164,6 @@ jobs:
           draft: true
           files: |
             _releases/jazzer-linux.tar.gz
```
simonresch: @oetr I didn't adjust the linux x86 artifact name so far to avoid a breaking change. WDYT? Should we just rename the artifact now to be consistent with the `jazzer-linux-arm64` artifact?

oetr: I think we should, and see what breaks 😆 

On another note, I am wondering why `win32-x86-64/attach_hotspot_windows.dll` and `win32-x86/attach_hotspot_windows.dll` are present in every release (or at least in the linux one)? They seem to be used by bytebuddy on windows.

PR #975, opened 2025-10-10
feat: add @ValuePool to propagate values to types

oetr:
This allows the users provide values directly to types by annotating them with `@ValuePool`.
All types, other than those that need cached-based mutators, are supported.

```java
  public static Stream<?> myPool() {
    return Stream.of(
        "0123456789abcdef".repeat(50),
        "sitting duck suprime".repeat(53),
        123);
  }

  @FuzzTest
  @ValuePool("myPool")
  public static void fuzzerTestOneInput(
      @NotNull @WithUtf8Length(max = 10000) String data,
      @NotNull @WithUtf8Length(max = 10000) String data2) {
    if (data.equals("0123456789abcdef".repeat(50))
        && data2.equals("sitting duck suprime".repeat(53))) {
      throw new FuzzerSecurityIssueLow("Found the long string!");
    }
  }
```

### DIFF
--- a/src/main/java/com/code_intelligence/jazzer/mutation/ArgumentsMutator.java
+++ b/src/main/java/com/code_intelligence/jazzer/mutation/ArgumentsMutator.java
@@ -23,6 +23,7 @@ import static java.lang.String.format;
 import static java.util.Arrays.stream;
 import static java.util.stream.Collectors.joining;
 
+import com.code_intelligence.jazzer.mutation.annotation.ValuePool;
 import com.code_intelligence.jazzer.mutation.api.ExtendedMutatorFactory;
 import com.code_intelligence.jazzer.mutation.api.PseudoRandom;
 import com.code_intelligence.jazzer.mutation.api.SerializingMutator;
@@ -31,6 +32,8 @@ import com.code_intelligence.jazzer.mutation.combinator.MutatorCombinators;
 import com.code_intelligence.jazzer.mutation.engine.SeededPseudoRandom;
 import com.code_intelligence.jazzer.mutation.mutator.Mutators;
 import com.code_intelligence.jazzer.mutation.support.Preconditions;
+import com.code_intelligence.jazzer.mutation.support.TypeSupport;
+import com.code_intelligence.jazzer.mutation.support.ValuePoolRegistry;
 import com.code_intelligence.jazzer.utils.Log;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
@@ -75,7 +78,7 @@ public final class ArgumentsMutator {
   }
 
   public static ArgumentsMutator forMethodOrThrow(Method method) {
-    return forMethod(Mutators.newFactory(), method)
+    return forMethod(Mutators.newFactory(new ValuePoolRegistry(method)), method)
         .orElseThrow(
             () ->
                 new IllegalArgumentException(
@@ -83,7 +86,7 @@ public final class ArgumentsMutator {
   }
 
   public static Optional<ArgumentsMutator> forMethod(Method method) {
-    return forMethod(Mutators.newFactory(), method);
+    return forMethod(Mutators.newFactory(new ValuePoolRegistry(method)), method);
   }
 
   public static Optional<ArgumentsMutator> forMethod(
@@ -97,11 +100,20 @@ public final class ArgumentsMutator {
       Log.error(validationError.getMessage());
       throw validationError;
     }
+
+    ValuePool[] valuePools = method.getAnnotationsByType(ValuePool.class);
+
     return toArrayOrEmpty(
             stream(method.getAnnotatedParameterTypes())
                 .map(
                     type -> {
-                      Optional<SerializingMutator<?>> mutator = mutatorFactory.tryCreate(type);
+                      // Forward all @ValuePool annotations of the fuzz test method to each
+                      // arg.
+                      AnnotatedType t = type;
+                      for (ValuePool pool : valuePools) {
+                        t = TypeSupport.withExtraAnnotations(t, pool);
+                      }
+                      Optional<SerializingMutator<?>> mutator = mutatorFactory.tryCreate(t);
                       if (!mutator.isPresent()) {
                         Log.error(
                             String.format(
--- a/src/main/java/com/code_intelligence/jazzer/mutation/annotation/ValuePool.java
+++ b/src/main/java/com/code_intelligence/jazzer/mutation/annotation/ValuePool.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2024 Code Intelligence GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.code_intelligence.jazzer.mutation.annotation;
+
+import static com.code_intelligence.jazzer.mutation.utils.PropertyConstraint.RECURSIVE;
+import static java.lang.annotation.ElementType.TYPE_USE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import com.code_intelligence.jazzer.mutation.utils.IgnoreRecursiveConflicts;
+import com.code_intelligence.jazzer.mutation.utils.PropertyConstraint;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/**
+ * Provides values to user-selected mutator types to start fuzzing from.
+ *
+ * <p>This annotation can be applied to fuzz test methods and any parameter type or subtype. By
+ * default, this annotation is propagated to all nested subtypes unless specified otherwise via the
+ * {@link #constraint()} attribute.
+ *
+ * <p>Example usage:
+ *
+ * <pre>{@code
+ * public class MyFuzzTargets {
+ *
+ *   static Stream<?> valuesVisibleByAllArgumentMutators() {
+ *     return Stream.of("example1", "example2", "example3", 1232187321, -182371);
+ *   }
+ *
+ *   static Stream<?> valuesVisibleOnlyByAnotherInput() {
+ *     return Stream.of("code-intelligence.com", "secret.url.1082h3u21ibsdsazuvbsa.com");
+ *   }
+ *
+ *   @ValuePool("valuesVisibleByAllArgumentMutators")
+ *   @FuzzTest
+ *   public void fuzzerTestOneInput(String input, @ValuePool("valuesVisibleOnlyByAnotherInput") String anotherInput) {
+ *     // Fuzzing logic here
+ *   }
+ * }
+ * }</pre>
+ *
+ * In this example, the mutator for the String parameter {@code input} of the fuzz test method
+ * {@code fuzzerTestOneInput} will be using the values returned by {@code
+ * valuesVisibleByAllArgumentMutators} method during mutation, while the mutator for String {@code
+ * anotherInput} will use values from both methods: from the method-level {@code ValuePool}
+ * annotation that uses {@code valuesVisibleByAllArgumentMutators} and the parameter-level {@code
+ * ValuePool} annotation that uses {@code valuesVisibleOnlyByAnotherInput}.
+ */
+@Target({ElementType.METHOD, TYPE_USE})
+@Retention(RUNTIME)
+@IgnoreRecursiveConflicts
+@PropertyConstraint
+public @interface ValuePool {
+  /**
+   * Specifies supplier methods that generate values for fuzzing the annotated method or type. The
+   * specified supplier methods must be static and return a {@code Stream <?>} of values. The values
+   * don't need to match the type of the annotated method or parameter. The mutation framework will
+   * extract only the values that are compatible with the target type.
+   */
+  String[] value();
+
+  /**
+   * This {@code ValuePool} will be used with probability {@code p} by the mutator responsible for
+   * fitting types.
+   */
+  double p() default 0.1;
+
+  /**
+   * Defines the scope of the annotation. Possible values are defined in {@link
+   * com.code_intelligence.jazzer.mutation.utils.PropertyConstraint}. By default it's {@code
+   * RECURSIVE}.
+   */
+  String constraint() default RECURSIVE;
+}
--- a/src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/StringMutatorFactory.java
+++ b/src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/StringMutatorFactory.java
@@ -17,7 +17,9 @@
 package com.code_intelligence.jazzer.mutation.mutator.lang;
 
 import static com.code_intelligence.jazzer.mutation.combinator.MutatorCombinators.mutateThenMapToImmutable;
-import static com.code_intelligence.jazzer.mutation.support.TypeSupport.*;
+import static com.code_intelligence.jazzer.mutation.support.TypeSupport.findFirstParentIfClass;
+import static com.code_intelligence.jazzer.mutation.support.TypeSupport.notNull;
+import static com.code_intelligence.jazzer.mutation.support.TypeSupport.withLength;
 
 import com.code_intelligence.jazzer.mutation.annotation.Ascii;
 import com.code_intelligence.jazzer.mutation.annotation.UrlSegment;
--- a/src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/ValuePoolMutatorFactory.java
+++ b/src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/ValuePoolMutatorFactory.java
@@ -1,0 +1,198 @@
+/*
+ * Copyright 2025 Code Intelligence GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.code_intelligence.jazzer.mutation.mutator.lang;
+
+import static java.lang.annotation.ElementType.TYPE_USE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import com.code_intelligence.jazzer.mutation.api.Debuggable;
+import com.code_intelligence.jazzer.mutation.api.ExtendedMutatorFactory;
+import com.code_intelligence.jazzer.mutation.api.MutatorFactory;
+import com.code_intelligence.jazzer.mutation.api.PseudoRandom;
+import com.code_intelligence.jazzer.mutation.api.SerializingMutator;
+import com.code_intelligence.jazzer.mutation.support.TypeHolder;
+import com.code_intelligence.jazzer.mutation.support.TypeSupport;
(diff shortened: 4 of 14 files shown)


tests/src/test/java/com/example/ValuePoolFuzzerLongString.java
```diff
@@ -0,0 +1,138 @@
...
+import com.code_intelligence.jazzer.mutation.annotation.DictionaryProvider;
+import com.code_intelligence.jazzer.mutation.annotation.NotNull;
+import com.code_intelligence.jazzer.mutation.annotation.WithSize;
+import com.code_intelligence.jazzer.mutation.annotation.WithUtf8Length;
+import java.util.List;
+import java.util.stream.Stream;
+
+public class DictionaryProviderFuzzerLongString {
+  private static final String str00 = repeat("0123456789abcdef", 50);
+  private static final String str01 = repeat("sitting duck suprime", 53);
```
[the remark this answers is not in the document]
oetr: joke

src/main/java/com/code_intelligence/jazzer/mutation/mutator/libfuzzer/LibFuzzerMutatorFactory.java
```diff
@@ -41,7 +42,8 @@ public final class LibFuzzerMutatorFactory {
   private static final int DEFAULT_MAX_LENGTH = 1000;
 
   @CheckReturnValue
-  public static Optional<SerializingMutator<?>> tryCreate(AnnotatedType type) {
+  public static Optional<SerializingMutator<?>> tryCreate(
+      MutatorRuntime runtime, AnnotatedType type) {
```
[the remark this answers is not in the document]
oetr: not yet

src/main/java/com/code_intelligence/jazzer/mutation/annotation/ValuePool.java
```diff
@@ -0,0 +1,95 @@
...
+@IgnoreRecursiveConflicts
+@PropertyConstraint
+public @interface DictionaryProvider {
+  /**
+   * Specifies supplier methods that generate dictionary values for fuzzing the annotated method or
+   * type. The specified supplier methods must be static and return a {@code Stream <?>} of values.
+   * The values don't need to match the type of the annotated method or parameter exactly. The
+   * mutation framework will extract only the values that are compatible with the target type.
+   */
+  String[] value() default {""};
```
simonresch: The error message could be better if we remove the default value which forces users to specify `value`.

src/main/java/com/code_intelligence/jazzer/mutation/annotation/ValuePool.java
```diff
@@ -0,0 +1,95 @@
...
+
+  /**
+   * This {@code DictionaryProvider} will be used with probability {@code 1/p} by the mutator
+   * responsible for fitting types. Not all mutators respect this probability.
+   */
+  int pInv() default 10;
```
simonresch: Would this make sense as a float setting?

src/main/java/com/code_intelligence/jazzer/mutation/support/DictionaryProviderSupport.java
```diff
@@ -0,0 +1,108 @@
...
+    return pInv;
+  }
+
+  /** Extract the provider streams using MutatorRuntime */
+  public static Optional<Stream<?>> extractRawValues(AnnotatedType type) {
+    DictionaryProvider[] providers =
+        Arrays.stream(type.getAnnotations())
+            .filter(a -> a instanceof DictionaryProvider)
+            .map(a -> (DictionaryProvider) a)
+            .toArray(DictionaryProvider[]::new);
```
simonresch: Wouldn't the `getAnnotationsByType` work here as well?

src/main/java/com/code_intelligence/jazzer/mutation/support/ValuePoolSupport.java
```diff
@@ -0,0 +1,108 @@
...
+            .map(a -> (DictionaryProvider) a)
+            .toArray(DictionaryProvider[]::new);
+    if (providers.length == 0) {
+      return Optional.empty();
+    }
+    HashSet<String> providerMethodNames =
+        Arrays.stream(providers)
+            .map(DictionaryProvider::value)
+            .flatMap(Arrays::stream)
+            .filter(name -> !name.isEmpty())
```
simonresch: Minor: Why filter out empty strings? IMO an error message that no such method is found would be cleaner.

src/main/java/com/code_intelligence/jazzer/mutation/support/ValuePoolSupport.java
```diff
@@ -0,0 +1,108 @@
...
+            .collect(Collectors.toCollection(HashSet::new));
+    if (providerMethodNames.isEmpty()) {
+      return Optional.empty();
+    }
+    Map<String, Method> fuzzTestMethods =
```
simonresch: The name of this variable threw me off when reading this. Could something like `dictionaryProviderMethods` or similar work? `fuzzTestMethods` sounds like the method annotated with `@FuzzTest`.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/StringMutatorFactory.java
```diff
@@ -198,4 +209,98 @@ public Optional<SerializingMutator<?>> tryCreate(
...
+              stream ->
+                  stream
+                      .flatMap(
+                          o -> {
+                            if (o instanceof String) {
+                              return Stream.of(((String) o).getBytes(StandardCharsets.UTF_8));
+                            } else {
+                              return Stream.empty();
+                            }
+                          })
```
simonresch: Maybe a dedicated `filter` and `map` step here could improve readability?

```
                      .filter(o -> o instanceof String)
                      .map(s -> ((String) s).getBytes(StandardCharsets.UTF_8))
```

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/StringMutatorFactory.java
```diff
@@ -198,4 +209,98 @@ public Optional<SerializingMutator<?>> tryCreate(
...
+      if (prng.trueInOneOutOf(pInv)) {
+        return prng.pickIn(dictionaryValues);
+      }
+      return basicMutator.init(prng);
+    }
+
+    @Override
+    public byte[] mutate(byte[] value, PseudoRandom prng) {
+      if (prng.trueInOneOutOf(pInv)) {
+        return prng.pickIn(dictionaryValues);
```
simonresch: This only completely replaces the current value with a dictionary value. WDYT about inserting into the current input or overwriting just part of it similar to libFuzzers implementation? https://github.com/llvm/llvm-project/blob/main/compiler-rt/lib/fuzzer/FuzzerMutate.cpp#L183

src/main/java/com/code_intelligence/jazzer/mutation/ArgumentsMutator.java
```diff
@@ -97,6 +98,8 @@ public static Optional<ArgumentsMutator> forMethod(
       Log.error(validationError.getMessage());
       throw validationError;
     }
+    MutationRuntime.fuzzTestMethod = method;
+    DictionaryProvider[] typeDictionaries = method.getAnnotationsByType(DictionaryProvider.class);
```
simonresch: Minor: This probably belongs to the next commit.

src/main/java/com/code_intelligence/jazzer/mutation/support/TypeSupport.java
```diff
@@ -578,6 +579,9 @@ private static Annotation[] checkExtraAnnotations(
               .collect(Collectors.toCollection(HashSet::new));
       for (Annotation annotation : extraAnnotations) {
         boolean added = existingAnnotationTypes.add(annotation.annotationType());
+        if (annotation.annotationType().isAnnotationPresent(IgnoreRecursiveConflicts.class)) {
```
simonresch, line 582: Since the entire purpose of this function seems to be checking for the recursive conflict maybe we can exit early in the first lines of this function if `IgnoreRecursiveConflicts.class` is set?

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/ArrayMutatorFactory.java
```diff
@@ -53,12 +55,16 @@ public Optional<SerializingMutator<?>> tryCreate(
 
     AnnotatedType elementType = ((AnnotatedArrayType) type).getAnnotatedGenericComponentType();
     AnnotatedType propagatedElementType = propagatePropertyConstraints(type, elementType);
-    Class<?> propagatedElementClazz = (Class<?>) propagatedElementType.getType();
-    return Optional.of(propagatedElementType)
-        .flatMap(factory::tryCreate)
-        .map(
-            elementMutator ->
-                new ArrayMutator<>(elementMutator, propagatedElementClazz, minLength, maxLength));
+    Type rawType = propagatedElementType.getType();
```
simonresch, line 58: Is this change relevant to the `DictionaryProvider`? If not, it could make sense to split it from this PR.

src/main/java/com/code_intelligence/jazzer/mutation/annotation/DictionaryProvider.java
```diff
@@ -0,0 +1,95 @@
...
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import com.code_intelligence.jazzer.mutation.utils.IgnoreRecursiveConflicts;
+import com.code_intelligence.jazzer.mutation.utils.PropertyConstraint;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/**
+ * Provides dictionary values to user-selected mutator types. Currently supported mutators are:
```
simonresch: I suspect that users might be confused on when to use the `DictionaryProvider` and when to use other dictionary annotations like `DictionaryEntry`, `DictionaryFile`.
"user-selected mutator types" might also not be descriptive enough and end with users annotating a `myFuzzTest(byte[] data)` fuzz test with `@DictionaryProvider` without realizing that the values are not used.

Would it make sense to add the `@DictionaryProvider` values in addition to the libFuzzer dictionary file? Otherwise I could foresee situations where one wants to specify the same dictionary entries with `@DictionaryProvider` _and_ `@DictionaryEntry`.

src/main/java/com/code_intelligence/jazzer/mutation/support/ValuePoolSupport.java
```diff
@@ -0,0 +1,108 @@
...
+        Arrays.stream(MutationRuntime.fuzzTestMethod.getDeclaringClass().getDeclaredMethods())
```
simonresch: Accessing global state in an unsuspecting `*Support` function is quite surprising IMO. Could we perform the "collect all possible dictionary provider methods" once before constructing the mutator and then pass those in (e.g. via mutator factory constructor)? Maybe even a dedicated mutator in the chain of possible mutators that only handles the custom dictionary and then delegates to an underlying type mutator?

src/main/java/com/code_intelligence/jazzer/mutation/support/TypeSupport.java
oetr: > `IgnoreRecursiveConflicts.class` is set

The purpose is to make sure that each annotation in `extraAnnotation` has no conflicts with existing annotations on the `base` type.
We can't exit early, because we have to check every single annotation individually.
Presence of `IgnoreRecursiveConflicts.class` on one of them doesn't change anything for the other annotations.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/ArrayMutatorFactory.java
oetr: It's a separate commit. Should be enough, no?

src/main/java/com/code_intelligence/jazzer/mutation/support/DictionaryProviderSupport.java
oetr: No, that would give us only the most immediate `ValuePool`, and not all of them.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/StringMutatorFactory.java
oetr: does not apply to the `ValuePoolMutatorFactory` version

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/StringMutatorFactory.java
oetr: We now use the value as a starting point for mutation with chance of 0.5.

src/main/java/com/code_intelligence/jazzer/mutation/ArgumentsMutator.java
oetr: does not apply to the `ValuePoolMutatorFactory` version

src/main/java/com/code_intelligence/jazzer/mutation/annotation/DictionaryProvider.java
oetr: - Renamed to `@ValuePool`
- Values are not really dictionary entries that libfuzzer limits to 64 bytes

I think, instead of pouring these values into a dictionary, we should always use the mutation framework even when the user only has `fuzzTestMethod(byte[] data)`, so that `ValuePool` will always work in a consistent manner. WDYT? We can fix that in another PR after discussing it

src/main/java/com/code_intelligence/jazzer/mutation/support/ValuePoolSupport.java
oetr: does not apply to the `ValuePoolMutatorFactory` version

src/main/java/com/code_intelligence/jazzer/mutation/support/TypeSupport.java
```diff
@@ -687,4 +690,21 @@ public static boolean annotatedTypeEquals(AnnotatedType left, AnnotatedType righ
     return left.getType().equals(right.getType())
         && Arrays.equals(left.getAnnotations(), right.getAnnotations());
   }
+
+  public static Optional<Class<?>> extractRawClass(Type type) {
+    if (type instanceof Class<?>) {
+      return Optional.of((Class<?>) type);
+    } else if (type instanceof ParameterizedType) {
+      return Optional.of((Class<?>) ((ParameterizedType) type).getRawType());
```
simonresch: The raw type of a `ParameterizedType` does not have to be a `Class<?>`, no? Maybe something line `return extractRawClass((ParameterizedType) type).getRawType());` would make sense?

src/main/java/com/code_intelligence/jazzer/mutation/annotation/ValuePool.java
```diff
@@ -0,0 +1,94 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/**
+ * Provides values to user-selected mutator types to start fuzzing from. Currently supported
+ * mutators are:
+ *
+ * <ul>
+ *   <li>String mutator
+ *   <li>Integral mutators (byte, short, int, long)
```
simonresch: Other types are now also supported, correct?

src/main/java/com/code_intelligence/jazzer/mutation/annotation/ValuePool.java
```diff
@@ -0,0 +1,94 @@
...
+ *   @ValuePool("valuesVisibleByAllArgumentMutators")
+ *   @FuzzTest
+ *   public void fuzzerTestOneInput(String input, @ValuePool("valuesVisibleOnlyByAnotherInput") String anotherInput) {
+ *     // Fuzzing logic here
+ *   }
+ * }
+ * }</pre>
+ *
+ * In this example, the mutator for the String parameter {@code input} of the fuzz test method
+ * {@code fuzzerTestOneInput} will be using the values returned by {@code provide} method during
```
simonresch: Should this be `{@code valluesVisibleByAllArgumentMutators}`?

src/main/java/com/code_intelligence/jazzer/mutation/annotation/ValuePool.java
```diff
@@ -0,0 +1,94 @@
...
+ * mutation, while the mutator for String {@code anotherInput} will use values from both methods:
+ * from the method-level {@code ValuePool} annotation that uses {@code provide} and the
+ * parameter-level {@code ValuePool} annotation that uses {@code provideSomethingElse}.
```
simonresch: And here `valuesVisibleOnlyByAnotherInput`.

src/main/java/com/code_intelligence/jazzer/mutation/annotation/ValuePool.java
```diff
@@ -0,0 +1,94 @@
...
+  /**
+   * Specifies supplier methods that generate values for fuzzing the annotated method or type. The
+   * specified supplier methods must be static and return a {@code Stream <?>} of values. The values
+   * don't need to match the type of the annotated method or parameter exactly. The mutation
+   * framework will extract only the values that are compatible with the target type.
+   */
+  String[] value();
+
+  /**
+   * This {@code ValuePool} will be used with probability {@code 1/p} by the mutator responsible for
```
simonresch: Should this be `{@code p}` instead of `1/p`?

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/ValuePoolMutatorFactory.java
```diff
@@ -0,0 +1,186 @@
...
+      try {
+        mutator.write((T) value, new DataOutputStream(byteStream));
+        byte[] originalSerialized = byteStream.toByteArray();
+        byteStream.reset();
+
+        T roundTrip =
+            mutator.read(new DataInputStream(new ByteArrayInputStream(originalSerialized)));
+        mutator.write(roundTrip, new DataOutputStream(byteStream));
+        byte[] roundTripSerialized = byteStream.toByteArray();
+        return Arrays.equals(originalSerialized, roundTripSerialized);
```
simonresch, line 138: For what mutator types do we need the roundtrip check? Wouldn't it be enough to check if the `mutator.write(...)` call does not throw, to filter out e.g. `CachedConstructorMutator`s or values where the type does not match?

src/main/java/com/code_intelligence/jazzer/mutation/support/ValuePoolRegistry.java
```diff
@@ -0,0 +1,141 @@
...
+   * {@code @ValuePool} annotation directly on the type is preferred; if there is none, the first
+   * one appended because of {@code PropertyConstraint.RECURSIVE} is used. Any further
+   * {@code @ValuePool} annotations appended later to this type because of {@code
+   * PropertyConstraint.RECURSIVE}, are ignored. Callers should ensure that at least one
+   * {@code @ValuePool} annotation is present on the type.
+   */
+  public double extractFirstProbability(AnnotatedType type) {
+    // If there are several @ValuePool annotations on the type, this will take the most
+    // immediate one, because @ValuePool is not repeatable.
+    ValuePool[] dictObj = type.getAnnotationsByType(ValuePool.class);
```
simonresch: Minor: `dictObj` should probably be something like `valuePoolAnnotations`.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/ValuePoolMutatorFactory.java
simonresch: Ah, this will also make sure that the value pool values respect other annotations of the particular mutator, correct? Clever!
Minor: Could be worth a comment that lists what this serialization stable entails.

oetr: Yes, the "take verbatim" part requires that values respect user annotations and default mutator limits. 

Instead of using the values "verbatim" vs. "as a starting point for mutation" with 50/50 split, we could make the split favor mutations higher, or even completely drop the verbatim part.
If we drop it, the round trip will become unnecessary.

We could also add another arg to the `ValuePool` annotation to set the max number of mutations made on the  selected pool value (a little like libFuzzer's mutation chains, but for a single value). Currently, it's just one mutation.

src/main/java/com/code_intelligence/jazzer/mutation/support/TypeSupport.java
oetr: I added your suggestion, since a defensive check through a recursive call won't hurt here, but do you know if "the class or interface that declared this type" can be anything other than `Class<?>`?

The API for `getRawType()` on a `ParameterizedType` is quite specific about this:
> Returns the Type object representing the class or interface that declared this type.

PR #986, opened 2025-10-15
fix: support maxSize > 1000 in map mutator

### LINKED ISSUE #977: `MapMutator` clamps size range values
`MapMutator` clamps the given min and max values:
https://github.com/CodeIntelligenceTesting/jazzer/blob/e13aef033a8183713131a21f0e2128e44b4a3bc9/src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java#L99-L100

This seems to differ from the other mutators such as `ListMutator` or `ArrayMutator` which use the given min and max values as is, and instead only perform validation using `require(...)`.

Is it intended that `MapMutator` differs here in behavior? For consistency, would it make sense to use `require` here as well to impose restrictions (at least for the `minSize`)?

simonresch:
The `@WithSize(max = maxSize)` annotation with `maxSize > 1000` was silently being ignored by the `MapMutator`.

Closes #977

### DIFF
--- a/src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/ArrayMutatorFactory.java
+++ b/src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/ArrayMutatorFactory.java
@@ -86,6 +86,11 @@ final class ArrayMutatorFactory implements MutatorFactory {
       require(
           minLength >= 0,
           format("WithLength#min=%d needs to be greater than or equal to 0", minLength));
+      require(
+          minLength <= maxLength,
+          format(
+              "WithLength#min=%d needs to be smaller or equal than WithLength#max=%d",
+              minLength, maxLength));
     }
 
     @Override


src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/MapMutatorFactory.java
Marcono1234, line 107: `ListMutator` has an additional check for `minSize <= maxSize`:
https://github.com/CodeIntelligenceTesting/jazzer/blob/d57e86957476816626177ef5eaca10d697d96dc6/src/main/java/com/code_intelligence/jazzer/mutation/mutator/collection/ListMutatorFactory.java#L82-L86

Would it make sense to add this here as well (since it is closely related to this change)?

(also applies to `ArrayMutator` but that might be out-of-scope here)

simonresch: Since `minSize` is currently required to be 0 in the `MapMutator` and `maxSize > 1` this should also cover any attempted `minSize > maxSize` usage.

But I added the check to the `ArrayMutator`. :+1:

PR #989, opened 2025-10-17
fix: guard against malformed corpus entries in protobuf mutator

simonresch:
When decoding protobuf messages no size limit was enforced which could lead to OOM failures for corpus entries that were produced by a different mutator / arbitrary binary data.

### DIFF
--- a/src/main/java/com/code_intelligence/jazzer/mutation/mutator/proto/BuilderMutatorFactory.java
+++ b/src/main/java/com/code_intelligence/jazzer/mutation/mutator/proto/BuilderMutatorFactory.java
@@ -57,6 +57,7 @@ import com.code_intelligence.jazzer.mutation.mutator.collection.CollectionMutato
 import com.code_intelligence.jazzer.mutation.mutator.lang.LangMutators;
 import com.code_intelligence.jazzer.mutation.support.Preconditions;
 import com.google.protobuf.Any;
+import com.google.protobuf.CodedInputStream;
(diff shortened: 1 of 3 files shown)


src/main/java/com/code_intelligence/jazzer/mutation/mutator/proto/BuilderMutatorFactory.java
```diff
@@ -86,6 +87,11 @@
 import java.util.stream.Stream;
 
 public final class BuilderMutatorFactory implements MutatorFactory {
+
+  // Generous size limit for decoded protobuf messages. This is necessary to guard against OOM
+  // errors when the corpus format changes e.g. due to a change in the fuzz test signature.
+  private static final int MAX_MESSAGE_SIZE = 32 * 1024 * 1024; // 32 MiB
```
oetr, line 93: I think, we should make this user-configurable in  form of a new annotaiton, and use this default there. WDYT?

simonresch: What would be the use case for changing this value? 32 MB should be more than enough for single encoded corpus entries, no?

kyakdan: This default makes sense to me and it it significantly larger than than the libFuzzer's default size of corpus entries.

kyakdan (approved):

PR #990, opened 2025-10-21
feat: add guidance hook for SpEL injection

simonresch:
Guide invocations of spring expression language towards an attack that invoces the jazzer honeypot class.


examples/src/main/java/com/example/SpringCloudFunctionRoutingFuzzer.java
```diff
@@ -0,0 +1,64 @@
...
+    // Empty function registry
+    SimpleFunctionRegistry registry =
+        new SimpleFunctionRegistry(
+            null,
+            new CompositeMessageConverter(Arrays.asList(new StringMessageConverter())),
+            new JacksonMapper(new ObjectMapper()));
+    router = new RoutingFunction(registry, new FunctionProperties());
+  }
+
+  public static void fuzzerTestOneInput(FuzzedDataProvider data) {
```
kyakdan:
```suggestion
  public static void fuzzerTestOneInput(String payload, String expr) {
```

This would slightly simplify the fuzz test.

PR #991, opened 2025-10-22
fix: fixes for sanitizer handling in honeypot class

simonresch:
A few small fixes to the sanitizer handling in the honeypot class `jaz.Zer`.

With this change invocations of the `Zer.el()` method produce a finding if the `ReflectiveCall` sanitizer is enabled not the `ExpressionLanguageInjection` sanitizer which only is responsible for guiding the fuzzer to invoke that method.

### DIFF
--- a/src/main/java/jaz/Zer.java
+++ b/src/main/java/jaz/Zer.java
@@ -71,7 +71,6 @@ public class Zer
   // serialized size is 41 bytes
   private static final byte REFLECTIVE_CALL_SANITIZER_ID = 0;
   private static final byte DESERIALIZATION_SANITIZER_ID = 1;
-  private static final byte EXPRESSION_LANGUAGE_SANITIZER_ID = 2;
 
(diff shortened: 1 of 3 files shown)


src/main/java/jaz/Zer.java
```diff
@@ -95,7 +104,7 @@ public Zer(byte sanitizer) {
   // choose a parameterless method to keep the string that the sanitizer guides the fuzzer to
   // generate within the 64-byte boundary required by the corresponding guiding methods.
   public static void el() {
-    if (isSanitizerEnabled(EXPRESSION_LANGUAGE_SANITIZER_ID)) {
+    if (isSanitizerEnabled(REFLECTIVE_CALL_SANITIZER_ID)) {
```
kyakdan, line 109: Could you add a comment here describing why we are using `REFLECTIVE_CALL_SANITIZER_ID`? This could avoid confusion in the future.

kyakdan (approved):

PR #994, opened 2025-10-23
feat: add hooks for additional EL evaluation methods

florianGla:
We want to add hooks for additional EL evaluation methods


sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ExpressionLanguageInjection.kt
```diff
@@ -33,6 +33,8 @@ object ExpressionLanguageInjection {
     private const val EXPRESSION_LANGUAGE_ATTACK =
         "\${Byte.class.forName(\"$HONEYPOT_CLASS_NAME\").getMethod(\"el\").invoke(null)}"
     private const val SPRING_EXPRESSION_LANGUAGE_ATTACK = "T($HONEYPOT_CLASS_NAME).el()"
+    private const val ELPROCESSOR_LANGUAGE_ATTACK =
+        "\"\".getClass().forName(\"$HONEYPOT_CLASS_NAME\").getMethod(\"el\").invoke(null)"
```
simonresch, line 37: @kyakdan changed the `EXPRESSION_LANGUAGE_ATTACK` from `\"\"` to `Byte` in https://github.com/CodeIntelligenceTesting/jazzer/commit/acb69630f04591fe85ed6706559787e0cc6af59f. Could this be relevant here as well? I'm not sure what the intention of the change was. @kyakdan maybe you can remember?

If we can use the same injection it might make sense to re-use the `ELPROCESSOR_LANGUAGE_ATTACK` in `EXPRESSION_LANGUAGE_ATTACK`.

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ExpressionLanguageInjection.kt
```diff
@@ -80,17 +85,50 @@ object ExpressionLanguageInjection {
...
+            targetMethod = "eval",
+        ),
+        MethodHook(
+            type = HookType.BEFORE,
+            targetClassName = "jakarta.el.ELProcessor",
+            targetMethod = "eval",
+        ),
+    )
+    @JvmStatic
+    fun hoolElProcessor(
```
simonresch:
```suggestion
    fun hookElProcessor(
```

MODULE.bazel
```diff
@@ -104,6 +104,7 @@ VULNERABLE_TEST_MAVEN_ARTIFACTS = [
     "com.h2database:h2:2.1.212",
     "com.mikesamuel:json-sanitizer:1.2.1",
     "com.unboundid:unboundid-ldapsdk:6.0.3",
+    "jakarta.validation:jakarta.validation-api:3.0.2",
```
simonresch: Do we need a specific version of this dependency? If I see correctly we only need it for the `HookBindingSanityTest` which means we could keep it in the list of test dependencies which are updated by renovate (dependency list above).

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ExpressionLanguageInjection.kt
kyakdan: @simonresch @florianGla Many EL engines allow `Type.class` but may restrict or differently resolve `"" .getClass()`. In our case, using `ELPROCESSOR_LANGUAGE_ATTACK` to guide fuzzing would cause the `//sanitizers/src/test/java/com/example:ExpressionLanguageInjection_fuzzValidator` test to fail to trigger the bug.

florianGla: Thank you! Maybe I can find a better name for `ELPROCESSOR_LANGUAGE_ATTACK` and reuse it for other EL engines.

simonresch: For me the test runs fine with `private const val EXPRESSION_LANGUAGE_ATTACK = "\${${ELPROCESSOR_LANGUAGE_ATTACK}}"`.
@kyakdan Do you know of specific EL engines that restrict `.getClass()` but allow `Type.class`?

kyakdan: I have to dig deeper, but I remember having issues triggering the bug with the old attack pattern, which is why we adjusted it in https://github.com/CodeIntelligenceTesting/jazzer/commit/acb69630f04591fe85ed6706559787e0cc6af59f

PR #995, opened 2025-10-23
fix: fuzz test invocation with referring protobufs

simonresch:
When constructing a mutator for a protobuf message the mutator is cached based on the protobuf descriptor. This led to issues for fuzz tests that take two protobuf messages where one message is referring to the other.

While building the mutator for the top level message nested messages are constructed with the type `DynamicMessage` which lead to 'argument type mismatch' exceptions when invoking the fuzz test method. This is fixed by including the message class in the mutator cache key which ensures that a generated protobuf message class does not re-use the corresponding `DynamicMessage` mutator.

Note, that this causes some recursion breaking to happen one layer deeper (see test assertion changes).

### DIFF
--- a/src/test/java/com/code_intelligence/jazzer/mutation/ArgumentsMutatorTest.java
+++ b/src/test/java/com/code_intelligence/jazzer/mutation/ArgumentsMutatorTest.java
@@ -23,6 +23,7 @@ import static java.util.Collections.singletonList;
 import com.code_intelligence.jazzer.mutation.annotation.NotNull;
 import com.code_intelligence.jazzer.mutation.mutator.Mutators;
 import com.code_intelligence.jazzer.mutation.support.TestSupport.MockPseudoRandom;
+import com.code_intelligence.jazzer.protobuf.Proto2;
 import com.code_intelligence.jazzer.protobuf.Proto2.TestProtobuf;
 import com.google.protobuf.DescriptorProtos;
 import java.io.ByteArrayInputStream;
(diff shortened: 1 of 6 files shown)


src/test/java/com/code_intelligence/jazzer/mutation/ArgumentsMutatorTest.java
```diff
@@ -371,4 +372,24 @@ void testProtoAndDescriptorParameters() throws NoSuchMethodException {
     //   because "productMutator" is null
     assertThat(maybeMutator.toString()).contains("-> Message");
   }
+
+  // Regression: ensure that the mutator can invoke a method with two protobuf messages where one
+  // message is a submessage of the other. This would fail with an "argument type mismatch" error
+  // when using the type `DynamicMessage` instead of `Proto2.TestSubProtobuf`.
```
oetr: Without more context, "when using the type `DynamicMessage` instead of `Proto2.TestSubProtobuf`" is super confusing. Before I went down this rabbit hole, I thought this means: if I replaced the type for the second argument `@NotNull Proto2.TestSubProtobuf` by `@NotNull DynamicMessage`.
Now I understand that there is a difference between trying to construct a mutator for top-level argument proto type, vs. constructing mutator for a proto field of some type, where the `messageClass` will be `DynamicMessge`.

simonresch: Your right, this comment is impossible to decipher. :upside_down_face:  I changed the wording to hopefully be a bit more descriptive, but I'm having a hard time explaining the bug in a concise but understandable way.

oetr: Nothing is concise or understandable about protobufs!

PR #996, opened 2025-10-24
chore: exclude potentially dangerous tests by default

### LINKED ISSUE #971: Make Jazzer tests safer / make dangerous tests opt-in
See https://github.com/CodeIntelligenceTesting/jazzer/pull/932#issuecomment-3156083058

It seems some of Jazzer's own tests, especially related to the sanitizers, are inherently unsafe and could in the worst case damage the OS they are running on.

For example:
- `OsCommandInjectionProcessBuilder`, `OsCommandInjectionRuntimeExec`: run arbitrary OS commands
- `ExpressionLanguageInjection`, `LdapDnInjection`, `LdapSearchInjection`, `ObjectInputStreamDeserialization`, `XPathInjection`: run arbitrary code in the worst case?
- `SsrfHttpClient`, `SsrfSocketConnect`, `SsrfSocketConnectToHost` (?), `SsrfUrlConnection`: send network requests to arbitrary hosts

I am not completely sure for all of them, and maybe there are more problematic tests.

If possible it would be good to edit or extend these tests so that they don't perform any potentionally dangerous actions (e.g. by using dummy implementations, similar to how it is done for the `ScriptEngineInjection` test, or using no-op method hooks as suggested in https://github.com/CodeIntelligenceTesting/jazzer/pull/932#issuecomment-3156083058).
Or if that cannot be avoided, make the tests opt-in (with an explicit command line option, e.g. `--enable-dangerous-tests`) and then only run them on CI?

What do you think?

simonresch:
All tests that can perform potentially harmful side effects are tagged and excluded by default. The CI still tests all.

Closes #971

### DIFF
--- a/sanitizers/src/test/java/com/example/BUILD.bazel
+++ b/sanitizers/src/test/java/com/example/BUILD.bazel
@@ -10,6 +10,7 @@ java_fuzz_target_test(
         "com.code_intelligence.jazzer.api.FuzzerSecurityIssueHigh",
         "java.lang.ExceptionInInitializerError",
     ],
+    tags = ["dangerous"],
     target_class = "com.example.ObjectInputStreamDeserialization",
 )
 
@@ -22,6 +23,7 @@ java_fuzz_target_test(
         "com.code_intelligence.jazzer.api.FuzzerSecurityIssueHigh",
         "java.lang.ExceptionInInitializerError",
     ],
+    tags = ["dangerous"],
     target_class = "com.example.ReflectiveCall",
 )
 
@@ -33,6 +35,7 @@ java_fuzz_target_test(
     allowed_findings = [
         "com.code_intelligence.jazzer.api.FuzzerSecurityIssueHigh",
     ],
+    tags = ["dangerous"],
     target_class = "com.example.LibraryLoad",
     # loading of native libraries is very slow on macos,
     # especially using Java 17
@@ -49,6 +52,7 @@ java_fuzz_target_test(
     ],
     allowed_findings = ["com.code_intelligence.jazzer.api.FuzzerSecurityIssueHigh"],
     expected_warning_or_error = "WARN: Some hooks could not be applied to class files built for Java 7 or lower.",
+    tags = ["dangerous"],
     target_class = "com.example.ExpressionLanguageInjection",
     # The reproducer can't find jaz.Zer and thus doesn't crash.
     verify_crash_reproducer = False,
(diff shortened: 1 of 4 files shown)


sanitizers/src/test/java/com/example/BUILD.bazel
```diff
@@ -489,6 +495,7 @@ java_fuzz_target_test(
     allowed_findings = [
         "com.code_intelligence.jazzer.api.FuzzerSecurityIssueMedium",
     ],
+    tags = ["dangerous"],
     target_class = "com.example.SsrfSocketConnect",
```
kyakdan, line 500: Shouldn't we tag other tests such as `ClassLoaderLoadClass` as dangerous?

simonresch: IMO the risk of calling the default constructor of classes on the class path should be minimal. Or am I missing something? What other tests would you consider potentially harmful that are not marked in this PR?

kyakdan: In principle, a Java class can also have static blocks that are executed during class loading, which can be "potentially" dangerous. For example, [jackson-databind](https://github.com/FasterXML/jackson-databind/blob/67f8ef7b1d10a1625c6405b9dadfd1cbcc2ae692/src/main/java/tools/jackson/databind/jsontype/impl/SubTypeValidator.java#L26) has a deny list of "nasty classes" that are not deserialized by default. IMHO tagging a fuzz test that can execute arbitrary code as a "dangerous" would be the better approach.

simonresch: Executing arbitrary code is not quite the same as constructing default instances for classes on the class path, no?
Out of interest I checked if we have some of the jackson known classes in our class path. These were available on the `ClassLoaderLoadClass` test:

```
java.util.logging.FileHandler
javax.swing.JEditorPane
javax.swing.JTextPane
com.sun.rowset.JdbcRowSetImpl
java.rmi.server.UnicastRemoteObject
com.sun.org.apache.xalan.internal.xsltc.trax.TemplatesImpl
```

I'll mark it and similar tests as `dangerous` just to be on the safe side.

kyakdan: > Executing arbitrary code is not quite the same as constructing default instances for classes on the class path, no?
I'm not arguing that. All I'm saying is executing code that we cannot control can be potentially dangerous or have unexpected side effects.

PR #999, opened 2025-10-27
feat: add xml parser guidance hook

simonresch:
A simple guidance hook to nudge Jazzer towards inserting an document xml tag with an external system identifier. If an xml parser is called with fuzzer input and configured to load external entities this will trigger a finding from the SSRF bug detector.

### DIFF
--- a/sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/Utils.kt
+++ b/sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/Utils.kt
@@ -16,8 +16,8 @@
 
 package com.code_intelligence.jazzer.sanitizers
 
-import com.code_intelligence.jazzer.api.Jazzer
 import java.io.InputStream
+import java.io.Reader
 
 /**
  * jaz.Zer is a honeypot class: All of its methods report a finding when called.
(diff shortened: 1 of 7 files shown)


sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/Utils.kt
```diff
@@ -60,12 +59,35 @@ internal fun guideMarkableInputStreamTowardsEquality(
             if (count < 0) break
             n += count
         }
-        return current
+        return if (n >= readlimit) current else current.copyOf(n)
```
Marcono1234, line 62: Really minor: Shouldn't this use `size` instead of `readlimit` for consistency, since `size` is the parameter of this local function (and it just happens to be called with `readlimit` as value)?

Also, would it make sense to switch the branches? Due to Copilot's suggestion to use `>=`, this statement looks a bit confusing now. What about this?
```suggestion
        return if (n < size) current.copyOf(n) else current
```

(Sorry if this comment is distruptive; please let me know in that case)

simonresch: Not disruptive at all. I completely agree. :)
I created a small PR: #1011

PR #1003, opened 2025-10-29
feat: add jexl expression guidance hook

florianGla:
A simple guidance hook to nudge Jazzer towards triggering a remote code execution via JEXL expression language injection.


sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ExpressionLanguageInjection.kt
```diff
@@ -171,4 +171,23 @@ object ExpressionLanguageInjection {
         val expr = arguments[0] as? String ?: return
         Jazzer.guideTowardsContainment(expr, SPRING_EXPRESSION_LANGUAGE_ATTACK, hookId)
     }
+
+    @MethodHook(
+        type = HookType.BEFORE,
+        targetClassName = "org.apache.commons.jexl2.JexlEngine",
+        targetMethod = "createExpression",
```
simonresch: Would the [createScript](https://commons.apache.org/proper/commons-jexl/javadocs/apidocs-2.1/org/apache/commons/jexl2/JexlEngine.html#createScript(java.lang.String)) functions also be interesting to hook?

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ExpressionLanguageInjection.kt
```diff
@@ -171,4 +171,23 @@ object ExpressionLanguageInjection {
...
+    )
+    @JvmStatic
+    fun hookJexlExpression(
+        method: MethodHandle?,
+        thisObject: Any?,
+        arguments: Array<Any>,
+        hookId: Int,
+    ) {
+        if (arguments.size != 1) {
```
simonresch: This check rejects the [two parameter overload](https://commons.apache.org/proper/commons-jexl/javadocs/apidocs-2.1/org/apache/commons/jexl2/JexlEngine.html#createExpression(java.lang.String,%20org.apache.commons.jexl2.JexlInfo)) of `createExpression` which could be interesting to hook.

sanitizers/src/test/java/com/example/ExpressionLanguageInjection.java
```diff
@@ -25,6 +25,7 @@
 import javax.el.ELProcessor;
 import javax.validation.Validation;
 import javax.validation.Validator;
+import org.apache.commons.jexl2.*;
```
simonresch: Minor: We are trying to not use wildcard imports. There is an IntelliJ setting to make it stop combining the imports.

PR #1004, opened 2025-10-30
feat: add ognl expression guidance hook

florianGla:
A simple guidance hook to nudge Jazzer towards triggering a remote code execution via OGNL expression language injection.


sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/ExpressionLanguageInjection.kt
```diff
@@ -171,4 +175,23 @@ object ExpressionLanguageInjection {
         val expr = arguments[0] as? String ?: return
         Jazzer.guideTowardsContainment(expr, SPRING_EXPRESSION_LANGUAGE_ATTACK, hookId)
     }
+
+    @MethodHook(
+        type = HookType.BEFORE,
+        targetClassName = "ognl.Ognl",
+        targetMethod = "parseExpression",
```
simonresch: WDYT about adding other Ognl methods that take a `String expression`? E.g. [compileExpression](https://javadoc.io/doc/ognl/ognl/latest/index.html) could be interesting and the other convenience methods around `parseExpression`.

simonresch (approved):
LGTM

PR #1006, opened 2025-10-30
feat: add support for proto maps with enum values

simonresch:
Adds mutator support for protobuf maps where the map value is a protobuf enum.

```
message TestProtobuf {
  enum Enum {
    Label1 = 0;
    Label2 = 1;
    Label3 = 2;
    Label4 = 3;
    Label5 = 4;
  }

  map<int32, Enum> map_field = 1;
}
```


oetr (approved):
LGTM! Nice use of recursion!
You can set the timeout="long" for the failing test. It is very sensitive to PRNG-related changes. In this case you added new fields to proto2 and changed the mutations that are always running with the same seed. You can also find a better seed, if you have time!

PR #1007, opened 2025-11-03
feat: add freemarker template injection guidance hook

florianGla:
A simple guidance hook to nudge Jazzer towards creating FreeMarker
templates with a pattern that can trigger OS command injections.


sanitizers/src/test/java/com/example/BUILD.bazel
```diff
@@ -314,6 +314,24 @@ java_fuzz_target_test(
     ],
 )
 
+java_fuzz_target_test(
```
Marcono1234, line 317: Should this have `tags = ["dangerous"]` in case it can trigger OS command execution?

florianGla: Thank you, you are right!

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/TemplateInjection.kt
```diff
@@ -0,0 +1,61 @@
...
+ * This does not report findings directly; it steers inputs towards OS command injections,
+ * which are detected by another bug detector.
+ */
+@Suppress("unused_parameter", "unused")
+object TemplateInjection {
+    private const val FREEMARKER_INJECTION_ATTACK: String = "\${\"freemarker.template.utility.Execute\"?new()(\"jazze\")}"
+
+    init {
+        require(FREEMARKER_INJECTION_ATTACK.length <= 64) {
+            "Expression language exploit must fit in a table of recent compares entry (64 bytes)"
```
simonresch:
```suggestion
            "FreeMarker injection must fit in a table of recent compares entry (64 bytes)"
```

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/TemplateInjection.kt
```diff
@@ -0,0 +1,61 @@
...
+    )
+    @JvmStatic
+    fun hookFreemarker(
+        method: MethodHandle?,
+        thisObject: Any?,
+        arguments: Array<Any>,
+        hookId: Int,
+    ) {
+        if (arguments.size < 2) return
+        val reader = arguments[1] as? Reader ?: return
```
simonresch: There are also constructors with `reader` as a [third parameter](https://javadoc.io/static/org.freemarker/freemarker/2.3.34/freemarker/template/Template.html#%3Cinit%3E(java.lang.String,java.lang.String,java.io.Reader,freemarker.template.Configuration)) and a [constructor that takes the template as a String](https://javadoc.io/static/org.freemarker/freemarker/2.3.34/freemarker/template/Template.html#%3Cinit%3E(java.lang.String,java.lang.String,freemarker.template.Configuration)).

sanitizers/src/test/java/com/example/BUILD.bazel
```diff
@@ -314,6 +314,25 @@ java_fuzz_target_test(
...
+    name = "TemplateInjection",
+    srcs = [
+        "TemplateInjection.java",
+    ],
+    allowed_findings = [
+        "com.code_intelligence.jazzer.api.FuzzerSecurityIssueCritical",
+    ],
+    fuzzer_args = [
+        "--instrumentation_includes=com.example.**",
```
simonresch, line 327: A comment here could be useful explaining why we have to modify the default instrumentation filters.

sanitizers/src/main/java/com/code_intelligence/jazzer/sanitizers/TemplateInjection.kt
florianGla: The constructor that takes the template as a String calls a constructor that takes a Reader.

simonresch: You might need to add `freemarker.template.Template` to the additional classes to instrument to make sure that this constructor call will be instrumented.

simonresch (approved):

PR #1008, opened 2025-11-05
feat: add mutator support for generic classes

simonresch:
Adds mutator support for generic classes.

Constructing mutators for generic classes would previously either fail or not work as expected due to parameter and return types being reported as `java.lang.Object` by the reflection API (type erasure). Since we can't provide a general `Object` mutator this would creation of a valid mutator for most generic classes.
Using the `AnnotatedType` of the class we resolve type parameters to create the correct underlying mutators.

Example fuzz test:
```
  public static class MyClass<L, S> {
    private final L[] list;
    private final Set<S> set;

    public MyClass(L[] list, Set<S> set) {
      this.list = list;
      this.set = set;
    }

    public L[] getList() {
      return this.list;
    }

    public Set<S> getSet() {
      return this.set;
    }
  }

  @FuzzTest
  void fuzzTest(@NotNull(constraint = PropertyConstraint.RECURSIVE) MyClass<String, Integer> data) {
    if (data.getList().length > 5
        && data.getList()[5].equals("foo")
        && data.getSet().size() == 5) {
      throw new RuntimeException("Can fuzz generics");
    }
  }
```

### DIFF
--- a/src/main/java/com/code_intelligence/jazzer/mutation/mutator/aggregate/BeanSupport.java
+++ b/src/main/java/com/code_intelligence/jazzer/mutation/mutator/aggregate/BeanSupport.java
@@ -32,6 +32,7 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.Executable;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
+import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.Arrays;
 import java.util.Comparator;
@@ -52,17 +53,27 @@ class BeanSupport {
     }
   }
 
+  private static Class<?> rawBeanType(Type classType) {
+    if (classType instanceof Class<?>) {
+      return (Class<?>) classType;
+    } else if (classType instanceof ParameterizedType) {
+      return rawBeanType(((ParameterizedType) classType).getRawType());
+    } else {
+      // Bail out on wildcard types or type variables neither of which are supported as "top level"
+      // types in a @FuzzTest. Also bail out on generic array types as they are handled by the
+      // ArrayMutatorFactory.
+      throw new UnsupportedOperationException("Unsupported type: " + classType);
+    }
+  }
+
   // Returns the annotated parameter types of a method or constructor resolving all generic type
   // arguments.
   public static AnnotatedType[] resolveAnnotatedParameterTypes(
       Executable e, AnnotatedType classType) {
-    Type[] generic = e.getGenericParameterTypes();
-    AnnotatedType[] annotated = e.getAnnotatedParameterTypes();
-    AnnotatedType[] result = new AnnotatedType[generic.length];
-    for (int i = 0; i < generic.length; i++) {
-      result[i] = resolveTypeArguments(e.getDeclaringClass(), classType, annotated[i]);
-    }
-    return result;
+    Class<?> clazz = rawBeanType(classType.getType());
+    return stream(e.getAnnotatedParameterTypes())
+        .map(t -> resolveTypeArguments(clazz, classType, t))
+        .toArray(AnnotatedType[]::new);
   }
 
   // Returns the parameter types of a method or constructor resolving all generic type arguments.
@@ -74,7 +85,7 @@ class BeanSupport {
 
   static Type resolveReturnType(Method method, AnnotatedType classType) {
     return resolveTypeArguments(
-            method.getDeclaringClass(), classType, method.getAnnotatedReturnType())
+            rawBeanType(classType.getType()), classType, method.getAnnotatedReturnType())
         .getType();
   }
 
--- a/src/main/java/com/code_intelligence/jazzer/mutation/support/ParameterizedTypeSupport.java
+++ b/src/main/java/com/code_intelligence/jazzer/mutation/support/ParameterizedTypeSupport.java
@@ -64,21 +64,50 @@ public class ParameterizedTypeSupport {
    */
   public static AnnotatedType resolveTypeArguments(
       Class<?> clazz, AnnotatedType classType, AnnotatedType type) {
-    if (!(classType instanceof AnnotatedParameterizedType)) {
+    Map<TypeVariable<?>, AnnotatedType> mapping = new HashMap<>();
+    updateTypeMappings(clazz, classType, mapping);
+    if (mapping.isEmpty()) {
       return type;
     }
-
-    TypeVariable<?>[] typeParameters = clazz.getTypeParameters();
-    AnnotatedType[] typeArguments =
-        ((AnnotatedParameterizedType) classType).getAnnotatedActualTypeArguments();
-
-    require(typeArguments.length == typeParameters.length);
-
-    Map<TypeVariable<?>, AnnotatedType> mapping = new HashMap<>();
-    for (int i = 0; i < typeParameters.length; i++) {
-      mapping.put(typeParameters[i], typeArguments[i]);
-    }
     return resolveRecursive(type.getType(), type, mapping);
+  }
+
+  private static void updateTypeMappings(
+      Class<?> clazz,
+      AnnotatedType annotatedClazzType,
+      Map<TypeVariable<?>, AnnotatedType> mapping) {
+    if (annotatedClazzType instanceof AnnotatedParameterizedType) {
+      TypeVariable<?>[] typeParameters = clazz.getTypeParameters();
+      AnnotatedType[] typeArguments =
+          ((AnnotatedParameterizedType) annotatedClazzType).getAnnotatedActualTypeArguments();
+      require(typeArguments.length == typeParameters.length);
+      for (int i = 0; i < typeParameters.length; i++) {
+        mapping.put(typeParameters[i], typeArguments[i]);
+      }
+    }
+
+    Class<?> superClass = clazz.getSuperclass();
+    AnnotatedType annotatedSuperclass = clazz.getAnnotatedSuperclass();
+    Type genericSuperclass = clazz.getGenericSuperclass();
+    if (superClass != null && annotatedSuperclass != null && genericSuperclass != null) {
+      AnnotatedType resolvedSuperclass =
+          resolveRecursive(genericSuperclass, annotatedSuperclass, mapping);
+      updateTypeMappings(superClass, resolvedSuperclass, mapping);
+    }
+
+    Class<?>[] interfaces = clazz.getInterfaces();
+    AnnotatedType[] annotatedInterfaces = clazz.getAnnotatedInterfaces();
+    Type[] genericInterfaces = clazz.getGenericInterfaces();
+    for (int i = 0; i < interfaces.length; i++) {
+      AnnotatedType annotatedInterface = annotatedInterfaces[i];
+      Type genericInterface = genericInterfaces[i];
+      if (annotatedInterface == null || genericInterface == null) {
+        continue;
+      }
+      AnnotatedType resolvedInterface =
+          resolveRecursive(genericInterface, annotatedInterface, mapping);
+      updateTypeMappings(interfaces[i], resolvedInterface, mapping);
+    }
   }
 
   /**
@@ -106,6 +135,10 @@ public class ParameterizedTypeSupport {
       AnnotatedType replacement = mapping.get(type);
       if (replacement == null) {
         return annotated;
+      }
+      if (replacement instanceof AnnotatedWildcardType) {
+        // Forwarding annotations to wildcard types is not supported
+        return replacement;
       }
       return TypeSupport.forwardAnnotations(annotated, replacement);
     }
--- a/src/test/java/com/code_intelligence/jazzer/mutation/mutator/aggregate/SetterBasedBeanMutatorTest.java
+++ b/src/test/java/com/code_intelligence/jazzer/mutation/mutator/aggregate/SetterBasedBeanMutatorTest.java
@@ -246,4 +246,14 @@ class SetterBasedBeanMutatorTest {
                 .createOrThrow(new TypeHolder<Generic<String>>() {}.annotatedType());
     assertThat(mutator.toString()).isEqualTo("Nullable<[Nullable<Nullable<String>[]>] -> Generic>");
   }
+
+  public static class Child extends Generic<String> {}
+
+  @Test
+  void genericClassChild() {
+    SerializingMutator<Child> mutator =
+        (SerializingMutator<Child>)
+            Mutators.newFactory().createOrThrow(new TypeHolder<Child>() {}.annotatedType());
+    assertThat(mutator.toString()).isEqualTo("Nullable<[Nullable<Nullable<String>[]>] -> Child>");
+  }
 }
--- a/src/test/java/com/code_intelligence/jazzer/mutation/support/ParameterizedTypeSupportTest.java
+++ b/src/test/java/com/code_intelligence/jazzer/mutation/support/ParameterizedTypeSupportTest.java
@@ -125,4 +125,89 @@ class ParameterizedTypeSupportTest {
     assertThat(parameterizedType.getActualTypeArguments()[0]).isEqualTo(String.class);
     assertThat(parameterizedType.getActualTypeArguments()[1]).isEqualTo(Integer.class);
   }
+
+  @Test
+  void resolveParameterizedTypeChildClass() throws NoSuchFieldException {
+    class Base<T, U> {
+      public Map<T, U> field;
+    }
+    class Child<U> extends Base<String, U> {}
+    AnnotatedType annotatedType = Child.class.getField("field").getAnnotatedType();
+    AnnotatedParameterizedType classType =
+        (AnnotatedParameterizedType) new TypeHolder<Child<Integer>>() {}.annotatedType();
+    AnnotatedType resolved =
+        ParameterizedTypeSupport.resolveTypeArguments(Child.class, classType, annotatedType);
+
+    assertThat(resolved).isInstanceOf(AnnotatedParameterizedType.class);
+
+    AnnotatedParameterizedType parameterType = (AnnotatedParameterizedType) resolved;
+    assertThat(((ParameterizedType) parameterType.getType()).getRawType()).isEqualTo(Map.class);
+    AnnotatedType[] elementTypes = parameterType.getAnnotatedActualTypeArguments();
+    assertThat(elementTypes).hasLength(2);
+    assertThat(elementTypes[0].getType()).isEqualTo(String.class);
+    assertThat(
+            TypeSupport.annotatedTypeEquals(
+                classType.getAnnotatedActualTypeArguments()[0], elementTypes[1]))
+        .isTrue();
+  }
+
+  @Test
+  void resolveParameterizedType_multiLevelHierarchy() throws NoSuchFieldException {
+    class Root<T> {
+      public List<T> field;
+    }
+    class Middle<U> extends Root<List<U>> {}
+    class Leaf<V> extends Middle<V> {}
+    class Concrete extends Leaf<String> {}
+
+    AnnotatedType annotatedType = Concrete.class.getField("field").getAnnotatedType();
+    AnnotatedType classType = new TypeHolder<Concrete>() {}.annotatedType();
+    AnnotatedType resolved =
+        ParameterizedTypeSupport.resolveTypeArguments(Concrete.class, classType, annotatedType);
+
+    assertThat(resolved).isInstanceOf(AnnotatedParameterizedType.class);
+
+    AnnotatedParameterizedType outerList = (AnnotatedParameterizedType) resolved;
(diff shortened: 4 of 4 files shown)


src/test/java/com/code_intelligence/jazzer/mutation/mutator/aggregate/CachedConstructorMutatorTest.java
```diff
@@ -192,4 +192,34 @@ void testEmptyArgsConstructor() throws IOException {
...
+
+  static class GenericClass<T> {
+    private final T t;
+
+    GenericClass(T t) {
+      this.t = t;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
oetr: We don't need `equals` and `hashCode` methods if the object instances are never compared to each other.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/aggregate/BeanSupport.java
```diff
@@ -49,6 +53,54 @@ static Optional<Class<?>> optionalClassForName(String targetClassName) {
     }
   }
 
+  // Returns the resolved type argument for a generic class if one exists.
+  // For example: For the class `class MyClass<T> {}` with annotated type `MyClass<String>`,
+  // calling `resolveTypeArgument(MyClass.class, annotatedType, "T")` returns
+  // `Optional.of(String.class)`.
+  private static Optional<AnnotatedType> resolveTypeArgument(
```
oetr: You missed an opportunity to use a stream, and to make an early exit!
```java
  private static Optional<AnnotatedType> resolveTypeArgument(
      Class<?> clazz, AnnotatedType classType, String typeName) {
    if (!(classType instanceof AnnotatedParameterizedType)) {
      return Optional.empty();
    }

    TypeVariable<?>[] typeParameters = clazz.getTypeParameters();
    AnnotatedType[] typeArguments =
        ((AnnotatedParameterizedType) classType).getAnnotatedActualTypeArguments();

    return IntStream.range(0, typeParameters.length)
        .filter(i -> typeParameters[i].getName().equals(typeName))
        .mapToObj(i -> typeArguments[i])
        .findFirst();
  }
```

src/main/java/com/code_intelligence/jazzer/mutation/mutator/aggregate/BeanSupport.java
```diff
@@ -49,6 +53,54 @@ static Optional<Class<?>> optionalClassForName(String targetClassName) {
...
+      Class<?> clazz, AnnotatedType classType, String typeName) {
```
oetr: String matching on types might be problematic in several non-intuitive ways, so it might make more sense to use the type obect equality. 
E.g.:

```java
class Box<T> {
  // Method-level T shadows class-level T
  public <T> T id(T x) { return x; }
}
```

I am not sure what happens when the code contains `java.utils.List` instead of `List`.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/aggregate/BeanSupport.java
simonresch: To bad we don't get a `zip` function for streams. The Guava `Streams.zip` is marked with `@Beta`. :shrug:

src/main/java/com/code_intelligence/jazzer/mutation/mutator/aggregate/BeanSupport.java
simonresch: Very good point! Something gave me the impression that the types can't be directly compared but it seems to work perfectly fine and likely handles the special cases you mentioned.

src/main/java/com/code_intelligence/jazzer/mutation/support/ParameterizedTypeSupport.java
```diff
@@ -0,0 +1,492 @@
...
+
+    public ParameterizedTypeWrapper(Type rawType, Type[] typeArguments, Type ownerType) {
+      this.rawType = rawType;
+      this.typeArguments = typeArguments;
+      this.ownerType = ownerType;
+    }
+
+    @Override
+    public Type[] getActualTypeArguments() {
+      return typeArguments;
```
oetr:
```suggestion
      return typeArguments.clone();
```

src/main/java/com/code_intelligence/jazzer/mutation/support/ParameterizedTypeSupport.java
```diff
@@ -0,0 +1,492 @@
...
+        AnnotatedType[] lowerBounds,
+        AnnotatedType[] upperBounds) {
+      super(annotatedType, type);
+      this.upperBounds = upperBounds;
+      this.lowerBounds = lowerBounds;
+    }
+
+    @Override
+    public AnnotatedType[] getAnnotatedLowerBounds() {
+      return lowerBounds;
```
oetr:
```suggestion
      return lowerBounds.clone();
```

src/main/java/com/code_intelligence/jazzer/mutation/support/ParameterizedTypeSupport.java
```diff
@@ -0,0 +1,492 @@
...
+    }
+
+    @Override
+    public AnnotatedType[] getAnnotatedUpperBounds() {
+      return upperBounds;
```
oetr:
```suggestion
      return upperBounds.clone();
```

src/main/java/com/code_intelligence/jazzer/mutation/support/ParameterizedTypeSupport.java
```diff
@@ -0,0 +1,492 @@
...
+
+    AnnotatedParameterizedTypeWrapper(
+        AnnotatedType annotatedType, Type type, AnnotatedType[] typeArguments) {
+      super(annotatedType, type);
+      this.typeArguments = typeArguments;
+    }
+
+    @Override
+    public AnnotatedType[] getAnnotatedActualTypeArguments() {
+      return typeArguments;
```
oetr:
```suggestion
      return typeArguments.clone();
```

src/main/java/com/code_intelligence/jazzer/mutation/support/ParameterizedTypeSupport.java
```diff
@@ -0,0 +1,492 @@
...
+  }
+
+  private static class ParameterizedTypeWrapper implements ParameterizedType {
+    private final Type rawType;
+    private final Type[] typeArguments;
+    private final Type ownerType;
+
+    public ParameterizedTypeWrapper(Type rawType, Type[] typeArguments, Type ownerType) {
+      this.rawType = rawType;
+      this.typeArguments = typeArguments;
```
oetr: Just making sure that callers cannot manipulate the internal state. This applies to the constructors like here, and also the array getters returning `AnnotatedType[]`. You did this for `WildcardTypeWrapper` but not for the others. I marked some more places below with missing `.clone()`
```suggestion
      this.typeArguments = typeArguments.clone();
```

src/main/java/com/code_intelligence/jazzer/mutation/support/ParameterizedTypeSupport.java
```diff
@@ -0,0 +1,492 @@
...
+    private final AnnotatedType[] lowerBounds;
+
+    AnnotatedWildcardTypeWrapper(
+        AnnotatedType annotatedType,
+        Type type,
+        AnnotatedType[] lowerBounds,
+        AnnotatedType[] upperBounds) {
+      super(annotatedType, type);
+      this.upperBounds = upperBounds;
+      this.lowerBounds = lowerBounds;
```
oetr:
```suggestion
      this.upperBounds = upperBounds.clone();
      this.lowerBounds = lowerBounds.clone();
```

src/main/java/com/code_intelligence/jazzer/mutation/support/ParameterizedTypeSupport.java
```diff
@@ -0,0 +1,492 @@
...
+  }
+
+  private static class AnnotatedParameterizedTypeWrapper extends AnnotatedTypeWrapper
+      implements AnnotatedParameterizedType {
+    private final AnnotatedType[] typeArguments;
+
+    AnnotatedParameterizedTypeWrapper(
+        AnnotatedType annotatedType, Type type, AnnotatedType[] typeArguments) {
+      super(annotatedType, type);
+      this.typeArguments = typeArguments;
```
oetr:
```suggestion
      this.typeArguments = typeArguments.clone();
```

src/main/java/com/code_intelligence/jazzer/mutation/support/ParameterizedTypeSupport.java
```diff
@@ -0,0 +1,492 @@
...
+   *
+   * <p>For example, given {@code class Box<T> { List<T> values; }} and the annotated type {@code
+   * Box<@NotNull String>}, calling this method with the {@code values} field's annotated type
+   * returns a new {@link AnnotatedType} representing {@code List<@NotNull String>}.
+   *
+   * @param clazz the generic class that declares the type variables
+   * @param classType the annotated instantiation of {@code clazz}
+   * @param type the annotated type to resolve (e.g. a constructor parameter or getter return type)
+   */
+  public static AnnotatedType resolveTypeArguments(
```
oetr, line 65: I tested various generic Classes with the help of our `ArgumentsMutatorFuzzTest`.

To kick things off, this non-generic version works and gives us correct mutator: `Arguments[[String] -> Child]`
```java
public static class Base {
    private String x;

    public void setX(String x) { this.x = x; }
    public String getX() { return x; }
  }
  public static class Child extends Base {}

@Solo
@SelfFuzzTest
public static void fuzzGenericClass(@NotNull(constraint = PropertyConstraint.RECURSIVE) Child pair) {}
```

The same but generic version with fixed generic type of the superclass Base gives us the trivial mutator that always returns null: `Arguments[[] -> Child]`
Expected: `Arguments[[String]->Child]`
```java
public static class Base<T> {
    private T x;

    public void setX(T x) { this.x = x; }
    public T getX() { return x; }
  }
  public static class Child extends Base<String> {
  }

@Solo
@SelfFuzzTest
public static void fuzzGenericClass(@NotNull(constraint = PropertyConstraint.RECURSIVE) Child pair) {}
```


This gives us the wrong mutator: `Arguments[[Integer] -> Child]`
Expected: `Arguments[[String] -> Child]`
```java
public static class Base<X> {
     protected X x;

     public void setX(X x) { this.x = x; }
     public X getX() { return x; }
   }
   public static class Child<U> extends Base<String> { }

 @Solo
 @SelfFuzzTest
 public static void fuzzGenericClass(@NotNull Child<Integer> pair) {}
```

A variation of above with the same problem--- it gives us the wrong mutator: `Arguments[[String, String] -> Child]`
Expected: `Arguments[[Integer,String]->Child]` (or first String and then Integer, depending on which class we start with first):
```java
public static class Base<X> {
     protected X x;

     public void setX(X x) { this.x = x; }
     public X getX() { return x; }
   }
   public static class Child<U> extends Base<Integer> {
     protected U u;
     public U getU() { return u; }
     public void setU(U u) { this.u = u; }
   }

   @Solo
   @SelfFuzzTest
   public static void fuzzGenericClass(@NotNull Child<@NotNull String> pair) {
      if (pair == null) {
        throw new IllegalArgumentException("pair is null");
      }
      // child gets resolved
      String u = pair.getU();
      // base does not get resolved

      Object x = pair.getX();
      if (x == null) {
        return;
      }
      // intelliJ tells me to remove this, because it's "always false"!
      if (!(x instanceof Integer)) {
        throw new IllegalArgumentException("x is not an Integer: " + x);
      }
   }
```

The following fuzz test runs into the requirement error in `resolveTypeArguments() in ParameterizedTypeSupport.java:75`
Expected: `Arguments[[String] -> Child]`
```java
public static class Base<T> {
     protected T x;

     public void setX(T x) { this.x = x; }
     public T getX() { return x; }
   }
   // Here U is not related to getters and setters
   public static class Child<T, U> extends Base<T> { }

   @Solo
   @SelfFuzzTest
   public static void fuzzGenericClass(@NotNull(constraint = PropertyConstraint.RECURSIVE) Child<String, Integer> pair) { }
```

Here are the values of involved variables:

`clazz`: class com.code_intelligence.selffuzz.mutation.ArgumentsMutatorFuzzTest$Base  
`typeParameters`: [T]
`classType`: @com.code_intelligence.selffuzz.jazzer.mutation.annotation.NotNull(constraint="JAZZER_PROPERTY_CONSTRAINT_RECURSIVE") com.code_intelligence.selffuzz.mutation.ArgumentsMutatorFuzzTest$Child<java.lang.String, java.lang.Integer>
`typeArguments`: [java.lang.String, java.lang.Integer]

src/main/java/com/code_intelligence/jazzer/mutation/support/ParameterizedTypeSupport.java
```diff
@@ -0,0 +1,492 @@
...
+      AnnotatedType replacement = mapping.get(type);
+      if (replacement == null) {
+        return annotated;
+      }
+      return TypeSupport.forwardAnnotations(annotated, replacement);
+    }
+    return annotated;
+  }
+
+  private static AnnotatedParameterizedType resolveParameterizedType(
```
oetr, line 148: Because the reflection API has both `ParameterizedType` and `AnnotatedParameterizedType`, I find this name imprecise. Should we rename all functions that return annotated types into `...Annotated...` and those that return `Type` into `...Raw...`?

src/main/java/com/code_intelligence/jazzer/mutation/mutator/aggregate/BeanSupport.java
```diff
@@ -49,6 +52,32 @@ static Optional<Class<?>> optionalClassForName(String targetClassName) {
     }
   }
 
+  // Returns the annotated parameter types of a method or constructor resolving all generic type
+  // arguments.
+  public static AnnotatedType[] resolveAnnotatedParameterTypes(
+      Executable e, AnnotatedType classType) {
+    Type[] generic = e.getGenericParameterTypes();
```
oetr: `generic` is used only to get the range that we use to access another array `annotated`. Why isn't the length of `annotated` used directly? In that we could use a stream! ^.^
```java
public static AnnotatedType[] resolveAnnotatedParameterTypes(
        Executable e, AnnotatedType classType) {
    return Arrays.stream(e.getAnnotatedParameterTypes())
          .map(a -> resolveTypeArguments(e.getDeclaringClass(), classType, a))
          .toArray(AnnotatedType[]::new);
  }
```

src/main/java/com/code_intelligence/jazzer/mutation/mutator/aggregate/BeanSupport.java
```diff
@@ -88,9 +117,11 @@ static Optional<Method[]> findGettersByPropertyNames(
...
   }
 
-  static Optional<Method[]> findGettersByPropertyTypes(Class<?> clazz, Stream<Class<?>> types) {
-    Map<Class<?>, List<Method>> gettersByType =
-        findMethods(clazz, BeanSupport::isGetter).collect(groupingBy(Method::getReturnType));
+  static Optional<Method[]> findGettersByPropertyTypes(
+      Class<?> clazz, AnnotatedType classType, Stream<Type> types) {
+    Map<Type, List<Method>> gettersByType =
+        findMethods(clazz, BeanSupport::isGetter)
+            .collect(groupingBy(m -> resolveReturnType(m, classType)));
```
oetr, line 135: If `m` is not declared in `classType`, which here is the instantiated class of the concrete bean, the resolution doesn't work, as seen in the examples in another comment below.
To fix it, we should probably resolve all necessary generic types untill we reach the superclass where `m` is declared.
The same applies for the return types of the setters.

src/main/java/com/code_intelligence/jazzer/mutation/support/ParameterizedTypeSupport.java
simonresch: Great test cases! Yes, generic superclasses were completely missing in the implementation.

I added another commit that walks up all the superclasses (and dependencies) to build up a more accurate mapping of type variables. That seems to cover all the cases you found not to be working.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/aggregate/BeanSupport.java
simonresch: Yes, absolutely.

src/main/java/com/code_intelligence/jazzer/mutation/support/ParameterizedTypeSupport.java
simonresch: If there is only one version of those functions I would personally prefer the shorter name. Its clear from the return type, that we get an `AnnotatedType`, no?

src/main/java/com/code_intelligence/jazzer/mutation/mutator/aggregate/BeanSupport.java
simonresch: I believe the additions in `ParameterizedTypeSupport.java` https://github.com/CodeIntelligenceTesting/jazzer/pull/1008/commits/a12e29d2b65b62f122f439cb730f2ca80fbaa376 should take care of this by going up the entire superclass chain to resolve all type variables.

src/main/java/com/code_intelligence/jazzer/mutation/support/ParameterizedTypeSupport.java
```diff
@@ -0,0 +1,525 @@
...
+      return annotatedType.getDeclaredAnnotations();
+    }
+
+    public AnnotatedType getAnnotatedOwnerType() {
+      return getAnnotatedOwnerTypeOrNull(annotatedType);
+    }
+
+    @Override
+    public String toString() {
+      // TODO: include annotations in string
```
oetr, line 375: Is this still a TODO?

src/main/java/com/code_intelligence/jazzer/mutation/mutator/aggregate/BeanSupport.java
```diff
@@ -49,6 +54,42 @@ static Optional<Class<?>> optionalClassForName(String targetClassName) {
     }
   }
 
+  private static Class<?> rawType(Type classType) {
```
oetr: This function is interesting because it has a bug compared to the recently added `extractRawClass` in `TypeSupport` when we go into `GenericArrayType` if-branch, because it will return the base type T.class and not the array type T[].class. 

Here is an example test:
```java
@Test
  void beanTest() {
    Type testType = new TypeHolder<List<List<String>>[]>() {}.type();
    System.out.println("B: " + BeanSupport.rawType(testType));
    System.out.println("T: " + TypeSupport.extractRawClass(testType).get());
  }
// B: interface java.util.List
// T: class [Ljava.util.List;
```
But I couldn't generate a test case to reach this if-branch from top level with a `@FuzzTest`.
As it turns out, the base class of a bean can never be a `GenericArrayType`!

So maybe we should 1) also throw an error for `GenericArrayType` and add a short comment? And also update the name of this function to make it more fitting (maybe `extractBeanClass`)?

Or 2) make it properly handle generic array types as in `TypeSupport.extractRawClass`.

I prefer 1) because if our bean class is a generic array type, something went wrong!

As a nice side-effect, while double checking this I discovered a bug in our list mutator: #1020

src/main/java/com/code_intelligence/jazzer/mutation/support/ParameterizedTypeSupport.java
```diff
@@ -64,21 +64,50 @@ public class ParameterizedTypeSupport {
...
+    if (mapping.isEmpty()) {
       return type;
     }
+    return resolveRecursive(type.getType(), type, mapping);
+  }
 
-    TypeVariable<?>[] typeParameters = clazz.getTypeParameters();
-    AnnotatedType[] typeArguments =
-        ((AnnotatedParameterizedType) classType).getAnnotatedActualTypeArguments();
+  private static void collectTypeMappings(
```
oetr: Minor: `collect` sounds functional to me! A name hinting at mutation might be better. Maybe `updateTypeMappings`?

src/main/java/com/code_intelligence/jazzer/mutation/support/ParameterizedTypeSupport.java
simonresch: Maybe for later. ;)

src/main/java/com/code_intelligence/jazzer/mutation/mutator/aggregate/BeanSupport.java
simonresch: Good catch! That is of course a bug. I went with your first approach of throwing an exception. We should never get there because the `ArrayMutatorFactory` applies first and strips the array part so we don't need to handle array types here.

PR #1009, opened 2025-11-05
fix: properly instrument nested records

oetr:
Using ClassWriter without ClassReader in asm, seems to produce
corrupted bytecode for nested records and records with annotated fields.

Prior to this fix, a segfault was triggered for nested records that
use Jazzer annotations when Jazzer was trying to access data on record
components, when trying to create a record mutator.

In addition, for non-nested records with annotated fields as for example:
`record Address(byte @WithLength(max=10) [] data) {}`
no suitable mutator could be found.

### DIFF
--- a/src/main/java/com/code_intelligence/jazzer/instrumentor/TraceDataFlowInstrumentor.kt
+++ b/src/main/java/com/code_intelligence/jazzer/instrumentor/TraceDataFlowInstrumentor.kt
@@ -52,7 +52,7 @@ internal class TraceDataFlowInstrumentor(
             }
         }
 
-        val writer = ClassWriter(ClassWriter.COMPUTE_MAXS)
+        val writer = ClassWriter(reader, ClassWriter.COMPUTE_MAXS)
         node.accept(writer)
         return writer.toByteArray()
     }
--- a/tests/BUILD.bazel
+++ b/tests/BUILD.bazel
@@ -1078,6 +1078,35 @@ java_fuzz_target_test(
 )
 
 java_fuzz_target_test(
+    name = "NestedRecordFuzzer",
+    timeout = "short",
+    srcs = [
+        "src/test/java/com/example/NestedRecordFuzzer.java",
+    ],
+    fuzzer_args = [
+        "-print_final_stats=1",
+        "-runs=10000",
+    ],
+    javacopts = [
+        "--release",
+        "17",
+    ],
+    tags = [
+        "no-jdk8",
+    ],
+    target_class = "com.example.NestedRecordFuzzer",
+    runtime_deps = [
+        "@maven//:org_junit_jupiter_junit_jupiter_engine",
+    ],
+    deps = [
+        "//deploy:jazzer-junit",
+        "//deploy:jazzer-project",
+        "@maven//:org_junit_jupiter_junit_jupiter_api",
+        "@maven//:org_junit_jupiter_junit_jupiter_params",
+    ],
+)
+
+java_fuzz_target_test(
     name = "BigDecimalFuzzer",
     srcs = [
         "src/test/java/com/example/BigDecimalFuzzer.java",
--- a/tests/src/test/java/com/example/NestedRecordFuzzer.java
+++ b/tests/src/test/java/com/example/NestedRecordFuzzer.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2024 Code Intelligence GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example;
+
+import com.code_intelligence.jazzer.junit.FuzzTest;
+import com.code_intelligence.jazzer.mutation.annotation.NotNull;
+import com.code_intelligence.jazzer.mutation.annotation.WithLength;
+
+/**
+ * This regression test checks if we instrument fuzz test classes with nested records correctly, and
+ * can start a fuzzing run at all without getting segfaults.
+ */
+public class NestedRecordFuzzer {
+  record Address(@NotNull String street, byte @NotNull @WithLength(min = 3, max = 4) [] data) {}
+
+  @FuzzTest
+  public void test(Address address) {
+    if (address != null) {
+      byte[] data = address.data();
+      if (data.length < 3 || data.length > 4) {
+        throw new IllegalStateException("Data length out of bounds: " + data.length);
+      }
+    }
+  }
+}


src/main/java/com/code_intelligence/jazzer/instrumentor/TraceDataFlowInstrumentor.kt
```diff
@@ -52,7 +52,7 @@ internal class TraceDataFlowInstrumentor(
             }
         }
 
-        val writer = ClassWriter(ClassWriter.COMPUTE_MAXS)
+        val writer = ClassWriter(reader, ClassWriter.COMPUTE_MAXS)
```
fmeum, line 55: How did you figure out that this was causing the bug? 😅

oetr: I am still unsure if this is the 1) right solution, or 2) if we should do this only for records; or 3) if it's a bug in asm.

My debugging approach was to get as far as possible with manual work and then ask AI (claude free vs) before I read any docs myself! Now that the segfault issue seems to be solved, I am actually reading the docs and the source code for [ClassWriter](https://asm.ow2.io/developer-guide.html#classwriter) to figure out which point above is true (1-3). 

After initial analysis, I could minimize the reproducer to the code below. Removing the `@NotNull` annotation, _or_ moving the record to the top level, _or_ using a local annotation, or some other java annotation (e.g. we tried `@Deprecated`) works fine. It must be a Jazzer annotation:
```java
package reproducer;

import com.code_intelligence.jazzer.mutation.annotation.NotNull;
import com.code_intelligence.jazzer.junit.FuzzTest;

public class ReproduceCrash {
  record NestedRecordWithExternalAnnotation(@NotNull String fileName) {}

  @FuzzTest
  public void fuzz_test (boolean ignored) {
    var components = NestedRecordWithExternalAnnotation.class.getRecordComponents();
  }
}
```
Here is the stack trace after a segfault that happens in fuzzing mode only, and not in regression mode:
```text
V  [libjvm.so+0x6b09c0]  ConstantPool::symbol_at(int) const+0x5c  (constantPool.hpp:442)
V  [libjvm.so+0xd0ba3f]  java_lang_reflect_RecordComponent::create(InstanceKlass*, RecordComponent*, JavaThread*)+0x245  (javaClasses.cpp:3398)
V  [libjvm.so+0xe22b58]  JVM_GetRecordComponents+0x2d4  (jvm.cpp:1828)
```

I narrowed down the cause to `TraceDataFlowInstrumentor.instrument()` by tracing the class with the nested record. Then I saw that simply reading and then writing back the bytecode already triggers the segfault (I removed the for-loop for methods):
```java
val node = ClassNode()
val reader = ClassReader(bytecode)
reader.accept(node, 0)
random = DeterministicRandom("trace", node.name)
//for (method in node.methods) {
//    if (shouldInstrument(method)) {
//        addDataFlowInstrumentation(method)
//    }
//}
val writer = ClassWriter(ClassWriter.COMPUTE_MAXS)
node.accept(writer)
return writer.toByteArray()
```
At this point I asked AI, and its first suggestion was to add `COMPUTE_FRAMES` reader/writer options, which didn't solve the segfault.
The second try was to add reader to the writer, and here we are!

fmeum: If you save the transformed class file to a file and load it in a separate JVM process, does that also trigger the crash? This looks like a JVM bug first and foremost, but reporting one would probably require a reproducer that doesn't use as many unsafe tricks as Jazzer in fuzzing mode. 😉

oetr: I am still debugging it, and my current guess is that ASM mangles the constant pool when writing without reusing the symbol table of the reader. It has something to do with the annotation, which doesn't even have to be a Jazzer annotation. For example, this also causes a segfault:
```java
package com.example;

import com.code_intelligence.jazzer.junit.FuzzTest;

import java.lang.annotation.Retention;
import java.lang.annotation.Target;

import static java.lang.annotation.ElementType.TYPE_USE;
import static java.lang.annotation.RetentionPolicy.RUNTIME;



public class NestedRecordFuzzer {
  record Address(@NotNull int a) {}

  @FuzzTest
  public void test(Address ignored) {
  }
}

@Target(TYPE_USE)
@Retention(RUNTIME)
@interface NotNull {}
```
I hacked my local jdk to print the value when it fails its assertion in `constantPool.hpp:441 symbol_at()`, and I get Constant pool tag at index 12 is 100. Maybe a 'd' in `"java/lang/invoke/MethodHandles$Lookup"`? I will find it out!

Why do you think it's a JVM bug and not ASM?

fmeum: It is probably also an ASM bug (it generates a corrupted class file), but the JVM must not crash even on an invalid class file.

oetr: Ah, I assumed that segfaulting after running into assertion errors is the standard JVM way:
```c++
  Symbol* symbol_at(int which) const {
    assert(tag_at(which).is_utf8(), "Corrupted constant pool");
    return *symbol_at_addr(which);
  }
```
These are all over the place in constantPool.hpp, etc.

fmeum: My assumption has been that there is a verifier pass that should avoid all these asserts so that pure Java can't hit them, but maybe that's not correct.

oetr: @fmeum Debugging this a bit further, the segfault happens only when using JUnit. Jazzer standalone doesn't run into it, but has problems with annotated record fields.
Checking the difference reveals that with JUnit, the fuzz test class and the nested record are loaded before Jazzer installs the instrumentor, and hence get marked for retransformation. During class redefinition,  the JVM returns an error when trying to merge constant pools, but somehow doesn't handle it. In `merge_cp_and_rewrite` function we run into this if-statement and return the error. So the record *is* broken, and it's an ASM bug:
```c++
if (!rewrite_cp_refs(scratch_class)) {
      return JVMTI_ERROR_INTERNAL; 
}
```

But now `load_new_class_versions()` doesn't check the return value `res`:
```c++
  res = merge_cp_and_rewrite(the_class, scratch_class, THREAD);
  if (HAS_PENDING_EXCEPTION) {
    Symbol* ex_name = PENDING_EXCEPTION->klass()->name();
    log_info(redefine, class, load, exceptions)("merge_cp_and_rewrite exception: '%s'", ex_name->as_C_string());
    CLEAR_PENDING_EXCEPTION;
    if (ex_name == vmSymbols::java_lang_OutOfMemoryError()) {
      return JVMTI_ERROR_OUT_OF_MEMORY;
    } else {
      return JVMTI_ERROR_INTERNAL;
    }
  }
```
As seen in the logs, the JVM continues with verification of the redefined record, so that `HAS_PENDING_EXCEPTION` mechanism is not working properly here. Maybe directly checking res would be a better option:
```code
[52.292s][info ][redefine,class,constantpool] old_cp_len=58, scratch_cp_len=58
[52.292s][debug][redefine,class,constantpool] after pass 0: merge_cp_len=58
[52.293s][debug][redefine,class,constantpool] after pass 1a: merge_cp_len=58, scratch_i=58, index_map_len=57
[52.293s][info ][redefine,class,constantpool] merge_cp_len=58, index_map_len=57
[52.293s][debug][redefine,class,annotation  ] num_annotations=1
[52.293s][debug][redefine,class,annotation  ] type_index=4864  num_element_value_pairs=16
[52.293s][debug][redefine,class,annotation  ] element_name_index=0
[52.293s][debug][redefine,class,annotation  ] length() is too small for a tag
[52.293s][debug][redefine,class,annotation  ] bad element_value at 0
[52.293s][debug][redefine,class,annotation  ] bad annotation_struct at 0
[52.293s][debug][redefine,class,annotation  ] bad record_component_type_annotations at 0
[52.293s][debug][class,loader,data          ] deallocate added for constant pool [116]/operands[7] {0x00007f783fae7570}
[52.293s][info ][class,init                 ] Start class verification for: reproducer.ReproduceCrashLong$MyRecord
```

So, JVM redefinition does notice that the record is mangled, but fails to handle its own error and makes the record even more broken!

Overall, the fix above seems correct. Now I gotta file the two bug reports and work out some small reproducers :sweat:

PBetzler: Good job @oetr!

PR #1010, opened 2025-11-05
Mention `Set` in `WithSize` Javadoc & slightly improve test

Marcono1234:


selffuzz/src/test/java/com/code_intelligence/selffuzz/mutation/ArgumentsMutatorFuzzTest.java
```diff
@@ -112,13 +112,11 @@ void fuzzListOfMaps(@WithSize(max = 4) Map<String, Integer> nullableMap) {
 
   @SelfFuzzTest
   void fuzzListOfSets(@WithSize(max = 10) @NotNull Set<@NotNull Integer> setWithSize) {
-    if (setWithSize != null) {
-      assertThat(setWithSize.size()).isAtMost(10);
-    }
+    assertThat(setWithSize.size()).isAtMost(10);
```
Marcono1234, line 115: I think the `null` check was redundant because the parameter is annotated with `@NotNull`

PR #1012, opened 2025-11-06
docs: document seed input sources for @FuzzTest

simonresch:


README.md
```diff
@@ -200,6 +200,16 @@ For example, if the fuzz test is in the class `src/test/java/com/example/ValidFu
...
+## Seeding via JUnit parameter sources
+
+You can provide initial seed inputs for a `@FuzzTest` using standard JUnit parameter sources such as `@MethodSource`,
+`@CsvSource`, `@ValueSource`, or custom `@ArgumentsSource`. In [regression mode](#regression-mode), Jazzer executes the
+fuzz test once for each provided argument set, similar to a JUnit `@ParameterizedTest`.
+In [fuzzing mode](#fuzzing-mode), the seed inputs are additionally used for further mutations.
+
+An example showcasing JUnit’s `@MethodSource` for a `@FuzzTest` can be found in
+[`JavaSeedFuzzTest.java`](examples/junit/src/test/java/com/example/JavaSeedFuzzTest.java).
+
```
oetr, line 212: Maybe we should add that it does not work for every type, and list explicitly which types are not supported.

oetr (approved):
LGTM, thanks!

README.md
simonresch: I added a note that cached constructor based classes don't work. I'm not aware of other mutators that can't serialize an object of the mutated type. Do you think there is other types we don't support?

PR #1013, opened 2025-11-07
feat: sometimes interpret char[] mutations as single bytes

florianGla:
When mutating char[] randomly interpret the bytes from libFuzzer as individual (single byte) chars. This helps to make use of libFuzzers table of recent compare entries (encoded as CESU8) if the char[] is used as a String inside the fuzz test.

### DIFF
--- a/src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/PrimitiveArrayMutatorFactory.java
+++ b/src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/PrimitiveArrayMutatorFactory.java
@@ -274,16 +274,23 @@ final class PrimitiveArrayMutatorFactory implements MutatorFactory {
       } else {
         char[] chars = new String(bytes, FUZZED_DATA_CHARSET).toCharArray();
 
-        for (int i = 0; i < chars.length; i++) {
-          chars[i] = (char) forceInRange(chars[i], minRange, maxRange);
-        }
+        int targetLength = Math.min(Math.max(chars.length, minLength), maxLength);
 
-        if (chars.length < minLength) {
-          return Arrays.copyOf(chars, minLength);
-        } else if (chars.length > maxLength) {
-          return Arrays.copyOf(chars, maxLength);
+        if (chars.length == targetLength) {
+          for (int i = 0; i < chars.length; i++) {
+            chars[i] = (char) forceInRange(chars[i], minRange, maxRange);
+          }
+          return chars;
+        } else {
+          char[] result = new char[targetLength];
+          for (int i = 0; i < targetLength; i++) {
+            result[i] =
+                i < chars.length
+                    ? (char) forceInRange(chars[i], minRange, maxRange)
+                    : (char) forceInRange(0, minRange, maxRange);
+          }
+          return result;
         }
-        return chars;
       }
     }
 


tests/src/test/java/com/example/CharArrayFuzzer.java
```diff
@@ -0,0 +1,29 @@
...
+
+package com.example;
+
+public class CharArrayFuzzer {
+  public static void fuzzerTestOneInput(char[] data) {
+    if (data == null) {
+      return;
+    }
+    String expression = new String(data);
+    if (expression.contains("jazzer")) {
```
fmeum: Could be worth testing with non-Latin1 Unicode strings as targets - with pure ASCII you can't be sure whether all of the CESU8 to char[] conversion is done correctly.

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/PrimitiveArrayMutatorFactory.java
```diff
@@ -250,6 +254,29 @@ private static AnnotatedType convertWithLength(AnnotatedType type, AnnotatedType
...
     }
 
+    // The strings we pass to native callbacks to trace data flow are CESU-8 encoded.
+    // As a result, libFuzzer's TORC contains CESU-8 encoded strings.
+    // Therefore, in 50% of times we decode the byte array as a CESU-8  string.
+    public char[] postMutateChars(byte[] bytes, PseudoRandom prng) {
+      if (prng.choice()) {
+        return (char[]) toPrimitive.apply(bytes);
+      } else {
+        char[] chars = new String(bytes, Charset.forName("CESU-8")).toCharArray();
```
oetr: > This doesn't seem to respect tha length provided by `@WithLength` annotation. 

Oh, it does, never mind!

Marcono1234: Out of curiosity: Doesn't the current code here (specifically `convertWithLength`) assume that 2 `byte` = 1 `char`?
Wouldn't this assumption be violated here because:
- 1 CESU-8 `byte` can be converted to 1 `char` (for ASCII chars) → array becomes too large
- 3 CESU-8 `byte` can be converted to 1 `char` (for chars >= U+0800) → array becomes too short

Maybe I am overlooking something though; sorry for the trouble in that case.

(Also side note: Would it make sense to store the `Charset.forName("CESU-8")` in a `static final` field?)

oetr: I also couldn't see how the length gets respected here, and did some fuzzing with various `@WithLength(min=..., max=...)` values, and couldn't get the array length out of bounds. So it seems to be respected, for some reason!

>Would it make sense to store the Charset.forName("CESU-8") in a static final field?

It would make sense, I think.

kyakdan: @oetr The length annotation is enforced by the `innerMutator`, which is used by `toPrimitive` and `toPrimitiveAfterMutate`: see https://github.com/CodeIntelligenceTesting/jazzer/blob/CIF-1863-string-compares-on-char-arrays/src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/PrimitiveArrayMutatorFactory.java#L95

kyakdan: @Marcono1234 You are right regarding the `convertWithLength` assuming two bytes per char, which could lead to enforcing incorrect length constraints when interepreting the byte array as `"CESU-8"`-encoded.  I adjusted the `convertWithLength` method to assume a maximum of 6 bytes for a char (the maximum number of bytes used with `CESU-8`). Then, when we do the actual conversion to a char array, we ensure to enforce the length constraints. @oetr Could you have a look at the last commit?

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/PrimitiveArrayMutatorFactory.java
```diff
@@ -209,11 +213,15 @@ private void extractRange(AnnotatedType type) {
       }
     }
 
-    private static AnnotatedType convertWithLength(AnnotatedType type, AnnotatedType newType) {
-      AnnotatedType elementType = ((AnnotatedArrayType) type).getAnnotatedGenericComponentType();
+    private void extractLength(AnnotatedType type) {
       Optional<WithLength> withLength = Optional.ofNullable(type.getAnnotation(WithLength.class));
-      int minLength = withLength.map(WithLength::min).orElse(DEFAULT_MIN_LENGTH);
-      int maxLength = withLength.map(WithLength::max).orElse(DEFAULT_MAX_LENGTH);
+      withLength.ifPresent(System.err::println);
```
oetr: can be removed

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/PrimitiveArrayMutatorFactory.java
```diff
@@ -263,9 +276,16 @@ public char[] postMutateChars(byte[] bytes, PseudoRandom prng) {
         return (char[]) toPrimitive.apply(bytes);
       } else {
         char[] chars = new String(bytes, FUZZED_DATA_CHARSET).toCharArray();
+
         for (int i = 0; i < chars.length; i++) {
           chars[i] = (char) forceInRange(chars[i], minRange, maxRange);
         }
+
+        if (chars.length < minLength) {
+          return Arrays.copyOf(chars, minLength);
```
oetr: Here, the newly padded chars might not be in range.

Maybe we can do a single-pass copy and forceInRange here? Or is Java smart enough to optimize it at runtime?
WDYT about the following:
```java
       int targetLength = Math.min(Math.max(chars.length, minLength), maxLength);

        if (chars.length == targetLength) {
          for (int i = 0; i < chars.length; i++) {
            chars[i] = (char) forceInRange(chars[i], minRange, maxRange);
          }
          return chars;
        } else {
          char[] result = new char[targetLength];
          for (int i = 0; i < targetLength; i++) {
            result[i] = i < chars.length
                  ? (char) forceInRange(chars[i], minRange, maxRange)
                  : (char) forceInRange(0, minRange, maxRange);
          }
          return result;
        }
```

oetr (commented):
Looks good, just one issue from my side!

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/PrimitiveArrayMutatorFactory.java
oetr: Nice catch @Marcono1234 , thanks!

Marcono1234: Isn't this `* 6` for a complete Unicode code point, specifically a code point >= 0x10000?
However a Java `char` is UTF-16, so those code points >= 0x10000 would actually require 2 `char`.

So wouldn't it suffice to do `* 3` here?

Or at worst `* 4` in case a malformed 4-byte CESU-8 encoding can lead to a single replacement `?` char.

kyakdan: @Marcono1234  I think you're right: 3 bytes should be enough to represent a single char

src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/PrimitiveArrayMutatorFactory.java
kyakdan: Good catch regarding the padding chars potentially falling out of range. I adopted your implementation.

PR #1016, opened 2025-11-13
almost automatic release to Maven Central

oetr:
Just push the tag to release.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -131,11 +131,13 @@ Run `./format.sh` to format all source files in the way enforced by the "Check f
 
 ## Releasing (CI employees only)
 
-1. Push a tag of the form `v1.2.3` to trigger the "Prerelease" GitHub Actions workflow.
-2. Wait for the workflow to finish (about 10 minutes).
-3. Download the artifact called "jazzer-maven-central-bundle" from the prerelease workflow run.
-4. Log into https://central.sonatype.com/publishing and click on "Publish Component".
-5. In the dialog that follows, under "Upload Your File", select the downloaded artifact. The "Deployment Name" doesn't matter. Click "Publish Component".
-6. Refresh the page, and you should see "Deployment Info" with status "VALIDATED".
-7. Click "Publish" to publish the component to Maven Central and wait for the status to change to "PUBLISHED" (a refresh or two may be necessary).
-8. Release the draft Github release. This will automatically create a tag, push the docker images and deploy the docs (can take about a few minutes to appear at [jazzer-docs]( https://codeintelligencetesting.github.io/jazzer-docs)).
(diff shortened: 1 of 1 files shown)


.github/workflows/prerelease.yaml
```diff
@@ -161,7 +176,6 @@ jobs:
         uses: softprops/action-gh-release@6da8fa9354ddfdc4aeace5fc48d7f679b5214090 # v2.4.1
         with:
           generate_release_notes: true
-          draft: true
```
simonresch, line 164: IMO it could still be useful to have a manual process to publish from a draft github release so that we can take a look and clean up the release notes.

.github/workflows/prerelease.yaml
```diff
@@ -128,19 +128,34 @@ jobs:
...
+      - name: Deploy to Maven Central
+        shell: bash
+        run: |
+          TAG=${{ github.ref_name }}
+          curl --request POST \
+                --verbose \
+                --fail-with-body \
+                --header "Authorization: Bearer ${{ secrets.SONATYPE_BEARER_TOKEN }}" \
+                --form bundle=@_tmp/jazzer-maven-central-bundle.tar.gz \
+                "https://central.sonatype.com/api/v1/publisher/upload?name=Jazzer%20${TAG#v}&publishingType=AUTOMATIC"
```
simonresch, line 155: Do you know how error during publishing would show up?
We could wait for successful publishing by polling the status API https://central.sonatype.org/publish/publish-portal-api/#verify-status-of-the-deployment. Or just make it a manual step to verify that the release was published.

.github/workflows/prerelease.yaml
oetr: We don't do this for cifuzz, why is Jazzer different?
But combined with the point below it makes sense to delay the release on GH until the dev has verified that it's actually published. I'll revert this!

.github/workflows/prerelease.yaml
oetr: Ah, yes, I though about it, but deemed not worth the effort!
Maybe the dev should just check the status and then release the GH draft after the publishing was successful?

simonresch: Sounds good. Making it a two step process is already a lot better than the current.

.github/workflows/prerelease.yaml
simonresch: > We don't do this for cifuzz, why is Jazzer different?

My hope would be that the Jazzer release notes are better than the cifuzz once. ;) Reading through the change log once during release would make sense IMO but if we take care to review all the PR titles we could also make it fully automatic.

oetr: The most useful would be to let some GH action auto-arrange PR messages into categories like features, bug fixes, ci/cd, dependency updates, etc.
The releasing dev can (but is not forced to) re-sort individual items based on some subjective significance, e.g. coolest feature goes on top.
I think our PR names are already pretty thoughtful.

PR #1017, opened 2025-11-20
fix: handle null values in Arrays.equal/compare hooks

simonresch:
`null` arrays would crash in the `Arrays.equal` and `Arrays.compare` hooks.

### DIFF
--- a/src/main/java/com/code_intelligence/jazzer/runtime/TraceCmpHooks.java
+++ b/src/main/java/com/code_intelligence/jazzer/runtime/TraceCmpHooks.java
@@ -744,6 +744,7 @@ public final class TraceCmpHooks {
     if (returnValue) return;
     byte[] first = (byte[]) arguments[0];
     byte[] second = (byte[]) arguments[1];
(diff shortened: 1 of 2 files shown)


src/main/java/com/code_intelligence/jazzer/runtime/TraceCmpHooks.java
```diff
@@ -744,6 +744,7 @@ public static void arraysEquals(
     if (returnValue) return;
     byte[] first = (byte[]) arguments[0];
     byte[] second = (byte[]) arguments[1];
+    if (first == null || second == null) return;
```
oetr, line 747: For a minute I was thinking that maybe this might be worth a try:
```java
TraceDataFlowNativeCallbacks.traceMemcmp(first != null ? first : new byte[0] , second != null ? second : new byte[0], 1, hookId);
```
But then I remembered, that randomly inserting strings (without knowing where to insert, because one arg has no data) is almost worthless, so it's best to just exit, like you do.

PR #1018, opened 2025-11-24
feat: SSRF env var to allow all connections before it's configured

oetr:
In some projects, network requests can happen after the fuzz test has started, but before the user had a chance to configure allowed/denied connections.

Now the user can start Jazzer with the environmental variable JAZZER_SSRF_PERMISSIVE_UNTIL_CONFIGURED set to a truthy value and all network requests will be allowed, until the user specifies otherwise in the fuzz test using BugDetectors.allowNetworkConnections(...)

### DIFF
--- a/sanitizers/src/test/java/com/example/SsrfAllowUntilConfigured.java
+++ b/sanitizers/src/test/java/com/example/SsrfAllowUntilConfigured.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2024 Code Intelligence GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example;
+
+import com.code_intelligence.jazzer.api.BugDetectors;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.util.concurrent.CountDownLatch;
+
+public class SsrfAllowUntilConfigured {
+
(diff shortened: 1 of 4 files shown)


sanitizers/src/test/java/com/example/SsrfAllowUntilConfigured.java
```diff
@@ -0,0 +1,58 @@
...
+            });
+    backgroundThread.setDaemon(true);
+    backgroundThread.start();
+  }
+
+  public static void fuzzerTestOneInput(boolean ignored) throws Exception {
+    fuzzTestStarted.countDown();
+    Thread.sleep(500); // Ensure background thread has time to run
+
+    BugDetectors.allowNetworkConnections((host, port) -> host.equals("localhost"));
```
simonresch, line 56: As I understand the `allowNetworkConnections` can not be configured in the `static {}` block directly because the allowed connections would be overwritten on fuzzer start because of

```
 Jazzer.onFuzzTargetReady(() -> connectionPermitted.set((host, port) -> false));
```

in the SSRF bug detector, correct? Couldn't we check at this point if `allowNetworkConnections` has been configured already and then not overwrite that property? That way users could control exactly at what point network restrictions should be activated and wouldn't have to configure an environment variable. Or am I missing something?

oetr: Can we be sure that the static block of the fuzz test class is always executed before the callback `Jazzer.onFuzzTargetReady` is run? In standalone and in junit? By using an environment variable, this issue is moot regardles of user setup. It guarantees 0 false-positives.

Also, sometimes in some fuzz tests (e.g. kafka), calling `BugDetectors`... in the static block will result in a `ClassNotFoundException` because the Sanitizer class has not been loaded yet. To see what it looks like, try calling BugDetectors.allowConnections() in a junit `@BeforeAll` block. So in kafka, I am currently not able to do that from a static block.

It would be really nice if users could call BugDetectors API from wherever they like (static blocks, and `@BeforeAll`) without worrying about class loading order.
To solve this probably needs a bit of reflective ping pong :smile:

simonresch: You're right, the reliance on reflection in the `BugDetectors.java` makes it tricky to use in `@BeforeAll` functions. Thanks for clarifying. The environment variable looks like a good solution. :+1:

simonresch (approved):
Looks good.

PR #1019, opened 2025-11-24
feat: add char and Character mutator

oetr:
We forgot about chars in the past!

Not sure if `@Positive` and `@Negative` , etc. make sense on chars (the only signed integral in Java? :laughing:)
`@InRange` can be used like in the sample fuzz test. 

```java
fuzzerTestOneInput(@InRange(min = '\u0000', max = '\uFFFF') char data) {
```


src/main/java/com/code_intelligence/jazzer/mutation/mutator/lang/IntegralMutatorFactory.java
```diff
@@ -112,6 +112,39 @@ public void write(Short value, DataOutputStream out) throws IOException {
...
+            }
+
+            @Override
+            public Character read(DataInputStream in) throws IOException {
+              return (char) forceInRange(in.readChar());
+            }
+
+            @Override
+            public void write(Character value, DataOutputStream out) throws IOException {
+              out.writeShort(value);
```
Marcono1234: Out of curiosity, why not use `writeChar` here for consistency with the `readChar` call above?

(though `DataOutput` defines both `writeChar` and `writeShort` to be implemented in the same way, so this might not matter)

oetr: The idea behind using `writeShort` here was sloppy copy & paste. Nice catch, thanks!

tests/src/test/java/com/example/CharFuzzer.java
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+
+package com.example;
+
+import com.code_intelligence.jazzer.api.FuzzerSecurityIssueLow;
+import com.code_intelligence.jazzer.mutation.annotation.InRange;
+
+public class CharFuzzer {
+  public static void fuzzerTestOneInput(@InRange(min = '\u0000', max = '\uFFFF') char data) {
```
kyakdan: I'd limit the range and add an assert that `data` is inside the specified range.

kyakdan (approved):

tests/src/test/java/com/example/CharFuzzer.java
oetr: Good idea, I've added a range and checks.

PR #1020, opened 2025-11-26
fix: support lists of arbitrary types

oetr:
Previously, types like `List<List<Integer>[]>` failed because the
`parameterTypeIfParameterized` helper function incorrectly filtered out
generic array types. Removed this filtering since the mutation
framework already handles supported type matching.


simonresch (approved):
Nice! Good find.

PR #1021, opened 2025-11-30
trophies: add entry for lz4-java

Marcono1234:
CC @yawkat, is it ok that I mentioned you like this in the new trophies entry?


simonresch:
Thanks for adding this here. Good find!

simonresch (approved):

PR #1051, opened 2026-03-10
feat: add hooks for temporal parse and string compare in equals methods

onionpsy:
Add hooks for `java.time`:

  - add guidance for equality checks
  - adds parsing hooks so failed time/date parses still provide useful guidance to the fuzzer

### DIFF
--- a/src/main/java/com/code_intelligence/jazzer/runtime/TraceCmpHooks.java
+++ b/src/main/java/com/code_intelligence/jazzer/runtime/TraceCmpHooks.java
@@ -19,6 +19,20 @@ package com.code_intelligence.jazzer.runtime;
 import com.code_intelligence.jazzer.api.HookType;
 import com.code_intelligence.jazzer.api.MethodHook;
 import java.lang.invoke.MethodHandle;
+import java.time.Duration;
+import java.time.Instant;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.MonthDay;
+import java.time.OffsetDateTime;
+import java.time.OffsetTime;
+import java.time.Period;
+import java.time.Year;
+import java.time.YearMonth;
+import java.time.ZoneId;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
 import java.util.*;
 
 @SuppressWarnings("unused")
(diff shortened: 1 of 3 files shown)


src/main/java/com/code_intelligence/jazzer/runtime/TraceCmpHooks.java
```diff
@@ -234,6 +248,172 @@ public static void equals(
     }
   }
 
+  private static final LocalDate FALLBACK_LOCAL_DATE = LocalDate.of(2000, 1, 1);
+  private static final ZonedDateTime FALLBACK_ZONED_DATE_TIME =
+      ZonedDateTime.of(FALLBACK_LOCAL_DATE, LocalTime.MIDNIGHT, ZoneId.of("Europe/Paris"));
+  private static final Map<Class<?>, String> TEMPORAL_PARSE_FALLBACKS =
```
fmeum: A `ClassValue` may provide faster access than a `HashMap` with `Class` keys.

src/main/java/com/code_intelligence/jazzer/runtime/TraceCmpHooks.java
```diff
@@ -234,6 +248,176 @@ public static void equals(
...
+      targetClassName = "java.time.YearMonth",
+      targetMethod = "parse")
+  @MethodHook(
+      type = HookType.REPLACE,
+      targetClassName = "java.time.MonthDay",
+      targetMethod = "parse")
+  @MethodHook(
+      type = HookType.REPLACE,
+      targetClassName = "java.time.Duration",
+      targetMethod = "parse")
```
oetr, line 400: Edit: @Marcono1234 was faster!
_Concern_: Some of the hooked classes have two arguments. E.g. `java.time.LocalDate.parse(CharSequence text, DateTimeFormatter formatter)`, and since here we don't specify `targetMethodDescriptor`, the hooks apply to all versions of the `parse` function.
For the two argument version, we are guiding towards the default formatter, and not the one that the user specified.
In case of `LocalDate` it's ISO-8601.


It might make sense to split up the hooks to address the two overloads of `parse` individually.

In that case it would make sense to double-check if the user formatter is valid (it might be the reason we end up in the `catch` block). In that case we could safely use our default formatter.
But if the user provides a valid formatter, we should guide towards a date that respects it.

I usually abuse our fuzz test in `selffuzz/src/test/java/com/code_intelligence/selffuzz/mutation/ArgumentsMutatorFuzzTest.java` for quickly fuzzing anything. A fuzz test like this really struggles finding the correct string:
```java
private static final LocalDateTime TARGET_DATE_TIME =
        LocalDateTime.parse("2000-01-01T00:00:00");

  @SelfFuzzTest
  @Solo
  void fuzzPrimitiveArrays(@NotNull String dateStr) {
    try {
      LocalDateTime ldt = LocalDateTime.parse(dateStr,  DateTimeFormatter.ofPattern("dd MMM yyyy, HH:mm:ss", Locale.ENGLISH));
      System.out.println(ldt); // this never happens
      if (ldt.equals(TARGET_DATE_TIME)) {
        throw new FuzzerSecurityIssueLow("Found the target date time!");
      }
    } catch (DateTimeParseException ignore) {
    }
  }
```

To quickly run use:
```bash
bazel run //selffuzz/src/test/java/com/code_intelligence/selffuzz/mutation:ArgumentsMutatorFuzzTest -- -runs=10000000`

File: src/main/java/com/code_intelligence/jazzer/runtime/TraceCmpHooks.java [hunk 3]
```diff
@@ -234,6 +248,176 @@ public static void equals(
...
+          if (type == YearMonth.class) {
+            return Optional.of(
+                YearMonth.of(FALLBACK_LOCAL_DATE.getYear(), FALLBACK_LOCAL_DATE.getMonth())
+                    .toString());
+          }
+          if (type == MonthDay.class)
+            return Optional.of(MonthDay.from(FALLBACK_LOCAL_DATE).toString());
+          if (type == ZonedDateTime.class) {
+            return Optional.of(
+                ZonedDateTime.of(FALLBACK_LOCAL_DATE, LocalTime.MIDNIGHT, ZoneId.of("Europe/Paris"))
```
oetr, line 282: _Irrelevant, but fun_: Opus is asking "Why Paris?  Someone reading this will wonder "why Paris?" and waste 5 minutes figuring out there's no reason." 😆

src/main/java/com/code_intelligence/jazzer/runtime/TraceCmpHooks.java
Marcono1234, line 400: Is it intended that the annotations for `MonthDay` and `Duration` are not sorted alphabetically here?
(might make it a bit more difficult at a glance to see if all classes are covered)

src/main/java/com/code_intelligence/jazzer/runtime/TraceCmpHooks.java
```diff
@@ -234,6 +248,176 @@ public static void equals(
...
+      MethodHandle method, Object alwaysNull, Object[] arguments, int hookId) throws Throwable {
+    try {
+      return method.invokeWithArguments(arguments);
+    } catch (Throwable throwable) {
+      if (throwable instanceof Error) {
+        throw throwable;
+      }
+
+      if (arguments[0] != null) {
+        String fallback = TEMPORAL_PARSE_FALLBACKS.get(method.type().returnType()).orElse(null);
```
Marcono1234, line 412: (Edit: See more extensive review comment regarding this in https://github.com/CodeIntelligenceTesting/jazzer/pull/1051#discussion_r2932006448)

Out of curiosity, the above hook annotations also match the `parse` overloads with custom `DateTimeFormatter` (if I see it correctly), could that be a problem when that format completely differs from the standard ISO format?

Or is that considered acceptable, assuming that most likely the format is still somewhat close to the default ISO format?

tests/src/test/java/com/example/TimeParseFuzzer.java
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright 2024 Code Intelligence GmbH
```
Marcono1234, line 2: Should this be 2026?

Marcono1234 (commented):
Hopefully these comments are useful and not disruptive; if not please let me know.

oetr (commented):
Looks nice, but there is an issue when `parse` has two parameters.

tests/src/test/java/com/example/TimeParseFuzzer.java
onionpsy: Yes. Thanks!

PR #1052, opened 2026-03-15
Add a Dev Container

oleg-nenashev:
Creating this pull request rather for a discussion. The attached basic image works, but it has limited value as Bazel will reinstall the C++ stack and Java anyway, and the first build is very slow,

Would there be, in principle, an interest in a Dev Container for this repo? If so, I will get it finished

## Before merge

- [x] Basic image
- [ ] Ensure tools are pre-installed and cached in the Dev Container image
- [ ] Enable in-repository build caching by default
- [ ] Add Dev Container to the documentation

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,1 @@
+FROM silkeh/clang:21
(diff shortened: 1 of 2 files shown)


simonresch:
Sure! I'm happy to try out a Dev Container setup if you want to continue with this PR. :+1: 

I suspect internally we would not be actively using it but it could make external contributions easier which is always a win.

